PR #28, opened 2015-01-15
Add tournament exercise

pminten:
This is my first exercise port to C#, so if anything's sketchy with the style or logic please tell me.

### DIFF
--- a/config.json
+++ b/config.json
@@ -43,7 +43,8 @@
     "secret-handshake",
     "clock",
     "linked-list",
-    "wordy"
+    "wordy",
+    "tournament"
   ],
   "deprecated": [
 
--- a/tournament/Example.cs
+++ b/tournament/Example.cs
@@ -1,0 +1,127 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+/// <summary>
+/// Tournament tally system.
+/// </summary>
+public class Tournament
+{
+    enum Outcome
+    {
+        LOSS,
+        DRAW,
+        WIN
+    }
+    
+    class TeamResult
+    {
+        internal uint Losses;
+        internal uint Draws;
+        internal uint Wins;
+        
+        internal uint Played {
+            get {
+                return Losses + Draws + Wins;
+            }
+        }
+        
+        internal uint Score {
+            get {
+                return Wins * 3 + Draws;
+            }
+        }
+        
+        internal void AddOutcome(Outcome outcome)
+        {
+            switch (outcome) {
+                case Outcome.LOSS:
+                    Losses++;
+                    break;
+                case Outcome.DRAW:
+                    Draws++;
+                    break;
+                case Outcome.WIN:
+                    Wins++;
+                    break;
+            }
+        }
+    }
+    
+    private Dictionary<string, TeamResult> teams;
+    
+    private Tournament()
+    {
+        teams = new Dictionary<string, TeamResult>();
+    }
+    
+    private void AddResult(string team1, string team2, Outcome outcome)
+    {
+        // Invert outcome for the second team.
+        Outcome outcome2 =
+            outcome == Outcome.WIN ? Outcome.LOSS :
+            outcome == Outcome.LOSS ? Outcome.WIN :
+            Outcome.DRAW;
+        AddTeamOutcome(team1, outcome);
+        AddTeamOutcome(team2, outcome2);
+    }
+    
+    private void AddTeamOutcome(string team, Outcome outcome)
+    {
+        TeamResult teamResult;
+        if (this.teams.TryGetValue(team, out teamResult)) {
+            teamResult.AddOutcome(outcome);
+        } else {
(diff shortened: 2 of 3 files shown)


dinkelburt:
Hi @pminten, I took a look at the code and it's an interesting exercise. Looking around I found only one implementation in the Rust track so I assume it's super new.

I'm curious, what is the focus of the exercise? I see an element of creating/modifying custom data structures, an element of string manipulation, and an element of file I/O.

pminten:
The idea behind the exercise was to have a somewhat realistic but still simple data crunching exercise. This exercise involves the steps you usually need for that: file I/O, parsing, transformation and formatting.

dinkelburt:
I guess the reason I ask is unit tests that open/write files is usually a coding smell. I think there is value in testing file I/O in integration tests but I'm not sure if that's a goal for exercism.io (@kytrinyx, please correct me if I'm wrong). On the other hand, file I/O is something interesting to work with.

Code smells aside, while users can implement the exercise any way they'd like, the current file I/O implementation here took some work to even get running in an IDE but I have an idea on how to make it more IDE friendly.

1. Remove the txt files.
2. Do all file I/O on-demand. This won't allow the code to use ```TestCase```s but that's OK.

For example:

        [Test]
        public void Tallies_scores()
        {
            const string INPUT =
    @"Allegoric Alaskians;Blithering Badgers;win
    Devastating Donkeys;Courageous Californians;draw
    Devastating Donkeys;Allegoric Alaskians;win
    Courageous Californians;Blithering Badgers;loss
    Blithering Badgers;Devastating Donkeys;loss
    Allegoric Alaskians;Courageous Californians;win";
            const string EXPECTED =
    @"Team                           | MP |  W |  D |  L |  P
    Devastating Donkeys            |  3 |  2 |  1 |  0 |  7
    Allegoric Alaskians            |  3 |  2 |  0 |  1 |  6
    Blithering Badgers             |  3 |  1 |  0 |  2 |  3
    Courageous Californians        |  3 |  0 |  1 |  2 |  1";

            File.WriteAllText("input.txt", INPUT);
            Tournament.Tally("input.txt", "output.txt");

            Assert.That(File.ReadAllText("output.txt"), Is.EqualTo(EXPECTED));
        }

What this buys us:

1. Don't need to worry about user directory structure.
2. Users can see the input and expected output without opening additional files.

You could swap out the string literal for newline characters, just keep in mind that line endings are different in Windows vs. Unixy systems.

Anyway, let me know what you think.

kytrinyx:
I think that it's useful to have simple real-world type problems like this, though it's probably worth having them towards the end so that people have already touched a lot of other core parts of the language that don't require network operations, etc.

The real point of exercism isn't the (unit/whatever) tests, but the conversation that happens after the code is submitted. The tests are really there to help people know when they've got something working.

I do want tests to be reasonably fast, but that doesn't seem to be a problem here.

pminten:
New version, I've replaced the files with strings as @dinkelburt suggested, but with a slight twist: the user gets streams. This way it's still somewhat close to the file I/O scenario but there's no actual files being created.

dinkelburt:
Very good, that works much better in an IDE. I appreciate the effort to get away from using files to make the code more testable. I'm curious though, which did you prefer? While I wouldn't go with the file approach in a production environment, it's probably OK for a trivial exercise like this. I just wanted to verify with Katrina first.

If you would like to go back to files, feel free to do so (with similar changes I suggested previously). I'm OK with going forward with streams as well.

If you want to stick with streams, I've got some nitpicks for the test class:

1. The test cases aren't buying you anything here. Create a common method that you can call and then create descriptive methods that call it with the appropriate input/expected strings. This also eliminates the need for the potentially confusing ```data``` variable.
2. If we're going to use streams, change the ```Tally``` method to take an input stream and return a resulting stream. This way you don't need to guess the buffer size on the output stream.
3. [minor] There are 2 ```UTF8Encoding``` objects being created when only one is needed.
4. [minor] It's considered good style to always put usings at the top of the file instead of including inline (e.g. ```System.Text.UTF8Encoding```).
5. [observation] String literals will include all newline characters including the ones just after ```@"``` and just before the single line ```";```. I don't know if this was intentional or not, but the tests pass.

pminten:
I guess in a production environment it would depend on the natural API interface for a class, but I'd prefer streams for testing as well since they're easier to use (no temp dirs to make and keep clean).

As for your nitpicks:

1. Good point, really cleaned things up.
2. The memory stream is expandable (the initial size was my mistake, a left over from some figuring-out work) so that's not really a problem. I pass a stream because that's closer to a scenario where the resource lifetime is managed using "using" blocks. Could change it though if you want.
3. Whoops.
4. Yeah, missed that one.
5. True. The tests pass because I use Trim() on the strings.

Thanks for the nits.

dinkelburt:
Not seeing the patch, but I could be failing at github.

pminten:
It was my git-fu failing on windows. Apparently doing the git commit in the git shell didn't work, I had to do it through TortoiseGit. Weird stuff.

PR #30, opened 2015-02-10
Added equality check to Clock exercise

ErikSchierboom:
The readme file of the Clock exercise states that clocks with the same time should be equal to each other. This PR adds a check to see if that is the case.


kytrinyx:
Thanks!

PR #32, opened 2015-03-27
Added acronym exercise

ErikSchierboom:


dinkelburt:
```TestCase```s are cool but instead of using an ```[Ignore]``` attribute for all of ```Phrase_abbreviated_to_acronym```, can you add ```Ignore = true``` for every test case? Some IDEs and test runners are smart about reporting test cases separately, but others aren't. This allows users to incrementally test-drive through the tests instead of all of them at once.

ErikSchierboom:
@dinkelburt I agree. The PR has been updated.

PR #40, opened 2015-12-01
Update sum of multiples test

duffn:
- Add the missing sum of multiples test. Addresses https://github.com/exercism/todo/issues/52
- Remove the NUnit Ignore attribute from the other tests.

### DIFF
--- a/sum-of-multiples/SumOfMultiplesTest.cs
+++ b/sum-of-multiples/SumOfMultiplesTest.cs
@@ -32,6 +32,12 @@ public class SumOfMultiplesTest
         Assert.That(sumOfMultiples.To(10), Is.EqualTo(23));
     }
 
+    [Test]
+    public void Sum_to_100()
+    {
+       Assert.That(sumOfMultiples.To(100), Is.EqualTo(2318));
+    }
+
     [Ignore]
     [Test]
     public void Sum_to_1000()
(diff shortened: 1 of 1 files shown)


kytrinyx:
I'm curious about the `Ignore` attribute thing (I haven't been following the C# track closely). Is there a different issue open somewhere that says to not have the ignore directives?

Would you mind splitting this into two different pull requests? I can comfortably merge the one that adds the test (with an ignore directive), but I'm not sure about removing ignore directives, since it addresses a different problem and I've not been a part of the discussion about that.

bressain:
I was under the impression that tests should be ignored until people are ready to continue on. Did this change?

kytrinyx:
@bressain ah, you would have heard about it if so. I don't think so.

@duffn The ignore directives are there on purpose for all but the first test in order to simulate test driven development. If the C# test runner has a "fail fast" mode and a way to run the tests in the order specified, then we could lean on that, but otherwise I would prefer to leave them in place unless there's a good argument for removing them.

duffn:
@kytrinyx I see, that makes sense. I will update.

duffn:
Closed in favor of PR without removing the Ignore attributes. https://github.com/exercism/xcsharp/pull/41

kytrinyx:
Cool, thanks.

PR #43, opened 2015-12-19
Switch to container-based Travis CI infrastructure

kytrinyx:
This gets us off the legacy build system, which means
that the builds will start more quickly and also likely
run more quickly (and with more consistent build times).

See https://docs.travis-ci.com/user/migrating-from-legacy.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,3 +3,4 @@ language: bash
 script:
 - bin/fetch-configlet
 - bin/configlet .
+sudo: false


bressain:
I don't have any experience with Travis CI so I'm going to take your word for it working :smile:. The change seems like it should work according to the docs so I can merge or you can go ahead.

kytrinyx:
OK, cool! Done.

PR #51, opened 2016-02-09
Add Fake build script

ErikSchierboom:
This PR adds a Fake build script that allows all tests to run against the example implementations.

In the process, I immediately benefitted from having these automatic tests: the `LargestSeriesProductTest` class contained an incorrect test (parameterized test which was not defined as such).

Furthermore, the tests for the crypto-square exercise failed. This turned out to be both an implementation issue as well as test for the wrong expected values. For the correct expected values, I check several other language tracks, including the Haskell and Clojure tracks, which both had the same expected value that our test did not have.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,8 @@
 tmp
 bin/configlet
 bin/configlet.exe
+.fake/
+.vs/
+tools/
+build/
+TestResult.xml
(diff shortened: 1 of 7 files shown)


jwood803:
Nice!

PR #56, opened 2016-02-18
Add 5 new exercises

ErikSchierboom:
This PR adds four new exercises:

1. nth-prime
2. palindrome-products
3. binary-search-tree
4. robot-simulator
5. simple-linked-list

These are all open exercises as listed in #2.

I hope someone would be willing to give some feedback on the design. In particular, I struggled with the design of the palindrome products.

### DIFF
--- a/exercises/binary-search-tree/Example.cs
+++ b/exercises/binary-search-tree/Example.cs
@@ -1,0 +1,79 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+
+public class BinarySearchTree : IEnumerable<int>
+{
+    public BinarySearchTree(int value)
+    {
+        Value = value;
+    }
+
+    public BinarySearchTree(IEnumerable<int> values)
+    {
+        var array = values.ToArray();
+
+        if (array.Length == 0)
+        {
+            throw new ArgumentException("Cannot create tree from empty list");
+        }
+
+        Value = array[0];
+
+        foreach (var value in array.Skip(1))
+        {
+            Add(value);
+        }
+    }
+
+    public int Value { get; }
+
+    public BinarySearchTree Left { get; private set; }
+
+    public BinarySearchTree Right { get; private set; }
+
+    public BinarySearchTree Add(int value)
+    {
+        if (value <= Value)
+        {
+            Left = Add(value, Left);
+        }
+        else
+        {
+            Right = Add(value, Right);
+        }
+
+        return this;
+    }
+
+    private static BinarySearchTree Add(int value, BinarySearchTree tree)
+    {
+        if (tree == null)
+        {
+            return new BinarySearchTree(value);
+        }
+
+        return tree.Add(value);
+    }
+
+    public IEnumerator<int> GetEnumerator()
+    {
+        foreach (var left in Left?.AsEnumerable() ?? Enumerable.Empty<int>())
+        {
+            yield return left;
+        }
+
+        yield return Value;
+
+        foreach (var right in Right?.AsEnumerable() ?? Enumerable.Empty<int>())
+        {
+            yield return right;
+        }
+    }
+
+    IEnumerator IEnumerable.GetEnumerator()
+    {
+        return GetEnumerator();
+    }
+}
--- a/exercises/nth-prime/NthPrimeTest.cs
+++ b/exercises/nth-prime/NthPrimeTest.cs
@@ -1,0 +1,20 @@
+﻿using NUnit.Framework;
+
+public class NthPrimeTest
+{
+    [TestCase(1, ExpectedResult = 2)]
+    [TestCase(2, ExpectedResult = 3, Ignore = "Remove to run test case")]
+    [TestCase(3, ExpectedResult = 5, Ignore = "Remove to run test case")]
+    [TestCase(4, ExpectedResult = 7, Ignore = "Remove to run test case")]
+    [TestCase(5, ExpectedResult = 11, Ignore = "Remove to run test case")]
+    [TestCase(6, ExpectedResult = 13, Ignore = "Remove to run test case")]
+    [TestCase(7, ExpectedResult = 17, Ignore = "Remove to run test case")]
+    [TestCase(8, ExpectedResult = 19, Ignore = "Remove to run test case")]
+    [TestCase(1000, ExpectedResult = 7919, Ignore = "Remove to run test case")]
+    [TestCase(10000, ExpectedResult = 104729, Ignore = "Remove to run test case")]
+    [TestCase(10001, ExpectedResult = 104743, Ignore = "Remove to run test case")]
+    public int Nth_prime_calculated(int nth)
+    {
+        return NthPrime.Nth(nth);
+    }
+}
--- a/exercises/palindrome-products/Example.cs
+++ b/exercises/palindrome-products/Example.cs
@@ -1,0 +1,75 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+public class Palindrome
+{
+    private Palindrome(int value, ISet<Tuple<int, int>> factors)
+    {
+        Value = value;
+        Factors = factors;
+    }
+
+    public int Value { get; }
+
+    public ISet<Tuple<int, int>> Factors { get; }
+
+    public static Palindrome Largest(int maxFactor)
+    {
+        return Largest(1, maxFactor);
+    }
+
+    public static Palindrome Largest(int minFactor, int maxFactor)
+    {
+        return FindPalindrome(minFactor, maxFactor, x => x.Max(p => p.Item1));
+    }
+
+    public static Palindrome Smallest(int maxFactor)
+    {
+        return Smallest(1, maxFactor);
+    }
+
+    public static Palindrome Smallest(int minFactor, int maxFactor)
+    {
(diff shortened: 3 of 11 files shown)


exercises/binary-search-tree/Example.cs
```diff
@@ -0,0 +1,74 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+public class BinarySearchTree<T>
+    where T : IComparable
```
bressain: The exercise only talks about using numbers so there's no point in making this generic. I would keep it simple and just make this non-generic and use `int`s.

exercises/binary-search-tree/Example.cs
```diff
@@ -0,0 +1,74 @@
...
+{
+    private BinarySearchTree(T value)
```
bressain: While I understand some of the reasonings behind using static methods to create objects, I don't feel that it adds much value here.

exercises/binary-search-tree/Example.cs
```diff
@@ -0,0 +1,74 @@
...
+        if (value.CompareTo(Value) <= 0)
+        {
+            Left = Insert(value, Left);
+        }
+        else
+        {
+            Right = Insert(value, Right);
+        }
+
+        return this;
```
bressain: I like that this is composable :-)

exercises/binary-search-tree/Example.cs
```diff
@@ -0,0 +1,74 @@
...
+            yield return right;
+        }
+    }
+
+    public static BinarySearchTree<T> Singleton(T value)
+    {
+        return new BinarySearchTree<T>(value);
+    }
+
+    public static BinarySearchTree<T> FromEnumerable(IEnumerable<T> values)
```
bressain: We have this sweet composable `Insert` method but never actually use it because of this method. I would suggest dropping this method and utilize the `Insert` method to build the trees in the tests. It would further reinforce what the algorithm is trying to teach and would look a little more like the other language tracks.

exercises/nth-prime/NthPrimeTest.cs
```diff
@@ -0,0 +1,20 @@
...
+    [TestCase(5, ExpectedResult = 11, Ignore = "Remove to run test case")]
+    [TestCase(6, ExpectedResult = 13, Ignore = "Remove to run test case")]
+    [TestCase(7, ExpectedResult = 17, Ignore = "Remove to run test case")]
+    [TestCase(8, ExpectedResult = 19, Ignore = "Remove to run test case")]
+    [TestCase(1000, ExpectedResult = 7919, Ignore = "Remove to run test case")]
+    [TestCase(10000, ExpectedResult = 104729, Ignore = "Remove to run test case")]
+    [TestCase(10001, ExpectedResult = 104743, Ignore = "Remove to run test case")]
+    public int Nth_prime_calculated(int nth)
+    {
+        return NthPrime.Nth(nth);
```
bressain: :+1:

exercises/palindrome-products/Example.cs
```diff
@@ -0,0 +1,65 @@
...
+    {
+        Value = value;
+        Factors = factors;
+    }
+
+    public int Value { get; }
+
+    public ISet<Tuple<int, int>> Factors { get; }
+
+    public static Palindrome Largest(int minFactor, int maxFactor)
```
bressain: I noticed that other language tracks allow for `Largest` & `Smallest` to take just a `maxFactor` and default `minFactor` to `1`. Maybe add an overload that does that?

exercises/robot-simulator/Example.cs
```diff
@@ -0,0 +1,122 @@
+﻿using System;
+
+public enum Bearing
```
bressain: I like that you used enums here instead of strings like some other tracks.

exercises/simple-linked-list/Example.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+
+public class SimpleLinkedList<T>
```
bressain: Looking at other implementations of this, there are some methods that would be nice to have represented in the tests: `Head`, `Push()`, `Pop()`, & `Peek()`. They do `Size()` but since this is .Net, I could go with `Count`, maybe even try implementing `IList`? Your call.

exercises/simple-linked-list/Example.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        Value = value;
+        Next = next;
+    }
+
+    public T Value { get; }
+
+    public SimpleLinkedList<T> Next { get; }
+
+    public static SimpleLinkedList<T> FromEnumerable(IEnumerable<T> values)
```
bressain: If this were to match the style of .Net collections, it would make sense to have this as a constructor. That said, I'd like to see the other list methods utilized (`Head`, `Push`, `Pop`, `Peek`) instead of just pulling everything in at once to exercise how a linked list would be used.

exercises/palindrome-products/Example.cs
ErikSchierboom: Good suggestion.

exercises/binary-search-tree/Example.cs
ErikSchierboom: Agreed, I'll remove the `Singleton` method (which is indeed superfluous).

exercises/simple-linked-list/Example.cs
ErikSchierboom: Implementing IList would be a great suggestion! I'll do that.

exercises/simple-linked-list/Example.cs
ErikSchierboom: A constructor makes sense, and maybe also add the other list methods?

exercises/binary-search-tree/Example.cs
ErikSchierboom: Agreed.

exercises/binary-search-tree/Example.cs
ErikSchierboom: Okay, I will convert it.

exercises/simple-linked-list/Example.cs
bressain: Your call. The more I think about implementing `IList`, the more I realize that it might not make sense for a linked list (since it has methods that involve indices that linked lists don't really support well). However, `IList` implements `ICollection` which might make more sense.

exercises/simple-linked-list/Example.cs
bressain: See my other note about possibly implementing `ICollection` instead of `IList`.

exercises/simple-linked-list/Example.cs
ErikSchierboom: Valid point about the indices. `ICollection` might indeed be a better choice. As for the `Head`, `Push`, `Pop` and `Peek` methods, I actually think they don't belong on the list. If you look at the exercise, it says:

> As a first take, lets create a singly linked list to contain the range (1..10), and provide functions to reverse a linked list and convert to and from arrays.

The `Head`, `Push`, `Pop` and `Peek` can be part of e.g. a queue or a stack, which uses a linked list as the underlying data type, but don't necessarily belong to the linked list themselves.

bressain: Ah, I should have looked at the exercise more closely. Well shoot, now I question if implementing `ICollection` would complicate things rather than help. Yeah, let's forget about `Head`, `Push`, `Pop` & `Peek`. You can leave the class as-is as well. Sorry about the misunderstanding.

kytrinyx:
@ErikSchierboom @bressain I'm going to leave this discussion to the two of you since I don't know C#. If you are happy with any individual exercises, it might be worth sticking them in separate PRs or cherry-picking them while you keep working on and discussing the others. Or not :)

exercises/binary-search-tree/Example.cs
ErikSchierboom: Done.

exercises/binary-search-tree/Example.cs
ErikSchierboom: Converted to a constructor.

exercises/binary-search-tree/Example.cs
ErikSchierboom: Singleton method is removed :) Furthermore, to make things more idiomatic C#, I've implemented `IEnumerable<int>`, although I only require the `AsEnumerable()` method to exist in the test. That means users can either implement `IEnumerable<int>` and get `AsEnumerable()` for free, or just implement `AsEnumerable()`.

exercises/simple-linked-list/Example.cs
ErikSchierboom: I did some modifications to better align this implementation with that of the binary-search-tree exercise. As such, the singleton methods are replaced by constructors, I've implemented IEnumerable<T> and I have added an `Add()` method that I use in the class itself to add new items to the collection.

exercises/binary-search-tree/Example.cs
ErikSchierboom: Furthermore, I've renamed the `Insert()` method to `Add()`, which is also more idiomatic and which, when combined with the `IEnumerable<int>` implementation, allows one to use a collection initializer. People could thus do this:

```csharp
var elements = new BinarySearchTree(4)
        {
            5,
            6,
            7
        };
```

although the tests don't use this fact.

bressain:
Looks good, thanks for doing these.

PR #78, opened 2016-08-10
Build improvements

ErikSchierboom:
- Use Paket for dependency management
- Normalize test file names
- Fix not all tests being run


kytrinyx:
Is this catching existing test failures?

ErikSchierboom:
Yes it is. There is still a bug somewhere, but when it's done it will automatically run all tests using the example code.

kytrinyx:
OK cool. How do you feel about merging this with a failing build? That always makes me super nervous, but on the other hand it's already broken, this is just finally saying so.

ErikSchierboom:
I will only merge this once it's working again. This is basically a work in progress.

PR #86, opened 2016-08-30
Add topics

ErikSchierboom:
As explained in #82, the new exercises format has support for specifying the topics of an exercise. This PR is an initial attempt to assign topics to the existing exercises. Hopefully, we can have a bit of a discussion on whether the suggested topics make sense or not.

More specifically, @jwood803 and @bressain, you are the other top contributors. Would you mind taking a look? Just to be clear, anyone is invited to share his/her opinion here.

### DIFF
--- a/config.json
+++ b/config.json
@@ -66,336 +66,460 @@
       "slug": "leap",
       "difficulty": 1,
       "topics": [
+        "Integers"
       ]
     },
     {
       "slug": "bob",
       "difficulty": 1,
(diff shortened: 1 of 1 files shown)


bressain:
Seems fine? I can't say that the topics are comprehensive but we can always go back and add more right?

ErikSchierboom:
@bressain Agreed. I'll merge this.

PR #87, opened 2016-08-30
Add HelloWorld solution and tests

robkeim:
While this one's a low hanging fruit, I enjoyed it in the F# track as it helped me get accustomed to the work flow without having any difficult code to write.

### DIFF
--- a/config.json
+++ b/config.json
@@ -59,7 +59,8 @@
     "simple-linked-list",
     "matrix",
     "queen-attack",
-    "ocr-numbers"
+    "ocr-numbers",
+	"hello-world"
   ],
   "exercises": [
     {
@@ -394,6 +395,12 @@
     },
     {
       "slug": "ocr-numbers",
+      "difficulty": 1,
+      "topics": [
+      ]
+    },
+    {
+      "slug": "hello-world",
       "difficulty": 1,
       "topics": [
       ]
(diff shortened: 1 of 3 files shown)


ErikSchierboom:
@robkeim Good idea! There is no thing left to do: the `hello-world` entry you added to the `exercises` array in the `config.json` file should become the first entry in that list.

robkeim:
I wasn't sure what the rule is for exercise ordering, but agreed that it makes sense to have `hello-world` be the first one :)

I've gone ahead and updated the PR.

config.json
```diff
@@ -4,6 +4,7 @@
   "repository": "https://github.com/exercism/xcsharp",
   "active": true,
   "problems": [
+	"hello-world",
```
ErikSchierboom: The indentation is off, sorry :)

config.json
```diff
@@ -59,10 +60,15 @@
     "simple-linked-list",
     "matrix",
     "queen-attack",
-    "ocr-numbers",
-	"hello-world"
+    "ocr-numbers"
   ],
   "exercises": [
+	{
```
ErikSchierboom: Also has the wrong indentation.

ErikSchierboom:
@robkeim There are just two small indentation problems, after that it's perfect!

config.json
robkeim: Ah, I got bit by my editor using tabs :)

ErikSchierboom:
Merged! Thanks. :tada:

PR #95, opened 2016-08-31
Add Minesweeper exercise

robkeim:


ErikSchierboom:
I think that perhaps the difficulty can be a bit lower. In F#, part of the difficulty comes from iterating over the input using a functional style. That is much easier in idiomatic C# (as you used). Perhaps it should be a 5 or 6 difficulty? Or do you think this was as difficult as the `wordy` exercise?

robkeim:
I simply copied the difficulty from the F# one, I didn't realize the difficulties varied per language, but that makes sense that they do. Are they defined somewhere that I can look them up?

ErikSchierboom:
Not really, it's just a scale from 1-10. In general, copying the difficulty from the corresponding F# exercise is fine, but in this case, I think it is easier in C#.

robkeim:
Ok sounds good, I've gone ahead and updated the difficulty.

ErikSchierboom:
Great, but you'll also need to update the ordering. At the moment, it is the last exercise, but it should be somewhere amongst the other exercises of difficulty 5 (both in the regular, old style exercises list and the new one). When I ordered the exercises, I also looked at the sort of exercise, such that two exercises with the same difficulty and similar topics would not follow each other.

Almost there :)

robkeim:
Third time's a charm? :)

config.json
```diff
@@ -50,6 +50,7 @@
     "simple-cipher",
     "roman-numerals",
     "binary-search-tree",
+	"minesweeper",
```
ErikSchierboom: Sorry, the indentation is off again :)

ErikSchierboom:
@robkeim Almost, the indentation is off again on one line. Sorry.

robkeim:
Ok, I've now changed my editor so it inserts spaces instead of tabs by default... I should have done that several exercises ago :)

ErikSchierboom:
@robkeim That's probably a wise decision :) Merged this. Thanks again! 🎉

PR #128, opened 2016-09-17
Align naming in LinkedList exercise

robkeim:
Here's the corresponding PR to make the same change in F#:
https://github.com/exercism/xfsharp/pull/222


robkeim:
Looks like this won't worked due to conflicts with the built-in .NET `LinkedList` class :(

ErikSchierboom:
Too bad

PR #151, opened 2016-12-19
Add two-bucket exercise

ErikSchierboom:


exercises/two-bucket/Example.cs
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    public void PourTo(BucketContainer other)
+    {
+        var amount = Math.Min(other.Capacity - other.Contents, Contents);
+        Contents -= amount;
+        other.Contents += amount;
+    }
+
+    public bool CanPourTo(BucketContainer other) => IsFull && !other.IsFull || !IsFull && other.IsEmpty;
```
robkeim: Is there a reason this logic isn't `!IsEmpty && !other.IsFull`? You should be able to pour whenever the current bucket isn't empty and the target bucket isn't full right?

robkeim:
Looks great @ErikSchierboom!

I'll wait for your response to my question before merging these changes.

ErikSchierboom:
@robkeim You were totally right. The check could be simplified to: `!IsEmpty && !other.IsFull`. I've updated the PR.

robkeim:
@ErikSchierboom is there a way to view the diff between what you pushed before and what you just pushed? I don't see an option to do so, and I only see one commit here. That would save me from having to look over the entire PR again.

ErikSchierboom:
@robkeim Oh whoops, that's because I rebased and squashed the PR myself. Now that that's possible from GitHub, I shouldn't really have to do that anymore. Old habits die hard :)

robkeim:
Ok, I'll help you try to break that habit then :)

PR #152, opened 2016-12-21
Add markdown exercise

ErikSchierboom:


robkeim:
Nice job @ErikSchierboom! It's not ever day you get to submit a PR with "bad" code and it will get accepted :)

I added one small question in the csproj file, and once I have your response, I'll go ahead and merge this change.

ErikSchierboom:
@robkeim :) What's the csproj file question? I can't seem to find it.

exercises/exercises.csproj
```diff
@@ -38,7 +38,7 @@
     <Reference Include="Microsoft.CSharp" />
   </ItemGroup>
   <ItemGroup>
-    <Compile Include="**\*.cs" Exclude="sgf-parsing\SgfParsing.cs" />
+    <Compile Include="**\*.cs" Exclude="sgf-parsing\SgfParsing.cs;markdown\Markdown.cs" />
```
robkeim: @ErikSchierboom - can you see this comment? I just "published" the review so I'm not sure it was visible before that (I'm still learning the nuances of GitHub)

ErikSchierboom: Now I can see it. The reason this file is excluded is simple: we otherwise have two files (the example and the default implementation) with the same namespace and class name, which is a compile error.

robkeim: Makes sense!

robkeim:
Thanks @ErikSchierboom this looks great!

PR #155, opened 2016-12-24
Add go-counting exercise

ErikSchierboom:


exercises/go-counting/Example.cs
```diff
@@ -0,0 +1,148 @@
...
+            new Point(col, row - 1),
+            new Point(col-1, row),
+            new Point(col+1, row),
+            new Point(col, row+1)
+        };
+
+        return coords.Where(IsValidCoordinate);
+    }
+
+    private IEnumerable<Point> NonEmptyNeighborCoordinates(Point coordinate) =>
```
robkeim: To be consistent with the terminology above this should be `TakenNeighborCoordinates` to match with `isTaken`/`isEmpty`

robkeim: @ErikSchierboom I've left one easy to address comment, and then I'll merge.

ErikSchierboom: Great point! I've updated the code and also renamed `isTaken` and `isEmpty` to `IsTaken` and `IsEmpty` and then used `IsTaken` in `TakenNeighborCoordinates` and removed the unnecessary call to `IsTaken` in `TerritoryOwner`.

robkeim:
Great work, thanks @ErikSchierboom!

PR #159, opened 2017-01-09
difference-of-squares: Change expected exception type

balazsbotond:
I think it would be better to use the more specific `ArgumentOutOfRangeException` instead of `ArgumentException` here.

### DIFF
--- a/exercises/difference-of-squares/DifferenceOfSquaresTest.cs
+++ b/exercises/difference-of-squares/DifferenceOfSquaresTest.cs
@@ -77,6 +77,6 @@ public class DifferenceOfSquaresTests
     [Test]
     public void Test_negative_numbers_throw_argument_exception()
     {
-        Assert.That(() => new Squares(-5), Throws.TypeOf<ArgumentException>());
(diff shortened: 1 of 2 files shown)


exercises/difference-of-squares/DifferenceOfSquaresTest.cs
```diff
@@ -77,6 +77,6 @@ public void Test_difference_of_sums_0()
     [Test]
     public void Test_negative_numbers_throw_argument_exception()
```
robkeim: Could you update the title to say `argument_out_of_range_exception`?

robkeim:
Thanks @balazsbotond for the PR! I've gone ahead and added my input, could you please take a look and update the changes?

This will be a small breaking change for anyone that's already implemented this but doesn't change any of the core problem, so I think it should be fine.

exercises/difference-of-squares/DifferenceOfSquaresTest.cs
balazsbotond: Sure!

balazsbotond:
@robkeim thanks for the review! I've renamed the test.

robkeim:
Merging this, thanks @balazsbotond!

PR #163, opened 2017-01-11
Resolve #160 - Use different output and input type parameters in Accumulate

### LINKED ISSUE #160: Use different output and input type parameters in Accumulate
The current example for the Accumulate exercise has the following method signature:

```csharp
public static IEnumerable<T> Accumulate<T>(this IEnumerable<T> collection, Func<T, T> func)
```

This does not allow converting the type of sequence elements, like this:

```csharp
"1 2 3 4".Split(' ').Select(s => Convert.ToInt32(s)); // does not compile
```

Is this deliberate? If not, the method signature could be changed like this:

```csharp
public static IEnumerable<U> Accumulate<T, U>(this IEnumerable<T> collection, Func<T, U> func)
```

What do you think?

If this is not deliberately being kept simple because it is an exercise for beginners, I could send a PR with the new method signature and a unit test.

balazsbotond:


ErikSchierboom:
Looks good @balazsbotond! Merged. :tada:

PR #165, opened 2017-01-12
Add dot-dsl exercise

ErikSchierboom:


exercises/dot-dsl/Example.cs
```diff
@@ -0,0 +1,90 @@
...
+    public Node(string name)
+    {
+        Name = name;
+    }
+
+    public string Name { get; }
+
+    public override bool Equals(object obj)
+    {
+        if (obj == null || GetType() != obj.GetType())
```
robkeim: Assuming you're using C# 6 you can simplify the if to:
`if (GetType() != obj?.GetType())`

Also applies to other classes where you've overridden `Equals`.

robkeim:
Great work @ErikSchierboom... I love the clean, concise solution!

Issue #171, opened 2017-01-16
Implement remaining exercises for track


robkeim:
@ErikSchierboom mentioned [here](https://github.com/exercism/xcsharp/pull/166#issuecomment-272947351) that we only have four exercises remaining in the C# track to be implemented:

- [x] [error-handling](https://github.com/exercism/x-common/tree/master/exercises/error-handling)
- [x] [forth](https://github.com/exercism/x-common/tree/master/exercises/forth)
- [x] [pov](https://github.com/exercism/x-common/tree/master/exercises/pov)
- [x] [zipper](https://github.com/exercism/x-common/tree/master/exercises/zipper)

There's light at the end of the tunnel :)

robkeim:
@ErikSchierboom - I'll give error-handling a go

ErikSchierboom:
@robkeim Cool!

ErikSchierboom:
@robkeim I'm working on the `pov` exercise.

robkeim:
Sounds good @ErikSchierboom!

ErikSchierboom:
@robkeim I'll probabyl work on the `zipper` exercise tonight.

robkeim:
You're a machine @ErikSchierboom :)

ErikSchierboom:
We'll, I do have the advantage that I already solved it in F#. That means it's more porting than solving :)

robkeim:
Fair enough :)

robkeim:
Thanks @ErikSchierboom, the C# track now has every available exercise implemented 🎉

ErikSchierboom:
Hurray! Thanks for the great help @robkeim!

@kytrinyx We now have all exercises available in C# too! :tada:

kytrinyx:
@robkeim @ErikSchierboom 🎉 Brilliant! 🎈

PR #174, opened 2017-01-17
Add pov exercise

ErikSchierboom:
One of the three remaining exercises as specified in #171.

### DIFF
--- a/exercises/pov/Example.cs
+++ b/exercises/pov/Example.cs
@@ -1,0 +1,125 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+public class Graph<T> : IEquatable<Graph<T>>
+{
+    public Graph(T value, IEnumerable<Graph<T>> children)
+    {
+        Value = value;
+        Children = children;
+    }
+
+    public T Value { get; }
+    public IEnumerable<Graph<T>> Children { get; }
+
+    public bool Equals(Graph<T> other) =>
+        Value.Equals(other.Value) && Children.SequenceEqual(other.Children);
+}
+
+public class Crumb<T>
+{
+    public Crumb(T value, IEnumerable<Graph<T>> left, IEnumerable<Graph<T>> right)
+    {
+        Value = value;
+        Left = left;
+        Right = right;
+    }
+
+    public T Value { get; }
+    public IEnumerable<Graph<T>> Left { get; }
+    public IEnumerable<Graph<T>> Right { get; }
+}
+
+public class Zipper<T>
+{
+    public Zipper(Graph<T> focus, IEnumerable<Crumb<T>> crumbs)
+    {
+        Focus = focus;
+        Crumbs = crumbs;
+    }
+
+    public Graph<T> Focus { get; }
+    public IEnumerable<Crumb<T>> Crumbs { get; }
+}
+
+public static class Pov
+{
+    public static Graph<T> CreateGraph<T>(T value, IEnumerable<Graph<T>> children) 
+        => new Graph<T>(value, children);
+
+    public static Graph<T> FromPOV<T>(T value, Graph<T> graph) where T : IComparable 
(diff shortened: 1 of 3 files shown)


exercises/pov/Example.cs
```diff
@@ -0,0 +1,125 @@
...
+    public Graph<T> Focus { get; }
+    public IEnumerable<Crumb<T>> Crumbs { get; }
+}
+
+public static class Pov
+{
+    public static Graph<T> CreateGraph<T>(T value, IEnumerable<Graph<T>> children) 
+        => new Graph<T>(value, children);
+
+    public static Graph<T> FromPOV<T>(T value, Graph<T> graph) where T : IComparable
```
robkeim: This is minor, but based on [C# style guidelines](https://msdn.microsoft.com/en-us/library/x2dbyw72(v=vs.71).aspx) this should be `FromPov` since POV is more than two letters. That would also make it consistent with the class name.

exercises/pov/Example.cs
```diff
@@ -0,0 +1,125 @@
...
+
+        return new Zipper<T>(graph, Enumerable.Empty<Crumb<T>>());
+    }
+
+    private static IEnumerable<T> ZipperToPath<T>(Zipper<T> zipper)
+    {
+        if (zipper == null)
+            return null;
+
+        return zipper.Crumbs.Select(c => c.Value).Reverse().Concat(new[] { zipper.Focus.Value });
```
robkeim: Are you using C# 6? If so the [`?.` operator](https://msdn.microsoft.com/en-us/library/dn986595.aspx) is your friend :) There are other places below as well that you could use it, such as lines 75 and 88.

exercises/pov/Example.cs
```diff
@@ -0,0 +1,125 @@
...
+        if (!zipper.Crumbs.Any())
+            return zipper.Focus;
+
+        var firstCrumb = zipper.Crumbs.First();
+        var focus = zipper.Focus;
+
+        var newZipper = new Zipper<T>(CreateGraph(firstCrumb.Value, firstCrumb.Left.Concat(firstCrumb.Right)), zipper.Crumbs.Skip(1));
+        var parentGraph = ChangeParent(newZipper);
+
+        var ys = focus.Children.Concat(new[] { parentGraph });
```
robkeim: What does `ys` stand for? Is there a reason you didn't go with `children`?

robkeim:
@ErikSchierboom I added some *really* nitpick comments, but I'll merge these changes as soon as the Travis CI build finishes.

exercises/pov/Example.cs
ErikSchierboom: Good point!

exercises/pov/Example.cs
ErikSchierboom: I am using C# 6, but I'm quite hesitant to use the `?.` operator, as I find it can easily obscure code.

exercises/pov/Example.cs
ErikSchierboom: No good reason :)

exercises/pov/Example.cs
robkeim: Do you have any specific examples in mind where you feel `?.` obscures the code? I'm curious because I've found that using it has simplified a lot of "noise" in the code by doing null checks all over the place.

ErikSchierboom: Well, not an exact example, but in something like `zipper.Crumbs.Select(c => c.Value).Reverse()?.Concat(new[] { zipper.Focus?.Value });` it can be a bit hard to notice the `?.` operators. I'm not against using the operator, but I am a little bit hesitant to use it.

robkeim: I still consider myself quite new to using it as well so I'm still trying to define the line between use and abuse of the operator :)

ErikSchierboom: It's a fine line :)

PR #177, opened 2017-01-18
Resolve #176 - Use em and strong tags correctly in Markdown

### LINKED ISSUE #176: Incorrect HTML tags for bold text in the tests for Markdown
I think the following test for the Markdown exercise expects the wrong tags:

```csharp
[Test]
public void Parsing_bold_text()
{
    var input = "__This will be bold__";
    var expected = "<p><em>This will be bold</em></p>";
    Assert.That(Markdown.Parse(input), Is.EqualTo(expected));
}
```

The `<em>` tag is rendered as italic text by default, so I think the correct tag here would be:

- either `<b>` (to match the `<i>` tag in the test for italic)
- or `<strong>`, but then the `<i>` tag in the italic test case should also be changed to `<em>`.

I think the latter option is the better one because `em` and `strong` are semantic tags.

The test case `Mixed_normal_italics_and_bold_text` should also be updated in both cases.

If you agree, I can send a PR that updates the tests and the example.

balazsbotond:


balazsbotond:
Thanks @robkeim!

ErikSchierboom:
Great work! :tada:

PR #178, opened 2017-01-19
Changed inputDigits in First_base_is_one test in AllYourBaseTest

FridaTveit:
Changed inputDigits in First_base_is_one test to be an array with values, not an empty array. Since the test for throwing an exception when inputDigits is empty is before this one, this test is redundant when inputDigits is empty. Assuming the intention is to test that an exception is thrown when the first base is one, this should fix that so that the correct thing is tested.

### DIFF
--- a/exercises/all-your-base/AllYourBaseTest.cs
+++ b/exercises/all-your-base/AllYourBaseTest.cs
@@ -156,7 +156,7 @@ public class AllYourBaseTest
     public void First_base_is_one()
     {
         const int inputBase = 1;
-        var inputDigits = new int[0];
+        var inputDigits = new[] { 1, 0, 1, 0, 1, 0 };
         const int outputBase = 10;
         Assert.Throws<ArgumentException>(() => Base.Rebase(inputBase, inputDigits, outputBase));
     }


ErikSchierboom:
Once more, fantastic work! Thanks a lot. :tada:

stkent:
@exercism/csharp xjava is struggling with this same test case [right now](https://github.com/exercism/xjava/issues/519). The change you made here seems like it only shifts ambiguity regarding input validation ordering, rather than removing that ambiguity, since 

    const int inputBase = 1;
    var inputDigits = new[] { 1, 0, 1, 0, 1, 0 };

could fail with either

1. an invalid base message (the error expected by the test); or
1. an invalid positive digit message (since the largest digit in the input digits is equal to the base).

In fact the only valid input digits for input base `1` would appear to be `{0}`, which you _could_ argue is no longer an erroneous pairing of inputs (if you treat this as a representation of zero).

Anything I'm missing? My inclination at this time is to reorder the tests in `x-common` so that validation occurs in parameter order (input base -> input digits -> output digits seems to be the natural ordering here). What would you all think of that as a change?

ErikSchierboom:
@stkent I think re-ordering would be the best option, and I quite like the ordering you suggest.

PR #182, opened 2017-01-24
Add forth exercise

ErikSchierboom:
This is the last of the C# exercises (as listed in #171)! Hurray! And it's not the least :)


robkeim:
Awesome work @ErikSchierboom... really clean solution!

PR #188, opened 2017-01-30
Use [Test] instead of [TestCase] in error-handling exercise

ErikSchierboom:
The [TestCase] attribute is for parameterized tests, [Test] for regular tests.


robkeim:
Good find, thanks @ErikSchierboom!

Issue #189, opened 2017-02-06
bank-account: Should GetBalance be a Balance property instead?


petertseng:
When I was doing this exercise, I remarked that it seemed unidiomatic that `GetBalance()` was a method, rather than having a `Balance` property.

But y'all know C# better than I do, so y'all are in a better position to answer - must it be a method? Or can/should it be a property?

ErikSchierboom:
It should indeed be a property, good catch!

ErikSchierboom:
@petertseng A [PR](#192) has been opened.

robkeim:
I've just gone ahead and merged the PR. Thanks @petertseng for bringing this up.

PR #192, opened 2017-02-06
Use property to retrieve balance in bank-account exercise

ErikSchierboom:
This PR implements the suggestion from #189 that retrieving the balance should be done through a property.


robkeim:
Nice work, thanks @ErikSchierboom!

PR #196, opened 2017-02-10
[WIP] xUnit

### LINKED ISSUE #193: Replace NUnit with Xunit
Xunit has slowly but steadily been replacing NUnit as the favored testing framework. Virtually all open-source Microsoft code now uses Xunit as the unit-testing framework. The integration with .NET Core is thus also better, as Microsoft has invested in Xunit for its own software and thus has to keep it running. An example of this improved integration is the fact that there is an Xunit template available in the .NET CLI.

p.s. I'm already working on this :)

ErikSchierboom:
This is a WIP PR to fix #193. There were several oddities, most notably the fact that xUnit doesn't out of the box handle nested enumerable equality. On the plus side, xUnit is now the preferred unit test framework for .NET Core and it will be very easy getting it to work there. Also, we no longer need a special case handling for AppVeyor (and possible CircleCI).

As there are so many changes, I'd really like a thorough review. If things are good, I think we should then leave it unmerged for a while, as otherwise we would have to adjust our documentation, which we would then have to anyway when we convert to .NET Core.

That would be the next thing I'm working on, converting to .NET Core. I will be doing that based on the code in this PR. Good times coming!


robkeim:
@ErikSchierboom I'd love to look through this but I'm moving tomorrow and I won't have time before then. I'll be back online after my vacation the second weekend of March.

I'm sure that you and @balazsbotond will keep everything up and running while I'm gone, and I'll "see" you soon!

ErikSchierboom:
But of course. No problem! When you're back, I hope to have . net core working.

ErikSchierboom:
This PR can be closed, as everything in this PR has been merged into #199.

Issue #207, opened 2017-02-20
Use static classes where possible


ErikSchierboom:
Some tests only require a static method to be present. If this is so, and if the implementation also only has static methods, we should convert the implementation to a static class.

tushartyagi:
@ErikSchierboom I can start working on this one. But this will also require changing the test cases and you are working on moving tests to xunit in [#193](https://github.com/exercism/xcsharp/issues/193) so it might be dependent on that. :)

What are your thoughts?

ErikSchierboom:
@tushartyagi There is indeed that xUnit dependency. You can work on this feature though, only you have to base your PR on the [`dotnet-core`](https://github.com/exercism/xcsharp/tree/dotnet-core) branch (which has the same contents my [PR](https://github.com/exercism/xcsharp/pull/199) which contains the xUnit changes). So you can definitely work on this, as long as you base it on the `dotnet-core` branch :)

By the way, if you will be working on this, there are some exercises that currently have a constructor with one parameter and one single method being tested. I think that in most cases, those classes can be replaced with static classes, where the method being tested just gets an extra parameter.

If you need any help, just contact me in the [Gitter room](https://gitter.im/exercism/xcsharp). Thanks for the help! It is much appreciated.

tushartyagi:
I have completed going through the exercises for classes which had:

1. only `static` methods but the classes were not `static`
2. tests were using `static` methods 

<br />

Following is the list of exercises:
- [X] atbash-cipher
- [X] grains
- [X] triangle
- [X] word-count
- [X] wordy
 
[PR#215](https://github.com/exercism/xcsharp/pull/215) submitted for this.

ErikSchierboom:
@tushartyagi Brilliant! I've requested on small change in the PR. Thanks for looking into this!

PR #215, opened 2017-02-28
Added static keyword to static classes

tushartyagi:
The following 5 exercises had only `static` methods but the classes were not `static`. Changes for the same.
Targets #207


tushartyagi:
@ErikSchierboom Done with changes.

ErikSchierboom:
Thanks a lot @tushartyagi!

PR #219, opened 2017-03-08
[WIP]: Use Cake instead of Fake.

tushartyagi:
There are a few things missing since this is a WIP. But gets the build working.

- [ ] Refactor the repeated `GetFiles(buildDir + "**/*.csproj")` pattern.
- [x] Change `project.ToString()` to `FilePath.FullPath`
- [ ] This currently runs all the projects. Allow project-name to be passed as command line param.
- [ ] Test on *nix.

- [ ] Remove paket related files.

Strangely, this is failing on AtbashTest.cs -- weirdly picking up the stub file even though that is replaced with the Example file. All the projects before it pass without any issues.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -1,0 +1,91 @@
+#addin "Cake.FileHelpers"
+
+// TODO:
+// Remove comments
+// Add option to take in project name from command line
+// Beautify
+// Remove limit of 2 from everywhere
+
+var target = Argument("target", "Default");
+var project = Argument("project", "*");
+
+var buildDir = "./build/";
+var sourceDir = "./exercises/";
+var projects = GetFiles(buildDir + "**/" + project + ".csproj");
+
+
+Task("Clean")
+	.Does(() => {
+		CleanDirectory("./build/");   
+    });
+
+
+// Copy everything to build so we make no changes in the 
+// actual files.
+Task("Copy-Exercises")
+    .IsDependentOn("Clean")
+    .Does(() => {
+        CopyDirectory(sourceDir, buildDir);
+    });
+
+
+Task("Restore-Nuget-Packages")
+    .IsDependentOn("Copy-Exercises")
+    .Does(() => {
+       foreach (var project in projects) {
+            DotNetCoreRestore(project.ToString());
+       }
+    });
+
+
+Task("Enable-All-Tests")
+    .IsDependentOn("Restore-Nuget-Packages")
+    .Does(() => {
+        var testFiles = GetFiles(buildDir + "**/*Test.cs");
+        foreach (var testFile in testFiles) {
+            var skipRE = @"Skip\s*=\s*""Remove to run test""";
+            ReplaceRegexInFiles(buildDir + "**/*Test.cs", skipRE, "");
+        }
+
+    });
+
+
+Task("Replace-Stub-With-Example")
+    .IsDependentOn("Enable-All-Tests")
+    .Does(() => {
+        foreach (var project in projects) {
+            var projectDir = project.GetDirectory();
+            var projectName = project.GetFilenameWithoutExtension();
+            var stub = projectDir.GetFilePath(projectName).AppendExtension("cs");
+            var example = projectDir.GetFilePath("Example.cs");
+            
+            DeleteFile(stub);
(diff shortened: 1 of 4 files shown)


build.cake
```diff
@@ -0,0 +1,108 @@
...
+        foreach (var project in defaultProjects) {
+            DotNetCoreBuild(project.ToString());
+        }
+});
+
+Task("EnableAllTests")
+    .IsDependentOn("RestoreAndBuildStubImplementations")
+    .Does(() => {
+        var testFiles = GetFiles(buildDir + "**/*Test.cs");
+        foreach (var testFile in testFiles) {
```
tushartyagi: This `foreach` can be removed since `ReplaceRegexInFiles` takes the glob as the param.

build.cake
```diff
@@ -0,0 +1,108 @@
+#addin "Cake.FileHelpers"
+
+var target = Argument("target", "Default");
+var project = Argument("project", "*");
```
ErikSchierboom: On my Mac, I got an error that this variable is shadowed, that is another variable is defined with the same name. To fix this, you could rename this variable to something like `projectFilter`.

build.cake
```diff
@@ -0,0 +1,108 @@
...
+/** These projects have full and working implementations, and have
+    all the tests enabled. These should pass without any changes.
+*/
+var refactoringProjects = GetFiles(buildDir + "**/TreeBuilding.csproj")
+    + GetFiles(buildDir + "**/Ledger.csproj")
+    + GetFiles(buildDir + "**/Markdown.csproj");
+
+
+Task("Clean")
+	.Does(() => {
```
ErikSchierboom: The indentation of this task seems a bit off (one indent step too much).

build.cake
ErikSchierboom: Good catch.

build.cake
```diff
@@ -0,0 +1,108 @@
...
+Task("TestUsingExampleImplementation")
+    .IsDependentOn("ReplaceStubWithExample")
+    .Does(() => {
+        foreach (var project in allProjects) {
+            DotNetCoreTest(project.ToString());
+        }
+    });
+
+
+Task("Default")
```
ErikSchierboom: This task is only there for serve as the default. I think you can remove this if you specify `"TestUsingExampleImplementation"` as the default value in the `var target = Argument("target", "Default");` line at the top of this file

build.cake
tushartyagi: This is part of the "run tests for a single project" feature. If this is not very important, we can skip this. The feature can be added as another PR?

build.cake
tushartyagi: I tried this. But `build.ps1` already has set the command line parameter `[string]$Target = "Default"`. Besides the `Default` task is set as the starting tasks in some of the cake build files that I saw on the internet, perhaps this is the convention?

tushartyagi:
@ErikSchierboom Travis is not running the tasks. Will have to check where it is failing.

build.cake
ErikSchierboom: Could be. Let's leave it then.

build.cake
ErikSchierboom: Yeah, let's skip it for now. There are more imporant things to work on.

ErikSchierboom:
@tushartyagi That's odd. It appears that on Travis, it doesn't find any files. I have some ideas how to fix that, which I'll post as review comments.

build.cake
```diff
@@ -0,0 +1,106 @@
+#addin "Cake.FileHelpers"
+
+var target = Argument("target", "Default");
+
+var buildDir = "./build/";
```
ErikSchierboom: Perhaps we could try using the [`DirectoryPath`](http://cakebuild.net/api/Cake.Core.IO/DirectoryPath/2EE6D32A) class instead of plain strings. This will allow us to convert this path to a full, absolute path using [`MakeAbsolute`](http://cakebuild.net/api/Cake.Core.IO/DirectoryPath/2FAC60BD).

ErikSchierboom:
@tushartyagi Thanks! I've fixed this in #226, which has just been merged.

Issue #194, opened 2017-02-07
Replace [TestCase] tests with individual tests


ErikSchierboom:
We have several places where we use `[TestCase]` to create a single test with multiple input values. While efficient (no code duplication), this is not the best of user experiences as it makes it harder to run a single test and also doesn't really allow us to mark the various input values with a descriptive text. It also doesn't align with how canonical data for exercises is defined, which only works on single test cases.

Consider the `isogram` exercises's tests:

```c#
[TestCase("duplicates", ExpectedResult = true)]
[TestCase("eleven", ExpectedResult = false, Ignore = "Remove to run test case")]
[TestCase("subdermatoglyphic", ExpectedResult = true, Ignore = "Remove to run test case")]
[TestCase("Alphabet", ExpectedResult = false, Ignore = "Remove to run test case")]
[TestCase("thumbscrew-japingly", ExpectedResult = true, Ignore = "Remove to run test case")]
[TestCase("Hjelmqvist-Gryb-Zock-Pfund-Wax", ExpectedResult = true, Ignore = "Remove to run test case")]
[TestCase("Heizölrückstoßabdämpfung", ExpectedResult = true, Ignore = "Remove to run test case")]
[TestCase("the quick brown fox", ExpectedResult = false, Ignore = "Remove to run test case")]
[TestCase("Emily Jung Schwartzkopf", ExpectedResult = true, Ignore = "Remove to run test case")]
[TestCase("éléphant", ExpectedResult = false, Ignore = "Remove to run test case")]
public bool Isogram_correctly_detects_isograms(string input)
{
    return Isogram.IsIsogram(input);
}
```

What's the difference between test case one and two? It's not immediately obvious. The [canonical-data](https://github.com/exercism/x-common/blob/master/exercises/isogram/canonical-data.json) of this exercise looks much nicer. I think we should convert the [TestCase] tests to individual tests.

robkeim:
In terms of the argument that we're missing the richness of the canonical data I think we could work around that by adding additional comments alongside of the test cases.

However, from a usability standpoint, I'm all for making the tests easier to run if they're difficult to run today. In my setup I'm using Visual Studio which has the ability to run all of the cases for a single test or to run them one by one very easily, but I know different tools don't support the same functionality.

Here are the list of exercises using the `TestCase` attribute:
- [ ] [acronym](https://github.com/exercism/xcsharp/tree/master/exercises/acronym)
[...]

PR #222, opened 2017-03-08
Replace Theory with Fact for Acronym exercise

ErikSchierboom:
Refs #194


robkeim:
Looks good, thanks @ErikSchierboom!

PR #225, opened 2017-03-12
[WIP] Start using CAKE instead of FAKE

tushartyagi:


tushartyagi:
@ErikSchierboom This one uses the `dotnet sln` approach and uses `DirectoryPath` instead of strings. Instead of picking #219, we should go ahead with this one.

ErikSchierboom:
@tushartyagi Thanks! I've fixed this in #226, which has just been merged.

Issue #218, opened 2017-03-07
Use the .NET Core compatible build of Cake


ErikSchierboom:
The work in #214 will replace our existing Fake script with Cake. However, the standard Cake executable requires the full .NET Framework, whereas this track runs on .NET Core. To make our build server integration (far) simpler, we should use the .NET Core compatible build of Cake named `Cake.CoreCLR` (which is available as a NuGet package).

This issue thus consists of two parts:

1. Replace the full .NET Framework version of Cake with the .NET Core version.
1. Update the build server configurations (which should allow us to simplify them a lot).

To help with this transition, I have been referred to two projects using the `Cake.CoreCLR` package: [Cake.Kudu](https://github.com/wcomab/Cake.Kudu) and [BitbucketPipelinesShield](https://github.com/devlead/BitbucketPipelinesShield/blob/master/.travis.yml).

Note: this issue is on-hold pending the work in #214.

ErikSchierboom:
Regarding the build server configurations, both [Travis](https://github.com/travis-ci/travis-ci/issues/7301#issuecomment-285468075) and [AppVeyor](https://github.com/appveyor/ci/issues/1179#issuecomment-285534695) have now updated their images to include the latest .NET CLI.

ErikSchierboom:
Hmmm, we'll have to update our build script a bit: https://github.com/Redth/Cake.FileHelpers/issues/10

PR #227, opened 2017-03-13
Cleanup build

ErikSchierboom:


ErikSchierboom:
FYI: the CircleCI build fails due to a Mono issue, so one more reason to work on #218 to be able to completely remove Mono.

Issue #195, opened 2017-02-10
Create test data generators


ErikSchierboom:
The canonical data of the various exercises regularly receives updates. Some of our exercises will be using outdated or incomplete data. We should try to fix this by writing test generators, support for which is added in #228.

ErikSchierboom:
- [ ] hello-world
- [ ] hamming
- [ ] gigasecond
- [ ] rna-transcription
- [ ] raindrops
- [ ] difference-of-squares
- [ ] pangram
- [ ] sieve
- [ ] roman-numerals
- [ ] robot-name
- [ ] nth-prime
- [ ] leap
- [ ] grains
- [ ] word-count
- [ ] bob
- [ ] run-length-encoding
- [ ] binary
- [ ] accumulate
- [ ] sum-of-multiples
- [ ] grade-school
- [ ] series
- [ ] phone-number
- [ ] prime-factors
- [ ] strain
- [ ] etl
- [ ] trinary
- [ ] beer-song
- [ ] bowling
- [ ] space-age
- [ ] anagram
- [ ] binary-search-tree
- [ ] crypto-square
- [ ] clock
- [ ] alphametics
- [ ] rail-fence-cipher
- [ ] acronym
- [ ] scrabble-score
- [ ] nucleotide-count
- [ ] flatten-array
- [ ] hexadecimal
- [ ] say
- [ ] meetup
- [ ] queen-attack
- [ ] palindrome-products
- [ ] bracket-push
- [ ] food-chain
- [ ] matrix
- [ ] saddle-points
- [ ] triangle
- [ ] atbash-cipher
- [ ] house
- [ ] secret-handshake
- [ ] proverb
- [ ] ocr-numbers
- [ ] pig-latin
- [ ] simple-linked-list
- [ ] luhn
- [ ] simple-cipher
- [ ] wordy
- [ ] allergies
- [ ] poker
- [ ] kindergarten-garden
- [ ] linked-list
- [ ] pythagorean-triplet
- [ ] robot-simulator
- [ ] twelve-days
- [ ] circular-buffer
- [ ] largest-series-product
- [ ] binary-search
- [ ] two-bucket
- [ ] pascals-triangle
- [ ] custom-set
- [ ] minesweeper
- [ ] scale-generator
- [ ] protein-translation
- [ ] perfect-numbers
- [ ] connect
- [ ] list-ops
- [ ] diamond
- [ ] all-your-base
- [ ] isogram
- [ ] transpose
- [ ] tournament
[...]

PR #228, opened 2017-03-13
Test generators

### LINKED ISSUE #224: Create test generators framework
Most exercises have a set of canonical data associated with them. This data can change from time to time, so the burden is on us to keep track of it. Wouldn't it be great if we could write a generator to generate the tests automatically based on the canonical data (which is described as JSON document)?

There is also an [ongoing effort](https://github.com/exercism/x-common/issues/625) to update the canonical data to a [fixed standard](https://github.com/exercism/x-common/pull/602/files). This should make it possible to write something like a base test generator.

ErikSchierboom:
This PR creates the basic infrastructure for having test generators, that is code that uses the canonical data as defined in the [x-common](https://github.com/exercism/x-common) repository to generate our C# test file. This has one huge advantage: we can very easily keep our track up-to-date with the latest canonical data, just by regenerating the tests using the generators. 

I would suggest that for the work in #195, where we verify for each exercise if the current tests match the canonical data, we start adding test generators. Note: not all exercises currently have canonical data, but a large portion of them do.

Closes #224


ErikSchierboom:
By the way, I've opted to keep things as simple as possible. No templating engine, no Roslyn syntax trees, just plain string manipulation. If anyone feels that is overly simple, let me know.

ErikSchierboom:
There are still some things that need to be added:

- Additional using statements
- Code block before the tests
- Code block after the tests

PR #243, opened 2017-03-15
HINTS added for n-th prime exercise

jmbradnan:


exercises/word-search/HINTS.md
```diff
@@ -0,0 +1,3 @@
+## Hints
+Tuples can be used to provide access to and manipulation of a data set.  For example, you can use them to pass multiple values to a method through a single parameter.
```
ErikSchierboom: Although I like the phrasing, it sort of implies that you *could* use tuples, whereas in fact you *have* to use tuples. Also, it doesn't list on of the prime use cases of tuples: returning multiple values from a function. Coincidentally, it is for returning multiple values from a function that this exercise uses tuples.

So perhaps you could rephrase this a bit to note that this exercise requires you to write a function that returns a tuple, which we use to allow us to easily return two values: the x- and y part of a coordinate.

ErikSchierboom:
Hi @jmbradnan! First of all, thanks for doing this! I have a couple of remarks:

1. Could you split this PR into two separate PR's? This has several advantages, one of them being that it becomes easier for me to review the PR. The other is that in the separate PR's version, when one file needs to have some changes, the other PR can still be merged. Grouping them all together in one PR makes this impossible.
1. This branch cannot be merged due to conflicts. How familiar are you with Git? Usually, the best option for a PR is to do a rebase based on the latest master branch of the original repository. There are several [great](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request) [articles](https://help.github.com/articles/using-git-rebase-on-the-command-line/) about rebasing, but in short it will rewrite your branch's history to make it look like you have started working on it from the latest version in the original repository. Here's how that would work for you:

- Open a command prompt in your cloned repository's directory
- Do a `git remote -v`. This should output something like this:

<pre>
exercism        https://github.com/exercism/xcsharp.git (fetch)
exercism        https://github.com/exercism/xcsharp.git (push)
origin  https://github.com/jmbradnan/xcsharp.git (fetch)
origin  https://github.com/jmbradnan/xcsharp.git (push)
</pre>

As you can see, there are two remotes: one named `origin` (which should be your fork) and one named `exercism` (or sometimes `upstream)`. 

- Then do `git fetch exercism` to retrieve the latest commits from the `exercism` remote
- Then run the following command: `git rebase exercism/master`. This will try to replay your local commits based on the remote versions. If a conflict occurs, you will be prompted to fix it. Just follow the on-screen instructions.
- In the end, when your rebase is successful, you have rewritten your local Git history. The final thing is to push these changes to your branch, which you can do with `git push --force`. Note that usually it is a very bad idea to do a force push, but in this case it is fine, as you are the only one working on your branch.

Hopefully this all makes sense, otherwise leave a message in the Gitter room.

Thanks for the help!

jmbradnan:
@ErikSchierboom Hi Erik, how do I split it into two?     That was what I was trying to do but every time I tried to make new pull request it had the earlier one in the same request.     I didn't see a way to not have my request include the earlier one.     I have been going through github documentation but I am not sure what I am doing wrong.    Do you need to fork before each change you make?

ErikSchierboom:
@jmbradnan Well, the rebase is not necessary for the split, but to resolve the conflict. In fact, it's fine if you just leave this PR as is. You can then use individual PR's if you like to add more hints :)

Good luck with the rebasing! If that turns out to be hard, another option is to just create a (or two) new PR. For that, store your HINTS.md files from this PR, do a `git checkout master`, `git fetch exercism`, `git merge exercism/master` and then create your branches with the HINTS.md files in it.

jmbradnan:
@ErikSchierboom  I am going to redo this pull request since I had been doing everything out of the same branch and it was messy.      OK to close it out?

ErikSchierboom:
Sure. Go ahead and close it

ErikSchierboom:
Thanks a lot!

PR #265, opened 2017-03-25
added Can_wrap_around_more_than_one_day to Clock

Thorocaine:
Hi.
I have added a test in the Clock problem.

You can also see the test in my [submitted solution](http://exercism.io/submissions/2a0b50aa110d4bd9b56d04897895a978) 

Thank you
P.S. This is the first time I am doing something on GitHub, so I hope I am doing this properly.

### DIFF
--- a/exercises/clock/ClockTest.cs
+++ b/exercises/clock/ClockTest.cs
@@ -111,4 +111,11 @@ public class ClockTest
         var clock2 = new Clock(38, 30);
         Assert.Equal(clock2, clock1);
     }
+
+    [Fact(Skip = "Remove to run test")]
+    public void Can_wrap_around_more_than_one_day()
+    {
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
Hi,

This looks good to me! Thanks a lot. 

Normally, the tests of an exercise should be based on its [canonical-data](https://github.com/exercism/x-common/blob/master/exercises/clock/canonical-data.json), which also has a similar test. For now, I think it's great to add this test. Eventually, we should be creating a test data generator (see #195), which automatically create tests based on the exercise's canonical data. Unfortunately, the clock exercise is a bit hard to do at the moment, due to the slightly weird test group descriptions, so we'll revisit that later.

Once again, thanks! And for a first effort on GitHub, this is perfect :) 🎉

Thorocaine:
Thank you very much

Issue #271, opened 2017-03-29
[Challenge GradeBook]  two functions to the same 


rwobben:
Hello 

I noticed that the grade and Roster have the same type definition and so do the same. 
If I put the code of Roster into Grade all tests will be green. 

Regards, 

Roelof

ErikSchierboom:
Good catch! We'll be fixing this.

rwobben:
Thanks, if I can help , ask

ErikSchierboom:
@rwobben Well, a PR would be appreciated! What I would like to see is two things:

1. Move the instance field (`private readonly School school = new School();`) to be a variable in each test (makes it easier to read).
1. The `Roster` function should not take any arguments, and should return the names of all student, sorted alphabetically. That means that the final test ("Student_names_in_each_grade_in_roster_are_sorted") should be using the `Roster` function and verifying that the return string enumerable is a sorted enumerable of *all* student names.

You can find the test's source in exercises/grade-school/GradeSchoolTest.cs. Having modified the test, you should then update the default and example implementation, which are in the same folder as the test.

rwobben:
Oke, First one is easy. 
The second can take a little more time. IM just a beginner/profient in c#. 

What do you mean with the default and example implementation ?

ErikSchierboom:
@rwobben If you look in the `exercises/grade-school` directory, you'll find the following two files:

1. `Example.cs`: this file contains a working implementation that passes all tests.
1. `GradeSchool.cs`: this file contains a default implementation, with only the minimum of methods to help the user get started. It should not contain any implementation.

Would you mind creating to separate PR's? So first the one where the instance field is moved, and then once that has merged, the bigger `Roster` change? It is always a good idea to create a PR that is as small as possible, it helps us review the code and you have a better chance of getting the changes merged sooner.

rwobben:
No problem.  So also make multiple issues.
The only things is that I use LINQ a lot to solve the exercises. So my solutions looks very different from yours.  I do not have the class SortedList in my code because it's not needed for solving it

rwobben:
One question. 

Is it allright with you I change the test to different parts 
So a arrange where all things are set up
A act part where the actual calling is done
A Assert part where the actual checking take place 

I think the test are then easier to read

ErikSchierboom:
@rwobben  Usually, I think the AAA pattern is fine, but we don't really use it a lot here. Also, there is no value returned by `Add`, so I would suggest the following:

```csharp
[...]

PR #273, opened 2017-03-30
Change roster so roster displays all students sorted

rwobben:
Here the second step. 

Roster now has no parameters and displays all students in a course. 

solves #271

### DIFF
--- a/exercises/grade-school/Example.cs
+++ b/exercises/grade-school/Example.cs
@@ -15,18 +15,16 @@ public class School
 
    public IEnumerable<string> Roster()
    {
-    var students = new List<string>(); 
-
-    foreach (var item in roster)
-    {
-        foreach (var student in item.Value)
+        var students = new List<string>();
+        
+        foreach (var item in roster)
         {
-            students.Add(student); 
-                
+            foreach (var student in item.Value)
+            {
+                students.Add(student); 
+            }
         }
-    }
-
-    return students;
+        return students;
    } 
 
     public IEnumerable<string> Grade(int grade)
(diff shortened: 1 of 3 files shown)


rwobben:
Right. that was the last step I wanted to do

rwobben:
Oke, 

Found a answer : 

Could you live with this code as a example code : 

     public IEnumerable<string> Roster()
    {
        var students = new List<string>(); 

        foreach (var item in roster)
        {
            foreach (var student in item.Value)
            {
                students.Add(student); 
                    
            }
        }

        return students;

    }

ErikSchierboom:
Sure, but you forgot to add the ordering :)

rwobben:
I did not  forget it. 
With this code all tests are already sorted. 

This test is green where the elements are not added in sorted order 

     public IEnumerable<string> Roster()
    {
        var students = new List<string>(); 

        foreach (var item in roster)
        {
            foreach (var student in item.Value)
            {
                students.Add(student); 
                    
            }
        }

        return students;

    }

ErikSchierboom:
@rwobben Ah yes, you're right! If you could update the PR with the modified Example.cs, the build should succeed again.

exercises/grade-school/Example.cs
```diff
@@ -13,7 +13,21 @@ public void Add(string student, int grade)
             roster.Add(grade, new SortedList<string> { student });
     }
 
-    public IEnumerable<string> Roster(int grade) => roster[grade];
+   public IEnumerable<string> Roster()
+   {
+    var students = new List<string>();
```
ErikSchierboom: The indentation of all lines in the `Roster` method are off. They should all be indented one tab more.

ErikSchierboom:
Okay, looking good! All that's left is to wait for the CI builds to pass.

ErikSchierboom:
Merged! Thanks! :tada:

PR #291, opened 2017-05-17
Minor fix

mattcbaker:
The value `HyperText Markup Language` is in PascalCase, the test has the name Camelcase. This just updates the test name to be PascalCase.


ErikSchierboom:
Good catch. Thanks! :tada:

PR #296, opened 2017-06-09
Rename the global exercise hints file

kytrinyx:
We've had some difficulty coming up with a good name for the file
that gets included in all of the exercise READMEs for a given track.

These are global hints, like how to run the test suite, which are
relevant to all the exercises on a track.

We started with SETUP.md in the root of the repository, then renamed
that to exercises/TRACK_HINTS.md because SETUP.md was misleading and
confusing, but then we realized that TRACK_HINTS.md was a bit ambiguous
and confusing as well.

Finally we settled on putting the file in the docs directory, since
this is user-facing documentation, and calling the file
EXERCISE_README_INSERT.md

See https://github.com/exercism/meta/issues/5 for context.


kytrinyx:
@ErikSchierboom I'm not sure how I failed the build with this one.

ErikSchierboom:
@kytrinyx Don't worry about it, that's due to some issue of the CI server.

PR #299, opened 2017-06-10
Normalize format of config file

kytrinyx:
This will let us script changes to it with less noise in the diffs.


ErikSchierboom:
Fixed by #298.

PR #300, opened 2017-06-11
Add test generator for Bob exercise

robkeim:
My first attempt at a test generator so I picked an easy one :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,7 @@ obj/
 bin/
 *.userprefs
 .idea/
+.dotnet/
+.fake/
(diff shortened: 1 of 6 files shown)


robkeim:
I'm not sure this is the right direction to go in in terms of whitespace handling given that some inputs are very long. Looking for input there.

ErikSchierboom:
@robkeim I think it is fine. Formatting the code is something that I was looking in last week. We could use Roslyn do automatically format our code.

ErikSchierboom:
Thanks! 🎉

PR #302, opened 2017-06-12
Simplify generators

ErikSchierboom:
This is the first step in refactoring the test generators. The aim is to make it really, really simple. In this PR, we default to using all the non-default properties as input properties. In a future PR I would like to get rid of the virtual methods as much as possible. Why not use constructor parameters instead, or otherwise use properties that can be set in the constructor?

I've also used used @robkeim's suggestion of losing the "Exercise" postfix for the exercises.


generators/Exercises/Acronym.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Generators.Exercises
+{
+    public class Acronym : EqualityExercise
+    {
+        public Acronym() : base("acronym")
```
robkeim: Another potential simplification would be to generate the exercise name from the Class name using reflection. The rules are pretty simple:
1. Acronym -> acronym
2. NthPrime -> nth-prime

You could argue it actually makes the code more obscure but it's some food for thought :)

ErikSchierboom: I like it! The PR has been updated. Don't merge it yet, as there is still a problem with the multi-line formatting in `beer-song`. I think that's the next thing I want to get right: figuring out how to best configure how the output is formatted.

generators/Exercises/Exercise.cs
```diff
@@ -42,40 +42,24 @@ private static void AddTestMethodUsingNamespaces(TestClass testClass)
...
 
-        protected virtual TestMethodData CreateTestMethodData(CanonicalData canonicalData, CanonicalDataCase canonicalDataCase, int index) 
-            => new TestMethodData
-                {
-                    CanonicalData = canonicalData,
-                    CanonicalDataCase = canonicalDataCase,
-                    Index = index,
-                    Options = CreateTestMethodOptions(canonicalData, canonicalDataCase, index)
-                };
+            if (testMethodData.Options.TestedMethodType == TestedMethodType.BooleanComparison)
```
robkeim: Minor: To future proof this you may want to have a switch with the two types existing today and then throw an exception for the default. That way if we need to add a new type in the future it's clear where it needs to be plumbed through to.

generators/Output/BooleanTestMethodGenerator.cs
```diff
@@ -0,0 +1,28 @@
...
+        protected override IEnumerable<string> Body
+        {
+            get
+            {
+                switch (TestMethodData.Options.TestedMethodFormat)
+                {
+                    case TestedMethodFormat.Static:
+                        return new[] { $"{Assertion}({TestedClassName}.{TestedMethodName}({Input}));" };
+                    case TestedMethodFormat.Instance:
+                        return new[] { $"var sut = new {TestedClassName}();", $"{Assertion}(sut.{TestedMethodName}({Input}));" };
```
robkeim: What is `sut`?

generators/Output/EqualityTestMethodGenerator.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        protected virtual string FormattedExpectedVariable
+        {
+            get
+            {
+                var lines = ExpectedToMultiLineString(Expected).Split('\n');
+                if (lines.Length == 1)
+                    return lines[0];
+
+                return "\n" + string.Join("\n", lines.Select(line => $"{line}"));
```
robkeim: What's the difference between what you have an `line => line`?

generators/Output/TestedMethodFormat.cs
```diff
@@ -0,0 +1,9 @@
+namespace Generators.Output
+{
+    public enum TestedMethodFormat
+    {
+        Static = 0,
```
robkeim: Minor: no need to start with 0 as it's the default

robkeim: Same below

generators/Exercises/Exercise.cs
ErikSchierboom: I like this!

generators/Output/BooleanTestMethodGenerator.cs
ErikSchierboom: It stands for "system under test" and is common jargon in some test frameworks.

generators/Output/EqualityTestMethodGenerator.cs
ErikSchierboom: Ah, it used to have tabs! I'll fix it.

generators/Output/BooleanTestMethodGenerator.cs
robkeim: Nice, you learn something new everyday!

ErikSchierboom:
@robkeim I've made some further simplifications that I think turned out quite nice :) There are two major open points left:

- Better handle large expected values (by introducing a variable)
- Better handle large strings (by concatenating them over multiple lines)
- Fix the failing tests (which are due to namespace issues)

ErikSchierboom:
@robkeim I've updated the PR and I am happy with the end result. Let me know how you feel!

robkeim:
Excellent work @ErikSchierboom, thanks for the hard work!

PR #318, opened 2017-07-09
Call configlet subcommand on CI

kytrinyx:
This changes configlet to pass a subcommand.

For now, we've released a version of configlet which handles both the old command:

    configlet path/to/track

as well as the new command:

    configlet lint path/to/track

This will let us update all the travis files to include the subcommand before we
release the version of configlet that requires the subcommand.

https://github.com/exercism/configlet/pull/23


ErikSchierboom:
Thanks!

PR #321, opened 2017-07-21
Remove redundant key/value pairs from config.json

kytrinyx:
The slug is not used anywhere. We initialize a track based on knowing the Track ID.
Since the repository is always named after the track ID, this field, too, is redundant,
as it can be inferred.

See https://github.com/exercism/meta/issues/19


ErikSchierboom:
Thanks!

PR #327, opened 2017-08-04
Implement TwoFer exercise in csharp

jpreese:
Just picked up the easiest unimplemented csharp exercise to get some exposure on how to implement exercises. 

The only thing I wasn't exactly sure on was the uuid generation in config.json. Looking at the configs history and previous exercise implementations, it doesn't look to be ever committed -- so I assume something is happening during the build/deployment. Is that correct?

### DIFF
--- a/config.json
+++ b/config.json
@@ -5,17 +5,28 @@
     {
       "uuid": "6c88f46b-5acb-4fae-a6ec-b48ae3f8168f",
       "slug": "hello-world",
-      "core": false,
+      "core": true,
       "unlocked_by": null,
       "difficulty": 1,
       "topics": [
-        "Text formatting"
+        "Strings"
+      ]
+    },
+    {
+      "uuid": "8ba15933-29a2-49b1-a9ce-70474bad3007",
+      "slug": "leap",
+      "core": true,
+      "unlocked_by": null,
+      "difficulty": 1,
+      "topics": [
+        "Integers",
+        "Control-flow (if-else statements)"
       ]
     },
     {
       "uuid": "da00f894-dbc8-485e-adba-a79d5ebee3ad",
       "slug": "bob",
-      "core": false,
+      "core": true,
       "unlocked_by": null,
       "difficulty": 1,
       "topics": [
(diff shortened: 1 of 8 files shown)


config.json
```diff
@@ -1172,6 +1172,15 @@
         "Recursion",
         "Searching"
       ]
+    },
+    {
+      "slug": "two-fer",
+      "core": true,
+      "unlocked_by": "hello-world",
```
ErikSchierboom: If we use this as a core exercise, and I think we should, this (unlocked_by) should be `null`.

config.json
```diff
@@ -1172,6 +1172,15 @@
...
+      "difficulty": 1,
+      "topics": [
+        "Strings"
```
ErikSchierboom: I would like to add the "Optional values" topic here (see other comment why)

config.json
```diff
@@ -1172,6 +1172,15 @@
         "Recursion",
         "Searching"
       ]
+    },
+    {
```
ErikSchierboom: The ordering of the core exercises is important. Could you move this exercise's entry up, I think between the `sum-of-multiples` and `raindrops` exercises.

exercises/two-fer/TwoFerTest.cs
```diff
@@ -0,0 +1,27 @@
+﻿// This file was auto-generated based on version 1.0.0 of the canonical data.
+
+using Xunit;
+
+namespace TwoFer
+{
+    public class TwoFerTest
```
ErikSchierboom: The test class looks good, but it does not match the [canonical-data](https://github.com/exercism/problem-specifications/blob/master/exercises/two-fer/canonical-data.json). As far as I can tell, this exercise is intended to introduce the concept of optional/default values. I think it is perfect for that, so I would like to ask you to use the same test cases as the canonical data, where the "no name given" test case should call `TwoFer.GetResponse()` (without an argument). The implementation could then assign the string parameter a default value of `null`.

ErikSchierboom:
Good question about the Guid! @kytrinyx Is one free to chose a guid for a new exercise?

config.json
kytrinyx: > The ordering of the core exercises is important.

We're actually reconsidering that. I need to open a discussion about it, but you can see the history over here: https://github.com/exercism/configlet/pull/34#issuecomment-320459371

kytrinyx:
> it doesn't look to be ever committed

@jpreese We recently added them, and generated (and committed) them for all the existing exercises in https://github.com/exercism/csharp/pull/320

> Is one free to chose a guid for a new exercise?

Yepp. The configlet tool has a uuid command `configlet uuid` which will generate a new one, but you could use any uuid tool that is handy (e.g. https://www.uuidgenerator.net/).

I'm trying to get everything documented properly. Would you take a look at https://github.com/exercism/docs/blob/master/language-tracks/configuration/README.md and see if it makes sense?

config.json
kytrinyx: Update: no we're not :) I misunderstood.

jpreese:
@kytrinyx yeah, that makes a lot more sense. I wasn't quite sure what core was referring to and the impact it had on the exercise, same with uuid generation -- that clears it up a bit.

@ErikSchierboom I've made the requested changes sans the exercise ordering in the config file as that approach seems to still have some discussions around it.

exercises/two-fer/Example.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace TwoFer
+{
+    public static class TwoFer
+    {
+        public static string GetResponse(string input)
```
ErikSchierboom: Please also make this parameter have a default value.

ErikSchierboom:
@jpreese Looking good! Just one small comment to fix the build.

ErikSchierboom:
Thanks a lot @jpreese ! 🎉

PR #337, opened 2017-08-13
Fix DotNet twitter handle and blog plug grammatical error

jpreese:
With markdown, the '@' symbol is used for citations, so the twitter handle is throwing markdown for a loop in the resources section.

### DIFF
--- a/docs/RESOURCES.md
+++ b/docs/RESOURCES.md
@@ -2,14 +2,14 @@
 
 ### Blogs
 * The official [.NET blog](https://blogs.msdn.microsoft.com/dotnet/) has lots of interesting C# articles, ranging from beginner to expert difficulty.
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
@jpreese Great catch! One little piece of maintainers etiquette: you usually don't merge your own PR's, unless there is an immediate bug that needs to be solved (which was the case here). Just as a FYI :)

jpreese:
@ErikSchierboom you got it! 

Anything above what I would consider "trivial" I was definitely going to request a second set of eyes. I don't even think I'd consider this "immediate", but just merged it because it was so trivial, similar to #336. 

Going forward, unless its a broken build / would cause pains for new pulls, I'll leave them up for you guys to check out.

PR #339, opened 2017-08-14
Refactored exercise generation and writing

ErikSchierboom:
I'm actually fairly happy with how this turned out :) The generators are now cleaner and the structure is better. But best of all: it is now far easier to customize the output. I think we can build on this to see if we need any more refactoring.


robkeim:
@ErikSchierboom I'll dig into this tomorrow :)

ErikSchierboom:
To see an example on how this PR enables customization, see the [Allergies](https://github.com/exercism/csharp/pull/339/files#diff-97806454e96d0b9753244e2813360b83) generator.

One area in which the generators code can still be improved, is in the parsing of data. Currently, everything is still in the format returned by JSON.net, so that means `JToken`, `JArray` and similar classes. Perhaps I can do some automatic normalization there too.

exercises/rotational-cipher/RotationalCipherTest.cs
```diff
@@ -1,25 +1,25 @@
-// This file was auto-generated based on version 1.0.0 of the canonical data.
+// This file was auto-generated based on version 1.1.0 of the canonical data.
```
robkeim: It's cool to see this get updated as the canonical data got updated and to know that it was done problematically... your hard work is paying off @ErikSchierboom :)

generators/Exercise.cs
```diff
@@ -1,23 +1,74 @@
...
+            Methods = RenderTestMethods(),
+            CanonicalDataVersion = CanonicalData.Version,
+            UsingNamespaces = GetUsingNamespaces()
+        };
+
+        private HashSet<string> GetUsingNamespaces()
+        {
+            var usingNamespaces = new HashSet<string> { "Xunit" };
+
+            foreach (var canonicalDataCase in CanonicalData.Cases.Where(canonicalDataCase => canonicalDataCase.ExceptionThrown != null))
```
robkeim: This could be simplified to:
```
if (CanonicalData.Cases.Any(canonicalDataCase => canonicalDataCase.ExceptionThrown != null))
    usingNamespaces.Add(canonicalDataCase.ExceptionThrown.Namespace);
```

generators/Exercise.cs
```diff
@@ -1,23 +1,74 @@
...
+                usingNamespaces.Add(canonicalDataCase.ExceptionThrown.Namespace);
+
+            return usingNamespaces;
         }
 
-        public string Name { get; }
-        public CanonicalData CanonicalData { get; }
+        protected virtual List<string> RenderTestMethods() => CanonicalData.Cases.Select(RenderTestMethod).ToList();
```
robkeim: You and F# have taught me to be a fan of immutability so I'd expose this as a `string[]` instead of a `List<string>` :)

generators/Output/JsonExtensions.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
```
robkeim: Does it make sense to combine this with `ConvertExtensions`?

exercises/rotational-cipher/RotationalCipherTest.cs
ErikSchierboom: :)

generators/Exercise.cs
ErikSchierboom: Well, yes and no. If there is only ever one exception thrown: yes. If not, no :)

generators/Exercise.cs
ErikSchierboom: Good point! This was a remainder of when I was manipulating the test methods :)

generators/Output/JsonExtensions.cs
ErikSchierboom: Good point, but I found out it wasn't even used! So I removed it.

ErikSchierboom:
@robkeim The PR has been updated!

generators/Program.cs
```diff
@@ -20,14 +21,29 @@ private static void SetupLogger()
                 .CreateLogger();
         }
 
-        private static void Generate(Options options)
+        private static void RegenerateTestClasses(Options options)
         {
-            Log.Information("Generating tests...");
+            Log.Information("Re-generating test classes...");
+
+            var canonicalDataParser = CreateCanonicalDataParser(options);
```
jpreese: CreateCanonicalDataParser eventually leverages Git2Sharp which can throw an exception if no network connection is found. We definitely cannot continue if the data cannot be retrieved. It might be nice to catch this instead of leaving it unhandled.

generators/Exercises/Alphametics.cs
```diff
@@ -24,7 +25,6 @@ public Alphametics()
         protected override TestClass CreateTestClass()
         {
             var testClass = base.CreateTestClass();
-            testClass.UsingNamespaces.Add(typeof(ArgumentException).Namespace);
             testClass.UsingNamespaces.Add(typeof(Dictionary<char, int>).Namespace);
```
jpreese: Totally an opinion. At first glance, I think I would want to abstract out adding Namespaces. So in Exercise.cs maybe add another override for adding additional namespaces, and then call that during the TestClass construction. I looked at some of the other tests, and I don't think adding additional namespaces (aside from exception throwing, which you've already taken care of) is going to be incredibly common anyway.

generators/Exercise.cs
```diff
@@ -1,23 +1,74 @@
...
+        }
+
+        protected virtual void UpdateCanonicalData(CanonicalData canonicalData)
+        {
+        }
+
+        public string Render() => RenderTestClass();
+
+        protected virtual string RenderTestClass() => CreateTestClass().Render();
+
```
jpreese: I might be missing the bigger picture here, of why these were split up, but it should be possible to just remove Render and update the call in ExerciseWriter so everyone is using RenderTestClass.

generators/Program.cs
ErikSchierboom: Good point. I think I'll have to catch exceptions in the Program class, to catch and log *any* exception.

generators/Exercises/Alphametics.cs
ErikSchierboom: I had this at some point, but for some reason decided against it. Will add it back!

generators/Exercise.cs
ErikSchierboom: Excellent point! I'll fix this.

generators/Exercise.cs
robkeim: Totally right... I misread this I was thinking it was a boolean of has exceptions or not.

ErikSchierboom:
@jpreese @robkeim I've updated the PR!

robkeim:
@ErikSchierboom it looks like you've cleaned the git history so that there's only one commit. In the future could you leave it so the delta of what you change after our comments is easier to review? Then we can squash before merging to ensure the final history is clean.

ErikSchierboom:
@robkeim Ah, that is still a remnant of the time when you couldn't squash from GitHub. Sorry about that.

robkeim:
@ErikSchierboom no worries :)

robkeim:
Awesome work @ErikSchierboom 👍

PR #343, opened 2017-08-17
Add test geneator for Hamming exercise

robkeim:
This is for issue #195

### DIFF
--- a/generators/Exercises/Hamming.cs
+++ b/generators/Exercises/Hamming.cs
@@ -1,0 +1,16 @@
+﻿using System;
+using Generators.Input;
+
+namespace Generators.Exercises
+{
+    public class Hamming : Exercise
+    {
+        protected override void UpdateCanonicalData(CanonicalData canonicalData)
+        {
+            foreach (var canonicalDataCase in CanonicalData.Cases)
+            {
+                canonicalDataCase.ExceptionThrown = canonicalDataCase.Expected is long ? null : typeof(ArgumentException);
+            }
+        }
+    }
+}
--- a/exercises/hamming/Example.cs
+++ b/exercises/hamming/Example.cs
@@ -3,8 +3,13 @@ using System.Linq;
 
 public static class Hamming
 {
-    public static int Compute(string firstStrand, string secondStrand)
+    public static int Distance(string firstStrand, string secondStrand)
     {
+        if (firstStrand.Length != secondStrand.Length)
+        {
+            throw new ArgumentException("strand lengths must be equal");
+        }
+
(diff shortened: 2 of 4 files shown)


generators/Exercises/Hamming.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace Generators.Exercises
+{
+    public class Hamming : Exercise
+    {
+        protected override void UpdateCanonicalData(CanonicalData canonicalData)
+        {
+            foreach (var canonicalDataCase in CanonicalData.Cases)
+            {
+                canonicalDataCase.ExceptionThrown = canonicalDataCase.Expected is long ? null : typeof(ArgumentException);
+            }
```
jpreese: I was a little curious why you chose long over int. Looking at the canonical data, I definitely would've gone int. But after debugging your branch a bit I noticed that it ultimately stems from JSON.NET parsing ints as int64 (http://json.codeplex.com/workitem/20832) -- so we do have to use long here.

jpreese:
Thanks @robkeim !

ErikSchierboom:
Great!

generators/Exercises/Hamming.cs
robkeim: @jpreese yes i could have cast to an `int` since I know the values are going to fit in the int range, but in this case I didn't think it was worth it given the only thing I'm using it for is to determine whether the test case expects an exception.

Thanks for the quick review!

jpreese: Haha, nope I was actually saying we **had** to use long here. I tried replacing `long` with `int` on a local branch, and generated test was nothing but throwing exceptions for every case because the Expected property was never true. The JSON parser parses every number (when using `object`) to `long`

I mostly pointed it out since I wasn't aware, and maybe others weren't as well :)

robkeim: Ah I see yes :) We have to deserialize it as a `long`, but afterwards we could convert the long to an int if having an int was more convenient for us with `Convert.Int32`.

It might be helpful to due in future exercises if there are more complex operations that we need to do with that deserialized value.

jpreese: For cases like this, I think I'd almost rather have an "IsNumber" or even "IsWholeNumber" if we feel the need to distinguish for future exercises. That way we're not trying to guess if we converted it to an int behind the scenes. At least this way we know that if we see a number, go `long`. 

It just gets us one step further away from guessing types and feels a little more fluent, in my opinion. Food for thought anyway.

ErikSchierboom: At the moment, I think just using `long` is sufficient, but once it gets problematic we can change things.

PR #344, opened 2017-08-18
Add CollatzConjecture exercise test generator

jpreese:
One thing I did notice after using the new generator is that UpdateCanonicalData(CanonicalData canonicalData) takes a param, but it really doesn't need to since it uses the property from the Exercise abstract anyway.

### DIFF
--- a/exercises/collatz-conjecture/CollatzConjecture.cs
+++ b/exercises/collatz-conjecture/CollatzConjecture.cs
@@ -2,7 +2,7 @@
 
 public static class CollatzConjecture
 {
-    public static int GetSteps(int input)
+    public static int Steps(int number)
     {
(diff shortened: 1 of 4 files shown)


ErikSchierboom:
@jpreese Well that's true, but it does make it harder to discover for new people creating a generator. Having the canonical data as an explicit parameter makes it easier to new people to get started IMHO.

ErikSchierboom:
Thanks @jpreese ! :tada:

jpreese:
Thats fair, it does showcase what the method is going to be working with! Maybe we should have the CanonicalData property be private and force the tests to use the parameter?

ErikSchierboom:
Maybe we should indeed.

PR #346, opened 2017-08-18
Add generator for gigasecond exercise

robkeim:
Another exercise for #195


ErikSchierboom:
Looking good! Too bad that the canonical data uses a string to specify the date :)

robkeim:
@ErikSchierboom agreed although even if they had specified it in something like a ticks format we still would have need to output it in a human readable format for the tests so it wouldn't save us that much work.

PR #348, opened 2017-08-18
[WIP] Add Run Length Encoding test generator

robkeim:
This is for issue #195

### DIFF
--- a/exercises/run-length-encoding/RunLengthEncodingTest.cs
+++ b/exercises/run-length-encoding/RunLengthEncodingTest.cs
@@ -1,60 +1,84 @@
-﻿using Xunit;
+// This file was auto-generated based on version 1.0.0 of the canonical data.
+
+using Xunit;
 
 public class RunLengthEncodingTest
 {
     [Fact]
-    public void Encode_simple()
+    public void Encode_empty_string()
     {
-        const string input = "AABBBCCCC";
-        const string expected = "2A3B4C";
-        Assert.Equal(expected, RunLengthEncoding.Encode(input));
+        Assert.Equal("", RunLengthEncoding.Encode(""));
     }
 
     [Fact(Skip = "Remove to run test")]
-    public void Decode_simple()
+    public void Encode_single_characters_only_are_encoded_without_count()
     {
-        const string input = "2A3B4C";
-        const string expected = "AABBBCCCC";
-        Assert.Equal(expected, RunLengthEncoding.Decode(input));
+        Assert.Equal("XYZ", RunLengthEncoding.Encode("XYZ"));
     }
 
     [Fact(Skip = "Remove to run test")]
-    public void Encode_with_single_values()
+    public void Encode_string_with_no_single_characters()
     {
-        const string input = "WWWWWWWWWWWWBWWWWWWWWWWWWBBBWWWWWWWWWWWWWWWWWWWWWWWWB";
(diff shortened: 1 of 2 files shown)


exercises/run-length-encoding/RunLengthEncodingTest.cs
```diff
@@ -1,60 +1,84 @@
...
     {
-        const string input = "⏰⚽⚽⚽⭐⭐⏰";
-        const string expected = "⏰3⚽2⭐⏰";
-        Assert.Equal(expected, RunLengthEncoding.Encode(input));
+        Assert.Equal("aabbbcccc", RunLengthEncoding.Decode("2a3b4c"));
     }
 
     [Fact(Skip = "Remove to run test")]
-    public void Decode_unicode()
+    public void Consistency_encode_followed_by_decode_gives_original_string()
```
robkeim: @ErikSchierboom this test case is to verify that `x == Decode(Encode(x))`., but the way that I have it now obviously isn't working. Any thoughts on how that should be handled in the generator?

jpreese:
@robkeim this is how I would approach it with the current generator framework.

```csharp
        protected override string RenderTestMethodBodyAssert(TestMethodBody testMethodBody)
        {
            if(testMethodBody.CanonicalDataCase.Property == "consistency")
            {
                return RenderConsistencyToAssert(testMethodBody);
            }

            return base.RenderTestMethodBodyAssert(testMethodBody);
        }

        private static string RenderConsistencyToAssert(TestMethodBody testMethodBody)
        {
            const string template = @"Assert.Equal(""{{ExpectedOutput}}"", {{ExerciseName}}.Encode({{ExerciseName}}.Decode(""{{ExpectedOutput}}"")));";

            var templateParameters = new
            {
                ExpectedOutput = testMethodBody.CanonicalDataCase.Expected,
                ExerciseName = testMethodBody.CanonicalData.Exercise.ToTestedClassName()
            };

            return TemplateRenderer.RenderInline(template, templateParameters);
        }
```

robkeim:
@jpreese great work that looks good to me! I've updated the PR with a shameless copy/paste from what you wrote with one small tweak of `Encode` being executed before `Decode`.

robkeim:
Also please tell me you had to check out the code to work through that and that you just didn't write it directly here in GitHub's online interface :)

jpreese:
Nah, I grabbed your branched and toyed around with it a little in Visual Studio. Though that would be impressive =P

ErikSchierboom:
More great work! 🎉 

About this generator, do you feel this is to much to write? I could also create a method that allows you to only override the expected value or the calling of the sut. Would that be better?

robkeim:
Thanks @ErikSchierboom! I think you did a great job and these have been really easy to write, although not so easy to wrap my head around the meta programming idea of writing a program to write a program... :)

This is the second case where we've needed an override like this right? I'd say if we have a third exercise that follows into the same pattern then it's probably adding a refactoring for this into the backlog.

ErikSchierboom:
@robkeim Correct. Let's see how the rest turns out then :) I'm particularly interested in the generators for code that uses state. For example, the code to test the `robot-simulator` class.

ErikSchierboom:
Oh and by the way, you have no idea how I struggled with writing a program to write a program :D It has taken many, many iterations to get to this somewhat acceptable format :D

PR #349, opened 2017-08-18
Add test generator for Pangram exercise

robkeim:
This is for #195

### DIFF
--- a/generators/Exercises/Pangram.cs
+++ b/generators/Exercises/Pangram.cs
@@ -1,0 +1,17 @@
+﻿using Generators.Input;
+
+namespace Generators.Exercises
+{
+    public class Pangram : Exercise
+    {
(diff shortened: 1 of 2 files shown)


generators/Exercises/Pangram.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Generators.Exercises
+{
+    public class Pangram : Exercise
+    {
+        protected override void UpdateCanonicalData(CanonicalData canonicalData)
+        {
+            foreach (var canonicalDataCase in CanonicalData.Cases)
+            {
+                canonicalDataCase.Description = canonicalDataCase.Description.Replace("'", string.Empty);
```
ErikSchierboom: I should fix this in the test method name converter.

PR #351, opened 2017-08-18
Add test generator for prime factors exercise

robkeim:
Another one for #195

### DIFF
--- a/generators/Exercises/PrimeFactors.cs
+++ b/generators/Exercises/PrimeFactors.cs
@@ -1,0 +1,6 @@
+﻿namespace Generators.Exercises
+{
+    public class PrimeFactors : Exercise
+    {
+    }
+}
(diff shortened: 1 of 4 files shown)


generators/Exercises/PrimeFactors.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Generators.Exercises
+{
+    public class PrimeFactors : Exercise
```
ErikSchierboom: I love it when the generator can look like this 😎

robkeim: Me too 👍

jpreese:
Eek, this PR removed all of the skips in the test suite.

robkeim:
Good catch @jpreese, thanks for fixing this! I'm working in Visual Studio which doesn't allow me to ignore the skips so I have to remove them to validate our sample solution is working. I forgot to re-run the generator in this case.

PR #353, opened 2017-08-18
Add test generator for rna transcription exercise

robkeim:
And another for #195

### DIFF
--- a/exercises/rna-transcription/Example.cs
+++ b/exercises/rna-transcription/Example.cs
@@ -1,7 +1,7 @@
 using System.Collections.Generic;
 using System.Linq;
 
-public class Complement
+public static class RnaTranscription
 {
     private static readonly Dictionary<char, char> DnaToRna = new Dictionary<char, char>
         {
--- a/exercises/rna-transcription/Example.cs
+++ b/exercises/rna-transcription/Example.cs
@@ -8,8 +8,13 @@ public class Complement
             { 'G', 'C' }, { 'C', 'G' }, { 'T', 'A' }, { 'A', 'U' }
         };
 
-    public static string OfDna(string nucleotide)
+    public static string ToRna(string nucleotide)
     {
+        if (nucleotide.Any(x => !DnaToRna.ContainsKey(x)))
+        {
+            return null;
+        }
+
         return string.Concat(nucleotide.Select(x => DnaToRna[x]));
     }
 }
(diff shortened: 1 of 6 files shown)


exercises/rna-transcription/Example.cs
```diff
@@ -8,8 +8,13 @@ public class Complement
             { 'G', 'C' }, { 'C', 'G' }, { 'T', 'A' }, { 'A', 'U' }
         };
 
-    public static string OfDna(string nucleotide)
+    public static string ToRna(string nucleotide)
```
ErikSchierboom: The build fails because this (example) class' name has not been renamed to `RnaTranscription`.

exercises/rna-transcription/RnaTranscriptionTest.cs
```diff
@@ -1,34 +1,54 @@
...
     public void Rna_complement()
     {
-        Assert.Equal("UGCACCAGAAUU", Complement.OfDna("ACGTGGTCTTAA"));
+        Assert.Equal("UGCACCAGAAUU", RnaTranscription.ToRna("ACGTGGTCTTAA"));
+    }
+
+    [Fact(Skip = "Remove to run test")]
+    public void Correctly_handles_invalid_input_rna_instead_of_dna_()
+    {
+        Assert.Equal(null, RnaTranscription.ToRna("U"));
```
ErikSchierboom: Should we return `null` or throw an exception?

generators/Exercises/RnaTranscription.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Generators.Exercises
+{
+    public class RnaTranscription : Exercise
+    {
+        protected override void UpdateCanonicalData(CanonicalData canonicalData)
+        {
+            foreach (var canonicalDataCase in CanonicalData.Cases)
+            {
+                canonicalDataCase.Expected = canonicalDataCase.Expected ?? new UnescapedValue("null");
```
ErikSchierboom: This is not ideal and should not be necessary. A `null` value should be displayed as `null`. I'll try and fix that.

exercises/rna-transcription/RnaTranscriptionTest.cs
robkeim: Either way has it's pros and cons, but now that I think about other exercises they're mostly throwing exceptions, so for consistency it probably makes sense to head in that direction.

exercises/rna-transcription/Example.cs
robkeim: Thanks, I fixed that and updated the PR.

robkeim:
@ErikSchierboom I fixed this generator, could you take a look when you get a chance?

ErikSchierboom:
Thanks! :tada

PR #357, opened 2017-08-18
Update canonical data to latest version

### LINKED ISSUE #347: Generator does not read latest canonical data
I recently opened a PR over in the problem-specifications repo to fix the TwoFer exercise property name (https://github.com/exercism/problem-specifications/pull/881). I was then going to use the generator to update its tests, but I couldn't get it to grab the latest canonical data.

For the generator, it looks like we're doing a Fetch(), but not actually pulling the data down into the local .problem-specifications repository, which is where its trying to read the canonical-data file from.

ErikSchierboom:
Fixes #347


jpreese:
awww yeah! Thanks @ErikSchierboom

Issue #356, opened 2017-08-18
Improve test generators


ErikSchierboom:
- [x] Render `null` as `null` without requiring wrapping in an unescaped values
- [x] Use `Assert.Null` if using `Assert.Equal(null, x)`
- [x] Don't render quote character in test method/class names

robkeim:
Nice work @jpreese getting these issues fixed! Are there any other issues we've discovered so far that we should be tracking here, or should we go ahead and close this issue?

ErikSchierboom:
I think we can close.

jpreese:
Yeah it depends if we want to keep this as a rolling issue and keep adding to it, or just close it out and open PRs as we find stuff. I'm a fan of the latter since I don't really like issues that just never have a clear finish line.

I have a couple other ideas, but I was just going to do PRs for them, so I'll go ahead and close this.

PR #359, opened 2017-08-18
Use Assert.Null when validating against null

jpreese:
Part of #356

### DIFF
--- a/generators/Exercise.cs
+++ b/generators/Exercise.cs
@@ -78,7 +78,7 @@ namespace Generators
                 return new TestMethodBodyWithBooleanCheck(canonicalDataCase, CanonicalData);
             }
 
-            if(canonicalDataCase.Expected is null)
+            if (canonicalDataCase.Expected is null)
             {
                 return new TestMethodBodyWithNullCheck(canonicalDataCase, CanonicalData);
             }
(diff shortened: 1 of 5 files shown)


generators/Exercise.cs
```diff
@@ -69,10 +69,19 @@ protected virtual string RenderTestMethodBody(CanonicalDataCase canonicalDataCas
...
+            {
                 return new TestMethodBodyWithExceptionCheck(canonicalDataCase, CanonicalData);
+            }
 
             if (canonicalDataCase.Expected is bool)
+            {
                 return new TestMethodBodyWithBooleanCheck(canonicalDataCase, CanonicalData);
+            }
+
+            if(canonicalDataCase.Expected is null)
```
robkeim: The Type A in me is screaming about a missing space after the `if` ;)

robkeim: I forgot that I can fix it inline :)

robkeim:
Nice work @jpreese! The only thing I wonder about this is should we be avoiding `null` at all and just using exceptions. I'll create an issue to discuss it further.

generators/Exercise.cs
jpreese: Oh man, that would bother me too. I didn't even notice that. Usually Visual Studio just automatically tosses that in for me.

robkeim:
Thanks @jpreese 🎉

ErikSchierboom:
> The only thing I wonder about this is should we be avoiding null at all and just using exceptions. I'll create an issue to discuss it further.

👍 I think we should definitely have some discussion on this issue.

ErikSchierboom:
@robkeim Issue created: https://github.com/exercism/csharp/issues/366

PR #362, opened 2017-08-19
Add test generator for say exercise

robkeim:
For issue #195

### DIFF
--- a/exercises/say/SayTest.cs
+++ b/exercises/say/SayTest.cs
@@ -1,6 +1,7 @@
-﻿// This file was auto-generated based on version 1.0.0 of the canonical data.
+// This file was auto-generated based on version 1.0.0 of the canonical data.
 
 using Xunit;
+using System;
 
 public class SayTest
 {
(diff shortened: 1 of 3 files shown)


exercises/say/SayTest.cs
```diff
@@ -1,101 +1,96 @@
...
-        Assert.Equal("nine hundred eighty-seven billion six hundred fifty-four million three hundred twenty-one thousand one hundred twenty-three", Say.InEnglish(987654321123L));
+        Assert.Equal("nine hundred eighty-seven billion six hundred fifty-four million three hundred twenty-one thousand one hundred twenty-three", Say.InEnglish(987654321123));
     }
 
     [Fact(Skip = "Remove to run test")]
-    public void Lower_bound()
+    public void Numbers_below_zero_are_out_of_range()
     {
-        Assert.Throws<ArgumentOutOfRangeException>(() => Say.InEnglish(-1L));
+        Assert.Equal("-1", Say.InEnglish(-1));
```
ErikSchierboom: I don't really like the change to -1 instead of throwing an exception. Was that on purpose or just to conform to the canonical data? If it is the latter, you should know that some canonical data exercises use -1 to signal errors, some use `null` and other use an `error` object. In this case, I think we should keep the exception throwing.

robkeim: I did this because it was what the canonical data had... I'm going to wait until we close on #366 before pushing a new iteration of this PR.

ErikSchierboom: I've added my thoughts to #366.

ErikSchierboom:
Thanks! 🎉

PR #363, opened 2017-08-19
Add test generator for scrabble score

robkeim:
For issue #195

### DIFF
--- a/exercises/scrabble-score/Example.cs
+++ b/exercises/scrabble-score/Example.cs
@@ -1,7 +1,7 @@
 using System.Collections.Generic;
 using System.Linq;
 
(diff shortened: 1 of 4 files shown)


exercises/scrabble-score/Example.cs
```diff
@@ -1,7 +1,7 @@
 using System.Collections.Generic;
 using System.Linq;
 
-public class Scrabble
+public class ScrabbleScore
```
ErikSchierboom: The class can be `static`.

robkeim: Good point @ErikSchierboom, fixed!

PR #365, opened 2017-08-19
Have two-fer core exercise unlock other exercises

ErikSchierboom:
The `two-fer` core exercise didn't unlock any non-core exercise. This PR fixes that.

### DIFF
--- a/config.json
+++ b/config.json
@@ -47,17 +47,6 @@
       ]
     },
     {
-      "uuid": "57f02d0e-7b75-473b-892d-26a7d980c4ce",
-      "slug": "two-fer",
-      "core": true,
-      "unlocked_by": "null",
-      "difficulty": 1,
-      "topics": [
-        "Strings",
-        "Optional values"
-      ]
-    },
-    {
       "uuid": "29ae7f8e-a009-4175-9350-a8c684c89730",
       "slug": "raindrops",
       "core": true,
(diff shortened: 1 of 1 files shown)


robkeim:
@ErikSchierboom do we want to keep `hello-world` as a core exercise or have `two-fer` replace it? When I implemented this exercise in the F# track I used `two-fer` to replace another core exercise.

Here's the F# PR for reference:
https://github.com/exercism/fsharp/pull/340

ErikSchierboom:
@robkeim Yes I tried to replace one of the core exercises, but found no good option. Do you have a suggestion perhaps?

jpreese:
I think I like adding it as a new core exercise. It adds an entirely new concept (optional parameterss), and keeps people invested in the track longer.

I'm still pretty new to the Exercism game, but is there an approach we take for what is/isn't a core exercise? Is there anything wrong with having a large amount? Obviously you don't want everything to be core, but if we're trying to teach something new and not just challenge the student, seems valid to me.

I think core to me says.. a new concept is introduced. You wouldn't have two core exercises that use the same concepts, just arranged differently -- different logic.

ErikSchierboom:
> Is there anything wrong with having a large amount?

There is a rough guideline that each core exercise should unlock about 5-6 non-core exercises, which leads us to around the current amount of core exercises.

robkeim:
@ErikSchierboom no I didn't have something in mind, I was just wondering if you had purposely not replaced a core track item or not.

This looks good to me, I'll go ahead and merge, thanks!

Issue #366, opened 2017-08-19
Working with null


ErikSchierboom:
How to handle `null` values? In general, they are widely used in C#, but are also a cause of pain due to the infamous null-reference exceptions. I think we should in general use exceptions when the input is invalid, as that is also what the .NET framework seems to do in most cases (int.Parse, etc.). We should use `null` when we want to indicate that no result could be found (e.g. when querying a collection for a value).

If we were to adopt this, that means that the [phone-number tests](https://github.com/jpreese/csharp/blob/3ab86cf34a8e8c55ccc585611fdac024d8ccbcad/exercises/phone-number/PhoneNumberTest.cs) should be using an exception instead of checking for null.

robkeim:
@ErikSchierboom we started discussing this a bit in #360, I shouldn't have closed it before letting you chime in.

In #362 I discovered that there are actually three different ways the canonical data are handling "invalid/error" cases:
- null
- adding an error with description
- invalid value (such as -1 for numbers)

ErikSchierboom:
Correct. For many exercises, the comments explicitly state that each track for themselves should decide how to treat invalid values. This makes sense as most functional languages would replace null with an option type, which you can't easily encode in JSON. Therefore, for each exercise we should determine what to do. The only thing the canonical data gives us is an indication *if* there is an error, an invalid value or a missing value. We as track maintainers should then decide how to encode it.

In #360 @jpreese said:

> So my opinion is that the canonical-data should be the source of truth. Whatever is there, is what we render.

It would be great if we could do this, but unfortunately the canonical data is far from standarized and uses several ways to encode failure. In my opinion we should thus feel empowered to decide for ourselves what to do with failing input.

To make my point, the comments of the `say` exercise state:

> "Here -1 is used as expected value to indicate that the",
> "input value is out of the range described in the exercise."

-1 is thus *used* as the expected value, but it is merely an indicator of an out of range value. It doesn't actually state that -1 should be the expected return value.

jpreese:
@ErikSchierboom so I have two approaches I would take on this then.

1) We update the canonical-data to be consistent. I think if the data isn't valid, an exception should be thrown. My gut says it would be better to be consistent. I assume the only reason why some canonical data is -1, others are null, others are exceptions is because its a community effort. Different contributors, different reviewers. It is a little heavy handed though.

[...]

PR #371, opened 2017-08-19
Use exceptions everywhere instead of null

robkeim:
After our discussion in #366 I've migrated what I think are the remaining test generated exercises to throw exceptions.

I didn't remove the `Assert.Null` method @jpreese added because we may want to use it at some point.


ErikSchierboom:
Brilliant! 

> I didn't remove the Assert.Null method @jpreese added because we may want to use it at some point.

Excellent idea.

PR #379, opened 2017-08-21
Add support for Dictionary<string,int>

jpreese:
Was trying to implement the word-count generator and noticed the results looked similar to calling ToString() on the resulting Dictionary object. After some debugging I noticed we explicity check for <char,int> inside of ValueFormatter.cs -- in this case <string,int> is needed.

I'm not a huge fan of the duplication and checking specifically for types, but outside of doing a lot of reflection work I'm not sold on an alternative approach right now.


ErikSchierboom:
@jpreese Yes, that's one of the less pretty parts of the generator setup.

jpreese:
It's pretty in my eyes, @ErikSchierboom

ErikSchierboom:
Thanks! 🎉

robkeim:
@jpreese I don't see another way to do this without reflection either. I think it might be worth exploring a reflection solution if we end up having many different types of dictionaries or other "repetitive" data structures.

PR #380, opened 2017-08-21
[WIP] Add StyleCop to build

jpreese:
Ok so, I marked this as WIP as we should probably have some discussions around what rules we want to actually enforce. Enforcing and ignoring rules is pretty trivial going forward from this point, so as we get a list build, we can just modify the ruleset!

List of rules can be found [here](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/tree/master/documentation)

The ones I have in there now are just for testing purposes. If you wish to checkout this branch yourself and give some rules a try, its as simple as tossing in a Rule with an Id, and then assigning an Action (e.g. Hidden, Error, Warning)

### DIFF
--- a/Cake.ruleset
+++ b/Cake.ruleset
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<RuleSet Name="Cake Ruleset" ToolsVersion="12.0">
+
+  <Rules AnalyzerId="StyleCop.Analyzers" RuleNamespace="StyleCop.Analyzers">
+    <Rule Id="SA1202" Action="None" />
+    <Rule Id="SA1649" Action="None" />
+    <Rule Id="SA1633" Action="None" />
+    <Rule Id="SA1600" Action="None" />
+    <Rule Id="SA0001" Action="None" />
+  </Rules>
+
+</RuleSet>
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -1,0 +1,12 @@
+<Project>
+
+  <!-- StyleCop -->
+  <PropertyGroup>
+    <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)Cake.ruleset</CodeAnalysisRuleSet>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="StyleCop.Analyzers" Version="1.1.0-beta001" PrivateAssets="All" />
+  </ItemGroup>
+
+</Project>


ErikSchierboom:
Well actually, what I meant with this issue is to add the analyzers to the exercise projects. My aim was to help the user guide towards building code that conforms to the .NET guidelines. It would be quite useful if people could get an IDE or command-line hint that classes should use PascalCasing.

But that doesn't prevent this PR from also being useful as it is, as we should also be conforming to (most) standards ourselves.

I'll play a bit with it sometime soon.

jpreese:
Oh! I see now. Yeah I actually approached this problem with trying NOT to update all of the project files, but I was thinking it was going to be mostly used for us. 

I think your approach makes a ton of sense from a learning perspective and it benefits any newcomer, seasoned programmer or not, because the coding standards can differ from language to language.

So right now at least, this will accomplish half of that since this does provide warnings within the IDE, but you have to have the Directory.Build.targets.. which obviously you won't have when you fetch from the exercism cli.

Let me do some refactoring on this a bit.

robkeim:
@jpreese I had understood the same thing that you did from the original issue :) Nice work on this, I had played around with it but didn't have any luck :(

In terms of rules for us I think for us, I'm in favor of most of the StyleCop rules as they enforce consistency, however, I've found some of the documentation rules to be pretty heavy handed from previous experience. I had one job where we had to use XML comments for every method parameter of every method and it ended up with a lot of copy/pasted documentation that was out of date. If we're going to enable them for us though I'd want non-complaint code to fail the build otherwise we might wind up in a situation where we just have tons of warnings that no one is even looking at anymore.

In terms of rules for the students I think we need to be careful not to overwhelm newcomers. I agree that having these rules would be a great way for people to improve their C# style and comply with best practices. I think in any case we'd probably want to have an "opt in" approach particularly if we were going to fail the builds because of rules as not everyone is going to want that extra level of challenge. We might also have a tiered approach to rules (beginner, intermediate, and expert) where the rules got more strict in complying with C# coding standards.

ErikSchierboom:
I think we should only guide users, not have builds fails explicitly. Therefore, we should only enable rules on the "Hint" level I think, and only a subset of the existing rules. The documentation rules you mentioned are not useful in the context of Exercism. Therefore I think we should hand-pick a specific set of rules we want to ignore and have the rest of the rules be "Hint".

jpreese:
So to clarify: we want to apply rule sets to all projects, and have the rules be hints? 

We'll just have to be cognizant of warnings when checking the results of Travis (since they won't be errors).

ErikSchierboom:
@jpreese Correct!

ErikSchierboom:
@jpreese I was looking at the StyleCop analyzers, but do they also provide code fixes? It appears that the code analysis code I referred to in the original issue does contain fixes: https://github.com/dotnet/roslyn/pull/18285/files

jpreese:
I don't believe so. At least in my experiences with StyleCop, I've never used it to fix up code automagically. It's always been shown as an IDE hint or during the build.

ErikSchierboom:
Hmmm, maybe it is worth checking out the Roslyn analyzers then.

robkeim:
I've also never heard of StyleCop providing code fixes, but I was using it before Rosyln analyzers existed. If we can get linting and automatic fixes that seems like a great buy 1 get one 1 opportunity.

jpreese:
I'm going to go ahead and close this out and look into alternatives in regards to product and approach. No need to keep refactoring over this PR.

ErikSchierboom:
@jpreese Good idea.

PR #383, opened 2017-08-22
Add poker generator

jpreese:
Whew.. apparently the poker canonical-data added a lot more test cases, and our Example.cs did not pass all of the new tests. Most having to deal with using more than one deck.

So this commit adds the generator as well as updates the Example.cs so that it passes all of the new tests.


ErikSchierboom:
Great work!

PR #384, opened 2017-08-22
Escape backslash in formatter

jpreese:


robkeim:
Nice work @jpreese! It's cool to see this generator simplified even further!

PR #385, opened 2017-08-22
Clean up Generator project

jpreese:
- Removes the mapped lib/ directory from Lib2GitSharp
- Removes the user specific config file and adds it to the gitignore

### DIFF
--- a/generators/.gitignore
+++ b/generators/.gitignore
@@ -1,2 +1,3 @@
 .problem-specifications/
 Properties/launchSettings.json
+*.csproj.user
--- a/generators/Generators.csproj.user
+++ b/generators/Generators.csproj.user
@@ -1,6 +1,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <PropertyGroup>
-    <ShowAllFiles>false</ShowAllFiles>
-  </PropertyGroup>
-</Project>


generators/Generators.csproj
```diff
@@ -15,4 +15,9 @@
   <ItemGroup>
     <EmbeddedResource Include="Output\Templates\*.liquid" />
   </ItemGroup>
+  <ItemGroup>
```
ErikSchierboom: Why did you add this item group? Was there anything specific going wrong?

jpreese: Hm.. on my IDE I was getting a big red x next to the lib/ directory as well as seeing the lib2git DLL. You can see my issue here, but unfortunately the opener just closed it without much resolve.

https://github.com/libgit2/libgit2sharp/issues/1425

ErikSchierboom: @jpreese I'll take a look into this tomorrow

jpreese: @ErikSchierboom yeah I'm not happy with this. 

I just came back to this and it looks like my red x's are back, even with this change. I'll need to dig into the root cause some more to find out why these are popping up.

ErikSchierboom: @jpreese Okay, good luck! I haven't encountered those in VS2017.

jpreese:
My gut says this is something related specifically to my machine and not necessarily a fault of the csproj as I originally thought. I'll hold out to see what the Lib2Git guys say. 

In the meantime, I've updated this branch to only include the user settings removal.

ErikSchierboom:
Thanks!

jpreese:
Not sure why I'm a special flower, but the maintainers said they plan on changing how lib/ gets added to projects. So this should be a non issue when they release the newest version.
 (https://github.com/libgit2/libgit2sharp/issues/1425#issuecomment-324350369)

ErikSchierboom:
Awesome! Well done!

PR #386, opened 2017-08-22
Use canonical data parameter

jpreese:
This is something that @ErikSchierboom and I briefly touched upon in #344. 

Currently CanonicalData is directly accessible to the derived classes. Our UpdateCanonicalData method passes a parameter, but most generators reference the property directly. This privatizes that property and forces the Update method to leverage the parameter.

### DIFF
--- a/generators/Exercise.cs
+++ b/generators/Exercise.cs
@@ -11,7 +11,7 @@ namespace Generators
         
         public string Name => GetType().ToExerciseName();
 
-        protected CanonicalData CanonicalData { get; private set; }
+        private CanonicalData CanonicalData { get; set; }
         
         public void Regenerate(CanonicalData canonicalData)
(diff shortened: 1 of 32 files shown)


generators/Exercise.cs
```diff
@@ -11,7 +11,7 @@ public abstract class Exercise
         
         public string Name => GetType().ToExerciseName();
 
-        protected CanonicalData CanonicalData { get; private set; }
+        private CanonicalData CanonicalData { get; set; }
```
ErikSchierboom: Could you make it `private CanonicalData _canonicalData` and move it to the top of the class to conform to the rest of the code?

ErikSchierboom:
@jpreese Just waiting for the CI build to complete, but looking good!

Issue #303, opened 2017-06-13
Reduce local build and test time


robkeim:
Now that we have separate csproj's per problem our test and build time has grown quite a bit:
```
========================================
Default
========================================

Task                           Duration
---------------------------------------------------
Clean                          00:00:00.1835870
CopyExercises                  00:00:01.1743155
RestoreNugetPackages           00:00:40.1306513
BuildStubImplementations       00:01:09.1509081
EnableAllTests                 00:00:00.0993630
TestRefactoringProjects        00:00:14.7734888
ReplaceStubWithExample         00:00:00.1808265
TestUsingExampleImplementation 00:10:54.2394493
Default                        00:00:00.7327512
---------------------------------------------------
Total:                         00:13:00.6653407
```

The `TestUsingExampleImplementation` is taking the bulk of the time although all of the sub-tasks that it is doing are parallelizable. Running them in parallel could significantly reduce the build time.

robkeim:
Maybe I'm missing something here as in https://github.com/exercism/xcsharp/issues/216 it states the builds are now run in parallel. Is there something I need to configure to enable the parallelization?

ErikSchierboom:
Well, they don't actually run in parallel I think. We could do that rather easily I think by modifying the build script. The problem is that you cannot run the dotnet test command on a solution file. That is possible for restoring and building, but not for testing unfortunately.

jpreese:
Closing!

robkeim:
Thanks for your work on this @jpreese! 🎉

I was looking at the build and after your changes and this is the current state:
```
Task                           Duration
---------------------------------------------------
Clean                          00:00:01.0026434
CopyExercises                  00:00:06.8279400
RestoreNugetPackages           00:00:26.6270243
EnableAllTests                 00:00:00.1144599
TestRefactoringProjects        00:00:08.2498699
ReplaceStubWithExample         00:00:00.1723076
TestUsingExampleImplementation 00:03:01.3678156
Default                        00:00:00.0064392
---------------------------------------------------
Total:                         00:03:44.3684999
```

The time is getting eaten up by `RestoreNugetPackages` (~30s) and `TestUsingExampleImplementation` (~3m).

I looked into each of their settings briefly but couldn't find any settings to significantly speed up either of their execution times unfortunately :(

```
var dotNetCoreRestoreSettings = new DotNetCoreRestoreSettings
{
    DisableParallel = false,
    DiagnosticOutput = false,
    NoCache = false
};
```

[...]

PR #387, opened 2017-08-22
Run builds and tests in parallel

jpreese:
For #303 

I couldn't find any native support within dotnet core or Cake for running things in parallel -- but since we're using Cake we have access to the TPL.

Our average build time was around 17-18minutes from what I could see in previous TravisCI builds, this drops it down to around 4 minutes on my machine.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -95,9 +95,13 @@ Task("TestUsingExampleImplementation")
     .IsDependentOn("ReplaceStubWithExample")
     .Does(() => {
         DotNetCoreBuild(allSln, dotNetCoreBuildSettings);
+        var parallelOptions = new ParallelOptions 
+        {
+            MaxDegreeOfParallelism = System.Environment.ProcessorCount
+        };
 
         var allProjects = GetFiles(buildDir + "/*/*.csproj");
-        Parallel.ForEach(allProjects, (project) => DotNetCoreTest(project.FullPath, dotNetCoreTestSettings));
+        Parallel.ForEach(allProjects, parallelOptions, (project) => DotNetCoreTest(project.FullPath, dotNetCoreTestSettings));
     });
 
 Task("Default")
--- a/build.cake
+++ b/build.cake
@@ -1,5 +1,6 @@
 using System.IO;
 using System.Text.RegularExpressions;
+using System.Threading.Tasks;
 
 var target = Argument("target", "Default");
 
@@ -10,7 +11,16 @@ var defaultSln     = buildDir + "/Exercises.Default.sln";
 var allSln         = buildDir + "/Exercises.All.sln";
 var refactoringSln = buildDir + "/Exercises.Refactoring.sln";
 
-var dotNetCoreBuildSettings = new DotNetCoreBuildSettings { NoIncremental = true };
+var dotNetCoreMSBuildSettings = new DotNetCoreMSBuildSettings
+{
+    MaxCpuCount = 0
+};
+
+var dotNetCoreBuildSettings = new DotNetCoreBuildSettings 
+{ 
+    NoIncremental = true,
+    MSBuildSettings = dotNetCoreMSBuildSettings 
+};
 
 Task("Clean")
     .Does(() => {
@@ -64,9 +74,7 @@ Task("TestRefactoringProjects")
             + GetFiles(buildDir + "/*/Ledger.csproj")
             + GetFiles(buildDir + "/*/Markdown.csproj");
 
-        foreach (var refactoringProject in refactoringProjects) {
-            DotNetCoreTest(refactoringProject.FullPath);
-        }
+        Parallel.ForEach(refactoringProjects, (project) => DotNetCoreTest(project.FullPath));
 });
 
 Task("ReplaceStubWithExample")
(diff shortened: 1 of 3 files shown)


jpreese:
Hrm. I don't think Travis liked that. I'm going to leave it here as to brew up discussion.. but right now I'm not sure why it hung.

robkeim:
Nice work @jpreese! These changes look good to me, I'll let @ErikSchierboom take a look before merging in case he has any concerns that I didn't think of.

jpreese:
Also this might seem like a silly question, but do we need to build the Default and Refactoring solutions? They're a subset of the All solution, so that should catch any problems I'd think.

build.cake
```diff
@@ -91,10 +104,12 @@ Task("TestUsingExampleImplementation")
         DotNetCoreBuild(allSln, dotNetCoreBuildSettings);
 
         var allProjects = GetFiles(buildDir + "/*/*.csproj");
+        var parallelOptions = new ParallelOptions 
+        {
+            MaxDegreeOfParallelism = 16
```
ErikSchierboom: Why set this explicitly? I assume the TPL will automatically set this to a sensible option depending on your machine's configuration.

build.cake
```diff
@@ -10,7 +11,21 @@ var defaultSln     = buildDir + "/Exercises.Default.sln";
...
+
+var dotNetCoreBuildSettings = new DotNetCoreBuildSettings 
+{ 
+    NoIncremental = true,
+    MSBuildSettings = dotNetCoreMSBuildSettings 
+};
+
+var dotNetCoreTestSettings = new DotNetCoreTestSettings
+{
+    NoBuild = true
```
ErikSchierboom: Good catch! This is something that I wanted to do a while back, but I forgot.

ErikSchierboom:
> Also this might seem like a silly question, but do we need to build the Default and Refactoring solutions? They're a subset of the All solution, so that should catch any problems I'd think.

Well, actually we can't lose them. The thing is, the All solution contains projects that don't have a working example, so that one is ruled out. And the Refactoring solution is indeed a subset of the All and Default solution, but if you look at the build step we don't build the example file, but the stub file! We do this as for the refactoring exercises it is vital that the stub file also compiles successfully. Hence the separate build.

I've also started to look at the xUnit test runner for .NET Core. At the moment, we run `dotnet test`, which in turn defers to the `xunit.runner.visualstudio` VSTest runner for xUnit. However, there is also a separate xUnit test runner: https://xunit.github.io/docs/getting-started-dotnet-core
This test runner has several advantages, one of the being having support for running tests in parallel: https://xunit.github.io/docs/running-tests-in-parallel.html We'll still to use the TPL to run the test assemblies in parallel, but this should speed up the testing of the individual tests within an assembly.

ErikSchierboom:
> Edit: Travis ran out of memory.

Haha, that's bizarre!

I think it's great that you're working on this! One other possible area of interest would be to examine if we could use .NET Core 2 to build and run the tests, as it might be faster than V1.

build.cake
jpreese: I wasn't super happy with it, as it seems like a relatively arbitrary number, but without it I got an error from Travis saying he ran out of memory -- presumably because he spun up too many threads and started doing more work than he could handle.

I then wanted to set it equal to Environment.ProcessorCount, but noticed the build slowed down (still faster than what we had, but slower than some other tests I had done with 30.. 20..). You can see this answer for a bit more clarification (https://stackoverflow.com/questions/20806238/maxdegreeofparallelism-environment-processorcount-slows-down-execution-time-on)

jpreese:
>However, having looked at it, it does appear that the Default solution doesn't really do much. We might be able to lose it. I thought I had that there as there were some exercises that has a non-compiling stub file, but now I'm not sure anymore.

Dropping Default would be a noticeable gain, Refactoring only has about 3-4 projects in it, but Default has close to the same as All -- so it'd be great to be able to remove this.

>I think it's great that you're working on this! One other possible area of interest would be to examine if we could use .NET Core 2 to build and run the tests, as it might be faster than V1.

Right! I'd need to look into seeing what all goes into running it against Core2. I know you can set the Framework in the Cake settings for builds and tests.

ErikSchierboom:
> Dropping Default would be a noticeable gain, Refactoring only has about 3-4 projects in it, but Default has close to the same as All -- so it'd be great to be able to remove this.

Well, you could try to just remove default and see what happens :)

build.cake
ErikSchierboom: Hmmm, weird. How much slower was it with `Environment.ProcessorCount`?

jpreese:
>Well, you could try to just remove default and see what happens :)
I did it once in my RFC branch, I'M NOT AFRAID TO DO IT AGAIN! Done.

ErikSchierboom:
@jpreese Well, that was certainly a lot quicker than before! It was around 10 minutes on Travis, is that comparable to what you were expecting?

jpreese:
Yeah I honestly tried MDOP 10,20,30 and all were about the same build times on Travis. It slowed on my machine at home by about 2 minutes, but only slowed this machine down by about 7 seconds. 

Though now that I've thought about it for a bit, I think I prefer it being ProcessorCount and not some arbitrary value so that it scales with the machine that its on.

ErikSchierboom:
@jpreese Well I think it's a brilliant result! I'll merge it and if you'd like to work on this more, you could explore the .NET CLI 2.0 and/or xUnit parallel tests. Thanks a lot! :tada:

PR #388, opened 2017-08-23
[RFC] Improve build performance

jpreese:
This is another variant removing the other builds, just wanted to get your opinions and see how TravisCI performs.


jpreese:
Drops execution time down to about 10minutes (4minutes locally), which is another 3-4 minutes off. Though if we need to rebuild the other solutions directly, then we can focus on #387

ErikSchierboom:
I've just commented on #387 on why we need to build the solutions we are already building. I'll close this.

jpreese:
Thanks @ErikSchierboom , I see whats going on now with Refactoring

PR #389, opened 2017-08-29
Add NuGet package caching to Travis script

ErikSchierboom:
This should speed up the build a bit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,3 +10,6 @@ script:
   - ./bin/fetch-configlet
   - ./bin/configlet lint .
   - ./build.sh
+cache:
+  directories:
+  - $HOME/.nuget/packages


jpreese:
So a couple things I noticed.

The [things not to cache](https://docs.travis-ci.com/user/caching/#Things-not-to-cache) documentation on TravisCI points out compiled binaries, and states that it has to download the cache anyway. NuGet packages, to me, would fall into the "large download, quick to install" camp.

Looking at the restore step from this build to previous builds, the time didn't really seem to change much. Which would make sense if the behavior is to download all of the files anyway.

Theres also [this](https://github.com/aspnet/Universe/issues/394) ticket relating to a similar problem. It's not super helpful, but they ended up deciding against it.

If I correctly understand how TravisCI caching works, I don't think we'll see a lot of improvement.

ErikSchierboom:
It is logical that we won't see any performance improvement the first time: the cache isn't there yet :D I'll restart it to see if things improve. If not, we can close this.

jpreese:
Yeah I had that theory as well, I should've just clicked the restart. :(

ErikSchierboom:
@jpreese The build appears to be a bit faster (15 sec.), so I think we can merge this.

jpreese:
Yep! 👍

ErikSchierboom:
Hurray! Now all I want to do in the future is to see if running the tests with the xUnit runner (which uses parallellism by default IIRC) speeds up things, but first getting the F# track ready.

PR #391, opened 2017-08-31
Add test generator for clock exercise

robkeim:
This is for #195 

I'd love your input @ErikSchierboom and @jpreese on how to compare that two clocks are equal or not equal using the test framework. Here's what I've got so far, but I don't know how to solve the last chunk of exercises.

I see several methods:
1. Add an `Assert.AreEqual(expected, sut)` and `Assert.AreNotEqual(expected, sut)` --> This feels like the ideal method but I don't know how to implement this with the current framework
2. Add an `Equals` into the `Clock` class --> This is easy but feels like a workaround for the fact that I can't get the other method working
3. Use `Assert.[IsTrue|IsFalse](expcted.ToString() == sut.ToString())` --> Same thing as the previous


ErikSchierboom:
Can't you just use `Assert.Equal` and have the clock class implement `IEquatable<T>`? Or else make it a `struct`? This makes sense as the exercise's [HINTS.md file](https://github.com/exercism/csharp/blob/master/exercises/clock/HINTS.md) states:

> This exercise requires you to implements a type-specific method for determining equality of instances.

jpreese:
I would personally lean towards making it a struct. The clock instance is going to be small. Going to be immutable. All great reasons to consider struct and not have the student deal with overriding equals.

ErikSchierboom:
> All great reasons to consider struct and not have the student deal with overriding equals.

Well, it depends on what we want to teach the student. If we decide on a struct, we should modify the hints. Perhaps we should do a struct here, and IEquatable<T> in a later exercise?

jpreese:
I'd be good with that. I'm in the camp of trying to write things that make sense, and not necessarily try to create an example that wouldn't make sense in the real world. So for this case, you've got two sides.

1) The know-it-all student: Why is this a class? Its so simple, if this was a struct I wouldn't have to do this and that.

2) The learning-student: Wtf is a struct? I've always used classes... (and then they hopefully go on to msdn or the like and find out the differences and when to apply both) Oh thats neat, DateTime is also a struct, I just assumed it was a class. This makes total sense!

I'd rather have equatable taught in a situation where it makes a little more sense introduce.

robkeim:
@jpreese I love the personification of the types of students :)

I can convert this to a struct but we won't get equality "out of the box" because the `Hours`/`Minutes` aren't exposed as public fields. Are you guys suggesting that I convert `Hours`/`Minutes` to be exposed as public fields? If that's the case, I think we should consider whether we still need the `ToString` method, but then we're starting to drift further away from the existing implementation which is going to result in a larger "breaking change" for the students that have already solved this question.

robkeim:
@ErikSchierboom and @jpreese I've got an implementation now that should be working... let me know what you think!

jpreese:
>Are you guys suggesting that I convert Hours/Minutes to be exposed as public fields?
That's what I was thinking, yes.

>If that's the case, I think we should consider whether we still need the ToString method, but then we're starting to drift further away from the existing implementation which is going to result in a larger "breaking change" for the students that have already solved this question.

I'm personally not too worried about this. My approach is to make the changes that make the most sense in the long run. I don't think we want to necessarily keep the solutions the same as we try to improve the test suites / examples. 

Obviously you don't want to make it a habit or else students will have a hard time comparing solutions, but I think its OK in this case since we're overhauling.

generators/Output/Templates/_AssertNotEqual.liquid
```diff
@@ -0,0 +1 @@
+﻿Assert.NotEqual({{ ExpectedParameter }}, {{ TestedValue }});
```
jpreese: Doesn't look like this file is used at all

generators/Exercises/Clock.cs
```diff
@@ -0,0 +1,81 @@
...
+                    canonicalDataCase.Property = propertyToString;
+                }
+                else if (canonicalDataCase.Property == propertyEqual)
+                {
+                    canonicalDataCase.Property = propertyEquals;
+                }
+                else
+                {
+                    //  Don't modify the property
+                }
```
jpreese: The else/comment can be removed in this case

generators/Exercises/Clock.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        private const string propertyCreate = "create";
+        private const string propertyEqual = "equal";
+        private const string propertyEquals = "equals";
+        private const string propertyToString = "to_string";
+
+        protected override void UpdateCanonicalData(CanonicalData canonicalData)
+        {
+            //canonicalData.Cases = canonicalData.Cases.Where(c => c.Property != "equal").ToArray();
+
```
jpreese: Lingering comment :)

generators/Exercises/Clock.cs
```diff
@@ -0,0 +1,81 @@
...
+                    {
+                        [paramHour] = canonicalDataCase.Input[paramHour],
+                        [paramMinute] = canonicalDataCase.Input[paramMinute]
+                    };
+                }
+                else
+                {
+                    canonicalDataCase.ConstructorInput = ((JObject)canonicalDataCase.Input[paramClock1]).ToObject<Dictionary<string, object>>();
+                    canonicalDataCase.Input.Remove(paramClock1);
+
```
jpreese: This .Remove() shouldn't be needed anymore

exercises/clock/Clock.csproj
```diff
@@ -15,4 +15,8 @@
     <PackageReference Include="xunit.runner.visualstudio" Version="2.2.0" />
   </ItemGroup>
 
+  <ItemGroup>
```
ErikSchierboom: I think this section can be remove.

generators/Exercises/Clock.cs
robkeim: Fixed

generators/Exercises/Clock.cs
robkeim: In this case it's still needed because I'm combined two input params `hours` and `minutes` into a singleParam so they don't automatically get removed by the `RemoveDuplicateInputEntires`.

generators/Exercises/Clock.cs
robkeim: I actually left this in here because I find that an `if`/`else if` without an `else` is often a code smell, so I do this in my own code to show this choice was intentional. I've gone ahead and removed it though.

generators/Output/Templates/_AssertNotEqual.liquid
robkeim: Yes thanks!

exercises/clock/Clock.csproj
robkeim: Yup my bad... My Visual Studio always adds these and I forgot to not commit this one.

exercises/clock/Clock.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
```
robkeim: I don't know an easy way to revert this change as the commit that contains it is mixed in with other changes :(

robkeim: Ok `git reset` is available on a per file basis... you learn something new everyday :)
`git reset <commit hash> <filename>`

robkeim:
@ErikSchierboom and @jpreese could you guys take a look one more time?

Moving to having fields was easier than I thought as I didn't end up needing to make any changes to the generator.

I've left the `ToString()` method for now as it's actually a good opportunity to learn about formatting numbers even though it's not the main point of the exercise.

exercises/clock/ClockTest.cs
```diff
@@ -1,121 +1,356 @@
...
+    {
+        var sut = new Clock(2, 20);
+        Assert.Equal("00:20", sut.Add(-3000).ToString());
+    }
+
+    [Fact(Skip = "Remove to run test")]
+    public void Clocks_with_same_time()
+    {
+        var sut = new Clock(15, 37);
+        Assert.True(sut.Equals(new Clock(15, 37)));
```
ErikSchierboom: Can't you use `Assert.Equal` here?

robkeim: @ErikSchierboom yes I should be able to but I don't know how to to get that to work in the generator for the cases where it should be `Assert.NotEqual`. Do you have an idea on how that could work with the current generator implementation we have?

ErikSchierboom: Ehm, I'm not completely sure. You could of course override some of the methods in the base Exercise class, for example the one for rendering the assertion.

ErikSchierboom: Or would it work if you set the expected value to an `UnescapedValue`?

jpreese:
@robkeim did you need any help on getting this past the goal line?

robkeim:
@jpreese - I got stalled on this one and had a pretty busy week here, so I haven't had time to look into it further. If you've got some time to look into it, I'd be more than appreciative of the help!

jpreese:
I mean honestly, looking over the changes and reviews again, I think we should be pretty much set on this. The only outstanding issue I can see is Assert.True vs .Equal which I'm OK with letting go just to get everything else in the repo and we can address that later if we so desire.

ErikSchierboom:
I'm fine with merging this _as is_, but I do like to see an issue made for changing it to a `Equal` call.

jpreese:
Awesome! Opened #435

ErikSchierboom:
Thanks!

PR #392, opened 2017-08-31
Add test generator for grains exercise

robkeim:
For #195

### DIFF
--- a/generators/Exercises/Grains.cs
+++ b/generators/Exercises/Grains.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using Generators.Input;
-using Generators.Output;
 
 namespace Generators.Exercises
 {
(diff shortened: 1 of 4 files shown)


generators/Exercises/Grains.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+            foreach (var canonicalDataCase in canonicalData.Cases)
+            {
+                if (ShouldThrowException(canonicalDataCase.Expected))
+                {
+                    canonicalDataCase.ExceptionThrown = typeof(ArgumentOutOfRangeException);
+                }
+                else
+                {
+                    canonicalDataCase.Expected =  new UnescapedValue($"{canonicalDataCase.Expected}UL");
```
ErikSchierboom: Perhaps it would be better if you were to convert the expected value to an actual `ulong` instance, and then add the `UL` suffix when the value is formatted.

robkeim: Agreed, updated!

jpreese:
This looks fine to me 👍

ErikSchierboom:
Thanks! 🎉

PR #393, opened 2017-09-01
Implement Queen Attack generator

jpreese:
One thing to note is that our Example for this one isn't quite complete, so I'm going to start that now.

I'm open to suggestions on the generator for this one.. it feels like a lot, but it could just be a more complicated test to generate. Nothing felt too hacky, and most of it made sense to me.

Edit: Example updated to pass new tests.

### DIFF
--- a/exercises/queen-attack/QueenAttackTest.cs
+++ b/exercises/queen-attack/QueenAttackTest.cs
@@ -6,7 +6,7 @@ using System;
 public class QueenAttackTest
 {
     [Fact]
-    public void Queen_with_a_valid_position()
+    public void Queen_with_a_valid_position_does_not_throw_exception()
     {
         var actual = QueenAttack.Create(2, 2);
     }
--- a/exercises/queen-attack/QueenAttackTest.cs
+++ b/exercises/queen-attack/QueenAttackTest.cs
@@ -1,55 +1,93 @@
-﻿using System;
+// This file was auto-generated based on version 1.0.0 of the canonical data.
+
 using Xunit;
+using System;
 
 public class QueenAttackTest
 {
     [Fact]
-    public void Cannot_occupy_same_space()
+    public void Queen_with_a_valid_position()
     {
-        var white = new Queen(2, 4);
-        var black = new Queen(2, 4);
-        Assert.Throws<ArgumentException>(() => Queens.CanAttack(white, black));
+        var actual = QueenAttack.Create(2, 2);
     }
 
     [Fact(Skip = "Remove to run test")]
-    public void Cannot_attack()
+    public void Queen_must_have_positive_rank()
     {
-        Assert.False(Queens.CanAttack(new Queen(2, 3), new Queen(4, 7)));
+        Assert.Throws<ArgumentOutOfRangeException>(() => QueenAttack.Create(-2, 2));
     }
 
     [Fact(Skip = "Remove to run test")]
-    public void Can_attack_on_same_row()
+    public void Queen_must_have_rank_on_board()
     {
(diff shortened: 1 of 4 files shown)


ErikSchierboom:
I'm a bit confused. I would expect the `Create` method to actually create a queen, not return a boolean value.

jpreese:
That was my initial approach as well. I added the Create() method in the stub and had it return type Queen. Though the canonical data has it returning 0 for success, and -1 for failure, I'm not sure why it uses a number for creation, but a boolean for attacking.

Though now that I'm thinking aloud, it may be better to not take the canonical data so literally, and actually go back to Create returning Queen, but throw an exception.

ErikSchierboom:
Agreed! The canonical data can be quite weird sometimes.

jpreese:
Lets give that approach a whirl.

jpreese:
In case the github ping didn't go off for some reason, this has been changed to throw exceptions.

exercises/queen-attack/QueenAttackTest.cs
```diff
@@ -1,55 +1,93 @@
...
 using Xunit;
+using System;
 
 public class QueenAttackTest
 {
     [Fact]
-    public void Cannot_occupy_same_space()
+    public void Queen_with_a_valid_position()
+    {
+        var actual = QueenAttack.Create(2, 2);
```
ErikSchierboom: Perhaps add a comment below this line to indicate that we are testing that no exception is thrown.

generators/Exercises/QueenAttack.cs
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        protected override string RenderTestMethodBodyAssert(TestMethodBody testMethodBody)
+        {
+            if (testMethodBody.CanonicalDataCase.Property == "canAttack")
+            {
+                return RenderCanAttackAssert(testMethodBody);
+            }
+
+            if(testMethodBody.UseVariableForTested)
```
ErikSchierboom: Add space after `if`

generators/Exercises/QueenAttack.cs
```diff
@@ -0,0 +1,84 @@
...
+
+namespace Generators.Exercises
+{
+    public class QueenAttack : Exercise
+    {
+        protected override void UpdateCanonicalData(CanonicalData canonicalData)
+        {
+            foreach(var canonicalDataCase in canonicalData.Cases)
+            {
+                if(canonicalDataCase.Property == "create")
```
ErikSchierboom: Add space after `if`

generators/Exercises/QueenAttack.cs
```diff
@@ -0,0 +1,84 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace Generators.Exercises
+{
+    public class QueenAttack : Exercise
+    {
+        protected override void UpdateCanonicalData(CanonicalData canonicalData)
+        {
+            foreach(var canonicalDataCase in canonicalData.Cases)
```
ErikSchierboom: Add space after `foreach`

jpreese:
I really need to remember to update my visual studio formatting preferences.

ErikSchierboom:
> I really need to remember to update my visual studio formatting preferences.

You really shouldn't have to remember, we should have the tooling do that! I've created an issue for this: #398

jpreese:
I don't think it's worth making a special template for the initial case of making a comment, so I just added it to the test description.

ErikSchierboom:
@jpreese Did you re-run the test generator after making that change? It doesn't appear in the test file :)

ErikSchierboom:
Thanks a lot! 🎉

PR #394, opened 2017-09-02
Infer value of properties in canonical data case model

jpreese:
This makes some assumptions that I think will always hold true and changes the CanonicalDataCase model to being purely anemic to a little richer. We have quite a few tests and this logic doesn't change any of them and reduces the amount of code needed to produce the tests.

I admittedly might be getting too clever here, but I noticed some patterns and wanted to get some feedback from you guys. I felt it was at least worth discussion.


ErikSchierboom:
Great work @jpreese ! 🎉

robkeim:
@jpreese nice changes, I like it too!

PR #395, opened 2017-09-02
Refactor exercise access modifiers

jpreese:
A little refactoring.

1) Most of the methods that live inside of the Exercise base class are protected virtual, so that they can be overridden. However, in practice, only a select few are really ever going to be overridden. This may change in the future, but its always easy to open up accessibility modifiers, rather than restrict them later down the road.

2) Cleaned up how we can add namespaces. I personally like to avoid having to call into the base method where possible, and just keep the methods that the test generators will be using to be as simple and straight forward as possible. If you want to add more namespaces, then just go ahead and return a list of namespaces you want to be added.

### DIFF
--- a/generators/Exercises/ComplexNumbers.cs
+++ b/generators/Exercises/ComplexNumbers.cs
@@ -62,17 +62,17 @@ namespace Generators.Exercises
 
         private static string RenderComplexNumberAssert(TestMethodBody testMethodBody)
         {
-            var tmeplate = "Assert.Equal({{ ExpectedParameter }}.Real(), {{ TestedValue }}.Real(), 15);\r\nAssert.Equal({{ ExpectedParameter }}.Imaginary(), {{ TestedValue }}.Imaginary(), 15);";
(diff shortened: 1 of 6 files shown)


generators/Exercises/ComplexNumbers.cs
```diff
@@ -62,17 +62,17 @@ protected override string RenderTestMethodBodyAssert(TestMethodBody testMethodBo
 
         private static string RenderComplexNumberAssert(TestMethodBody testMethodBody)
         {
-            var tmeplate = "Assert.Equal({{ ExpectedParameter }}.Real(), {{ TestedValue }}.Real(), 15);\r\nAssert.Equal({{ ExpectedParameter }}.Imaginary(), {{ TestedValue }}.Imaginary(), 15);";
+            var template = "Assert.Equal({{ ExpectedParameter }}.Real(), {{ TestedValue }}.Real(), 15);\r\nAssert.Equal({{ ExpectedParameter }}.Imaginary(), {{ TestedValue }}.Imaginary(), 15);";
```
robkeim: Thanks for fixing my typo :)

ErikSchierboom:
I agree with all points. Thanks!

ErikSchierboom:
Whoops. I fat fingered and accidentally closed the PR.

ErikSchierboom:
Thanks! :tada:

PR #397, opened 2017-09-04
Simplify build script

ErikSchierboom:
This PR is a combination of four things all related to the build script, most of which are a backport of what has been done in the F# track:

- Remove the unneeded `Exercises.Default.sln` and `Exercises.Refactoring.sln` solutions.
- Remove unneeded x86 and x64 configurations.
- Use .NET CLI version 2.0 to build and test the projects.
- Allow testing of a single exercise.

If this is merged, please don't squash the commits this time as they are valuable to retain as separate commits I feel.


robkeim:
Looks like there's an issue with the build on this one @ErikSchierboom.

ErikSchierboom:
@robkeim Fixed!

robkeim:
Nice job @ErikSchierboom!

jpreese:
Hurrah one solution! Does .NET Core 2 default no build on tests to true? I see we removed that setting.

ErikSchierboom:
> Does .NET Core 2 default no build on tests to true? I see we removed that setting.

Nope, but we don't need to do an explicit build or rastore anymore as `dotnet test` now does that implicitly.

Issue #264, opened 2017-03-24
Add documentation on how to add or update a test data generator


ErikSchierboom:

jpreese:
I'd also like to include in this, **why** we have a generator. Talk about the canonical data, versions, etc.

ErikSchierboom:
Good point. We should take care not to duplicate any of the content that already exists, e.g. https://github.com/exercism/docs/tree/master/language-tracks/exercises and https://github.com/exercism/docs/blob/master/language-tracks/exercises/anatomy/test-suites.md.

ErikSchierboom:
Fixed by #399.

jpreese:
Closing

PR #400, opened 2017-09-05
Add blurb about running single exercise

jpreese:
Continuation of #264 to include @robkeim 's comments.


robkeim:
Nice work @jpreese!

PR #429, opened 2017-09-07
Remove Exe OutputType from PropertyGroup

jpreese:
This shouldn't really be necessary as we do not have main methods / run programs; just ensure test suites pass.

I also snuck in a couple upgrades to .NET Core 2 since we had some 1.1s that got through the find/replace.


robkeim:
Thanks @jpreese!

PR #431, opened 2017-09-11
Store problem-specifications cache in user specific folder

ErikSchierboom:
This PR moves the problem-specifications repository to a folder within the user's application data. This has two benefits:

1. Visual Studio no longer tries to commit the downloaded `.problem-specifications`  directory
1. We can share the downloaded problem specifications between the C# and F# generators


jpreese:
Great idea!

PR #433, opened 2017-09-14
Parallel.ForEach

thirdppp:
Parallel.ForEach

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -74,7 +74,9 @@ Task("TestUsingExampleImplementation")
     .IsDependentOn("ReplaceStubWithExample")
     .Does(() => {
         var allProjects = GetFiles(buildDir + "/*/*.csproj");
-        Parallel.ForEach(allProjects, parallelOptions, (project) => DotNetCoreTest(project.FullPath));
+       Parallel.ForEach(refactoringProjects, (project) => {
+           DotNetCoreBuild(project.FullPath, dotNetCoreBuildSettings);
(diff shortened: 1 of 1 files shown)


build.cake
```diff
@@ -74,11 +74,13 @@ Task("TestUsingExampleImplementation")
     .IsDependentOn("ReplaceStubWithExample")
     .Does(() => {
         var allProjects = GetFiles(buildDir + "/*/*.csproj");
-        Parallel.ForEach(allProjects, parallelOptions, (project) => DotNetCoreTest(project.FullPath));
+       Parallel.ForEach(refactoringProjects, (project) => {
```
robkeim: Hello @thirdppp! Could you give us some more context on what problem you're trying to solve here?

@ErikSchierboom knows the build system better than I do, but my understanding is that `DotNotCoreTest` now contains `DotNetCoreBuild` so it's not necessary to explicitly call it.

jpreese: @thirdppp , @robkeim is correct in this case. DotNetCoreTest() will call the build for us so there's no need to put it into the ForEach as we'd be building the solution twice.

ErikSchierboom: Note: this is a feature of the .NET Core version 2.0, it doesn't work with version 1.x

jpreese:
@thirdppp I'm going to close this due to inactivity. There's a ton of other opportunities in this repository if you would like to still contribute!

PR #442, opened 2017-09-28
Added DifferenceOfSquares Test Generator

felix91gr:
Thank you @jpreese for the help :)

I think I'm starting understand the basics of how the Exercises framework works :D


ErikSchierboom:
@felix91gr Just to be safe: did you run the test generator project? That is, did you run `dotnet run` in the `generators` directory? You might have forgotten that, as I get a compile error when I run your branch on my machine. This compile error is due to a problem in the test generators that was just fixed on the `master` branch. 

What you should do is to rebase your branch on the latest master and then try to run the generator again. If there is no compile error, the test generator should run. Usually, the test file generated by the newly added generator is slightly different from what it was, which I can confirm is the case once you have done the rebase and re-run the generator. 

When the generated tests file is different, you have to check to see if the example and default (stub) implementation of that exercise are still correct. You can verify that my running the build script (build.ps1 or build.sh, depending on your operating system).

If you have any further questions, feel free to ask!

felix91gr:
> Just to be safe: did you run the test generator project? That is, did you run `dotnet run` in the generators directory?

No, I didn't, I didn't know. I should've asked. What I did do was run `dotnet test` in the `generators/Exercises/` directory.

> This compile error is due to a problem in the test generators that was just fixed on the `master` branch.

Ohhh, I see. Thanks! I'll rebase it first then :)

> If you have any further questions, feel free to ask!

I will! Thank you so much :)

felix91gr:
I did the rebase.

But I don't know if the PR now works... I'm Starbucks' internet flaked out and I think NuGet needs a stable connection to work. NuGet gets invoked when I write `run dotnet` for some reason. 
If Travis passes, it's good?

Anyhow... when I'm back home I'll fix it if it doesn't work :)

ErikSchierboom:
dotnet run indeed needs an internet connection as it is downloading the Nuget packages. Having the Travis build pass is not enough unfortunately, as it currently does not run the generators project. 

Good luck fixing things and we really appreciate the help!

felix91gr:
Now I ran `dotnet run` without any failures. It generated everything allright.

What should I check next before the merge?

felix91gr:
I changed and tested the test generator and the tests generated. Git didn't let me work without automatically changing the files that have `CRLF`, replacing it with `LF`, so... there's that :thinking: 

But everything else worked c:

felix91gr:
>  So could you revert this back to DifferenceOfSquares?

Yeah, sure! :)

> could you remove all files not related to the DifferenceOfSquares exercise from the PR?

I'm not sure. `git` isn't letting me go on without changing those files. I think that when they are cloned, they aren't written as-is to my machine. I try to `git stash` them, but nothing happens. 

I'll investigate a bit more on this.

felix91gr:
:thinking: do we have a [git attributes file](https://help.github.com/articles/dealing-with-line-endings/#per-repository-settings)? It seems we could streamline the `CRLF` v/s `LF` process with one of those :)

ErikSchierboom:
I'm off to bed now. I'll have some tips how to deal with the git issues tomorrow!

felix91gr:
Speaking of this:

> as the test generator name has to match the exercise name

:thinking: we might have to change the `canonical-data.json` file for this one. 

Let me explain: using the name _DifferenceOfSquares_ for the class isn't allowed, as the class has a member with the same name:

```csharp
public static int DifferenceOfSquares(int max)
{
}
```

With the class being named the same as this method, I get the following error while running `build.sh`:

```console
DifferenceOfSquares.cs(19,23): error CS0542: 'DifferenceOfSquares': member names cannot be the same as their enclosing type [/home/felix/Documents/Programming/exercism/csharp/build/difference-of-squares/DifferenceOfSquares.csproj]
```

And if I have understood the framework correctly, the method name can't change unless the property name described in the `canonical-data.json` file changes as well.

What do you think? Should we keep the Tester Generator being named `Squares`, or should we modify the canonical data file?

ErikSchierboom:
> What do you think? Should we keep the Tester Generator being named Squares, or should we modify the canonical data file?

Excellent question. In general, we never modify the canonical data itself (which we don't control), we just modify our internal representation (which we _do_ control). An example of this is the [test generator for the leap exercise](), which changes the tested method to a different name by overriding the `UpdateCanonicalData` method:

```csharp
public class Leap : Exercise
{
    protected override void UpdateCanonicalData(CanonicalData canonicalData)
    {
        foreach (var canonicalDataCase in canonicalData.Cases)
            canonicalDataCase.Property = "IsLeapYear";
    }
}
```

I think in our case we have two options:

1. Use a different name for the class (e.g. Squares). However, I would then also would like the test file named SquaresTests, which would require a modification to the generator framework.
1. Use a different for the tested method. I'm in favor of this option, where we rename the tested method from `DifferenceOfSquares` to `CalculateDifferenceOfSquares` (and I think we should then also change `SumOfSquares` to `CalculateSumOfSquares`).

If you agree with me and go for option 2, you should be able to do this using the same approach used in the `Leap` exercise (as shown above)

> do we have a git attributes file? It seems we could streamline the CRLF v/s LF process with one of those :)

We do have a [.gitattributes](https://github.com/exercism/csharp/blob/master/.gitattributes) file, and I thought that this worked well on my Mac. I'll have to check another time.

> I'm not sure. git isn't letting me go on without changing those files. I think that when they are cloned, they aren't written as-is to my machine. I try to git stash them, but nothing happens.

You shouldn't change the files themselves, you should just use git to revert them to their original version. You can do this using `git checkout`, but you should first make sure you are up to date with the latest version of this repository. So we'll start with doing that:

```bash
git remote add exercism https://github.com/exercism/csharp
git fetch exercism
```

You can skip the first line if you already have add the `exercism` remote.

We then reset the files we don't want to the version in the exercism remote:

```bash
git checkout exercism/master exercises/largest-series-product/Example.cs
git checkout exercism/master exercises/largest-series-product/LargestSeriesProductTest.cs
git checkout exercism/master generators/Exercises/Clock.cs
git add exercises/largest-series-product/Example.cs
git add exercises/largest-series-product/LargestSeriesProductTest.cs
git add generators/Exercises/Clock.cs
git commit --amend
git push --force
```

This should remove the changes in those files from your PR. As long as you never run `dotnet run`, but always run `dotnet run -e difference-of-squares`, those changes will not come back.

felix91gr:
> (...) If you agree with me and go for option 2 (...)

Awesome explanation! I get it better now, and I completely agree. I'll do that.

> We do have a .gitattributes file,

Hmm. It looks like the `eol` for `*` files isn't predefined. That might be the problem. One of the commits in the repo's history put `CRLF` endings in these two files, or something like that:

```bash
~/D/P/exercism $ git clone https://github.com/exercism/csharp/ csharp_original
# blablabla
~/D/P/exercism $ cd csharp_original/
~/D/P/e/csharp_original (master) $ git status
On branch master
Your branch is up-to-date with 'origin/master'.
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   exercises/largest-series-product/Example.cs
	modified:   exercises/largest-series-product/LargestSeriesProductTest.cs

no changes added to commit (use "git add" and/or "git commit -a")
~/D/P/e/csharp_original (master) $ git add .
warning: CRLF will be replaced by LF in exercises/largest-series-product/Example.cs.
The file will have its original line endings in your working directory.
warning: CRLF will be replaced by LF in exercises/largest-series-product/LargestSeriesProductTest.cs.
The file will have its original line endings in your working directory.
```
Git's messages are a bit ambiguous here. But I think I'm understanding them allright :thinking: 

The `CRLF` endings must be messing with my computer, which has Git set it to default to `LF`. I think we might be better off picking a default line ending, and saying goodbye to this problem once and for all :thinking:

What do you think?

ErikSchierboom:
> I think we might be better off picking a default line ending, and saying goodbye to this problem once and for all

@robkeim @jpreese What do you think? Should we just use `\n` as the line endings?

jpreese:
As long as we choose one, for consistency, I don't really think it matters. https://help.github.com/articles/dealing-with-line-endings/

ErikSchierboom:
@felix91gr Okay, we'll solve the git issue in another issue. Could you cleanup this PR as requested (to remove the extra files)?

felix91gr:
Awesome!

And yes, I will :)

I’ll be back in a couple of days: atm, I’m kinda busy and also kinda sleep
deprived xD
I hope I’ll be back by... Wednesday, or so.

See ya!

robkeim:
@ErikSchierboom - I'm with @jpreese and as long as we have consistent way of doing things that sounds good to me! I use Windows and every editor that I know renders `\n` correctly these days so I don't see any potential concerns from people trying to contribute on Windows if we go that direction.

ErikSchierboom:
I've created an issue for the line-ending issue: #452

jpreese:
Did you need any help with this @felix91gr ?

felix91gr:
🤔 I shouldn’t be having more git issues, and therefore this PR should be
streamlined from now on.

I’ll try to finish it during this week... I’ll ask you for help if I’m
stuck ^^

felix91gr:
I deleted my fork and started over again. I will link to this PR as reference from a new PR

jpreese:
You can push your changes to the same branch -- no need to close and open a new PR.

felix91gr:
Yeah... but I messed up too many things:

- Used `master` instead of a new branch
- Merged progress made in `exercism/csharp/master` and put it as my commit
- Made spurious commits with my deprecated `git`

I get what you mean. But I _really_ would prefer to start this one again clean.

PR #447, opened 2017-09-30
Add Two Bucket Test Generator

GKotfis:
I've mixed feelings about what I've done but maybe it's ok ;)
1. SetConstructorInputParameters() is this hardcoded properties array is ok?
2. RenderTestMethodBodyAssert. I tried to do this by using some for loops by Expected values (to avoid this hardcoded property) but without any good result.
Please feel free to give some advice in case something needs change.

### DIFF
--- a/exercises/two-bucket/TwoBucket.cs
+++ b/exercises/two-bucket/TwoBucket.cs
@@ -9,13 +9,13 @@ public enum Bucket
 public class TwoBucketResult
 {
     public int Moves { get; set; }
-    public Bucket Goal_bucket { get; set; }
-    public int Other_bucket { get; set; }
+    public Bucket GoalBucket { get; set; }
+    public int OtherBucket { get; set; }
 }
 
 public class TwoBucket
 {
-    public TwoBucket(int bucket_one, int bucket_two, Bucket start_bucket)
+    public TwoBucket(int bucketOne, int bucketTwo, Bucket startBucket)
     {
         throw new NotImplementedException("You need to implement this function.");
     }
--- a/exercises/two-bucket/TwoBucket.cs
+++ b/exercises/two-bucket/TwoBucket.cs
@@ -9,18 +9,18 @@ public enum Bucket
 public class TwoBucketResult
 {
     public int Moves { get; set; }
-    public Bucket GoalBucket { get; set; }
-    public int OtherBucketContents { get; set; }
+    public Bucket Goal_bucket { get; set; }
+    public int Other_bucket { get; set; }
 }
 
-public class TwoBuckets
+public class TwoBucket
 {
-    public TwoBuckets(int bucketOneSize, int bucketTwoSize, Bucket startBucket)
+    public TwoBucket(int bucket_one, int bucket_two, Bucket start_bucket)
     {
         throw new NotImplementedException("You need to implement this function.");
     }
 
-    public TwoBucketResult Solve(int goal)
+    public TwoBucketResult Measure(int goal)
     {
         throw new NotImplementedException("You need to implement this function.");
     }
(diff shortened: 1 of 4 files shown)


exercises/two-bucket/TwoBucket.cs
```diff
@@ -9,18 +9,18 @@ public enum Bucket
 public class TwoBucketResult
 {
     public int Moves { get; set; }
-    public Bucket GoalBucket { get; set; }
-    public int OtherBucketContents { get; set; }
+    public Bucket Goal_bucket { get; set; }
```
ErikSchierboom: Please rename this to `GoalBucket`, to conform to standard [.NET naming guidelines](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/general-naming-conventions).

exercises/two-bucket/TwoBucket.cs
```diff
@@ -9,18 +9,18 @@ public enum Bucket
...
+    public int Other_bucket { get; set; }
```
ErikSchierboom: Please rename this to `OtherBucket`, to conform to standard [.NET naming guidelines](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/general-naming-conventions).

exercises/two-bucket/TwoBucket.cs
```diff
@@ -9,18 +9,18 @@ public enum Bucket
...
 }
 
-public class TwoBuckets
+public class TwoBucket
 {
-    public TwoBuckets(int bucketOneSize, int bucketTwoSize, Bucket startBucket)
+    public TwoBucket(int bucket_one, int bucket_two, Bucket start_bucket)
```
ErikSchierboom: Could you rename the parameters to `bucketOne`, `bucketTwo` and `startBucket`?

generators/Exercises/TwoBucket.cs
```diff
@@ -0,0 +1,49 @@
...
+namespace Generators.Exercises
+{
+    public class TwoBucket : Exercise
+    {
+        protected override void UpdateCanonicalData(CanonicalData canonicalData)
+        {
+            foreach (var canonicalDataCase in canonicalData.Cases)
+            {
+                canonicalDataCase.TestedMethodType = TestedMethodType.Instance;
+                canonicalDataCase.SetConstructorInputParameters(new [] { "bucket_one", "bucket_two", "start_bucket"});
```
ErikSchierboom: The `SetConstructorInputParameters` take a `param string[]`, which means you can simplify this to: `SetConstructorInputParameters("bucket_one", "bucket_two", "start_bucket")`

GKotfis:
I was aware of breaking .net coding guidelines here. Wanted match canonical data properties with variables names. But I'll simply replace them in update canonical data method. Stay tuned 😉

ErikSchierboom:
Great! Let me say once again that I'm impressed with the quality of your PR!

GKotfis:
Thanks for the compliment! I think I'll grab another issue :)

jpreese:
Yeah, this looks great @GKotfis! Just as a heads up, since you changed the name of the class, you'll also need to update the `Example.cs` file so that it matches.

exercises/two-bucket/TwoBucket.cs
```diff
@@ -10,17 +10,17 @@ public class TwoBucketResult
 {
     public int Moves { get; set; }
     public Bucket GoalBucket { get; set; }
-    public int OtherBucketContents { get; set; }
+    public int OtherBucket { get; set; }
 }
 
-public class TwoBuckets
+public class TwoBucket
```
jpreese: Example.cs should be changed as well.

ErikSchierboom:
Everything's green, so I'll merge it. Thanks a lot! 🎉

PR #448, opened 2017-10-01
Adds Generators solution to build

### LINKED ISSUE #441: Include Generators Solution in Build
Cake currently only builds the Exercises solution, and not the Generators solution. Without this check, we don't get the benefit of Travis validating the entirety of the build for us. This gap makes it possible to merge in a broken build.

jakauppila:
Added the Generators solution to build.cake. Includes downloading Nuget.exe to build.ps1.

The only thing I'm not sure of at this point is how to to perform the NuGetRestore on Linux for Travis; I assume this will involve utilizing Mono.

Fixes #441

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -5,8 +5,8 @@ using System.Threading.Tasks;
 var target = Argument("target", "Default");
 var exercise = Argument<string>("exercise", null);
 
-var sourceDir = "./exercises";
-var buildDir  = "./build";
+var exercisesSourceDir = "./exercises";
+var exercisesBuildDir  = "./build";
 
 var parallelOptions = new ParallelOptions 
 {
@@ -15,21 +15,21 @@ var parallelOptions = new ParallelOptions
 
 Task("Clean")
     .Does(() => {
-		CleanDirectory(buildDir);   
+		CleanDirectory(exercisesBuildDir);   
     });
 
 // Copy everything to build so we make no changes in the actual files.
 Task("CopyExercises")
     .IsDependentOn("Clean")
     .Does(() => {
-        CopyDirectory($"{sourceDir}/{exercise}", $"{buildDir}/{exercise}");
+        CopyDirectory($"{exercisesSourceDir}/{exercise}", $"{exercisesBuildDir}/{exercise}");
     });
 
 Task("EnableAllTests")
     .IsDependentOn("CopyExercises")
     .Does(() => {
         var skipRegex = new Regex(@"Skip\s*=\s*""Remove to run test""", RegexOptions.Compiled);
-        var testFiles = GetFiles(buildDir + "/*/*Test.cs");
+        var testFiles = GetFiles(exercisesBuildDir + "/*/*Test.cs");
 
         foreach (var testFile in testFiles) {
             var contents = System.IO.File.ReadAllText(testFile.FullPath);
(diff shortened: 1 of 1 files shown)


build.cake
```diff
@@ -73,12 +84,21 @@ Task("ReplaceStubWithExample")
...
-        var allProjects = GetFiles(buildDir + "/*/*.csproj");
+        var allProjects = GetFiles(exercisesBuildDir + "/*/*.csproj");
         Parallel.ForEach(allProjects, parallelOptions, (project) => DotNetCoreTest(project.FullPath));
     });
 
+Task("BuildGeneratorsSolution")
+    .IsDependentOn("CopyGenerators")
+    .Does(() => {
+        var generatorsProject = GetFiles(generatorsBuildDir + "/*.csproj");
+        Parallel.ForEach(generatorsProject, parallelOptions, (project) => NuGetRestore(project.FullPath));
```
jpreese: We shouldn't need to call NuGetResore() explicitly since DotNetCoreBuild() will do an implicit restore for us.

build.cake
```diff
@@ -73,12 +84,21 @@ Task("ReplaceStubWithExample")
...
+        Parallel.ForEach(generatorsProject, parallelOptions, (project) => DotNetBuild(project.FullPath));
```
jpreese: This should be DotNetCoreBuild as this is a .NET Core 2 project.

jpreese: The generators solution contains one project, so we shouldn't need to run this build step in parallel.

build.cake
```diff
@@ -19,17 +24,23 @@ Task("Clean")
     });
 
 // Copy everything to build so we make no changes in the actual files.
+Task("CopyGenerators")
+    .Does(() => {
+        CopyDirectory($"{generatorsSourceDir}/", $"{generatorsBuildDir}");
+    });
+
```
jpreese: We don't perform any file replacements in the generators project like we do in the exercises (replacing the example with the stub). There shouldn't be any need to copy the generators directory.

build.ps1
```diff
@@ -74,6 +76,14 @@ if (!(Test-Path $CakePath)) {
...
 
 ###########################################################################
+# Download NUGET
+###########################################################################
+
+$NuGetExePath = Join-Path $ToolPath "nuget.exe"
+
+(New-Object System.Net.WebClient).DownloadFile($NuGetExeUri, $NuGetExePath);
+
+###########################################################################
```
jpreese: Travis already has NuGet on the CI server, so we shouldn't need to download it.

jpreese: Travis already has NuGet, so we shouldn't need to install it.

jpreese:
Thanks for working on this, @jakauppila ! I've requested some changes for your PR. If you have any questions, please do not hesitate to ask.

jakauppila:
@jpreese Made the requested changes, Travis is now green!

jpreese:
Excellent, thanks!

build.cake
```diff
@@ -5,7 +5,11 @@ using System.Threading.Tasks;
 var target = Argument("target", "Default");
 var exercise = Argument<string>("exercise", null);
 
-var sourceDir = "./exercises";
+var exercisesSourceDir = "./exercises";
```
ErikSchierboom: I don't think we need both an `exercisesSourceDir` and `buildDir` variable, as the only thing that is built there are the exercises. I think we could just revert to the old situation here, where the exercise directory are directly placed and built in the `./build` directory. I do think I like your naming better, so I would like to see something like this:

```csharp
var exercisesSourceDir = "./exercises";
var exercisesBuildDir = "./build";
```

And then you could drop the `buildDir` parameter and replace it with `exercisesBuildDir` wherever it is used.

build.cake
```diff
@@ -73,12 +77,18 @@ Task("ReplaceStubWithExample")
 Task("TestUsingExampleImplementation")
     .IsDependentOn("ReplaceStubWithExample")
     .Does(() => {
-        var allProjects = GetFiles(buildDir + "/*/*.csproj");
+        var allProjects = GetFiles(exercisesBuildDir + "/*/*.csproj");
         Parallel.ForEach(allProjects, parallelOptions, (project) => DotNetCoreTest(project.FullPath));
     });
 
+Task("BuildGeneratorsSolution")
```
ErikSchierboom: Could you change this to `BuildGenerators`? The solution part is just an implementation detail.

jakauppila:
@ErikSchierboom Requested changes have been made.

ErikSchierboom:
@jakauppila Looking great! Merged. Thanks a lot. :tada:

Issue #417, opened 2017-09-06
Add OCR Numbers Test Generator


jpreese:
This issue is part of an overall initiative to complete #195.

The documentation on how to add generators can be found [here](https://github.com/exercism/csharp/blob/master/docs/GENERATORS.md). If you get stuck or have any questions about adding this generator, please do not hesitate to reach out!

GKotfis:
I'll deal with it

GKotfis:
I need to figure it out how to generate valid digit representation as code:
![image](https://user-images.githubusercontent.com/7391978/31068688-75710c6e-a759-11e7-9fbb-636447f97e72.png)

Don't know if DotLiquid have such option to generate non string characters that are part of code - here '+' sign

GKotfis:
Have something that generate almost desired output:
![image](https://user-images.githubusercontent.com/7391978/31071573-d9521b0a-a764-11e7-8a07-0fdc51f3b412.png)
as you see I lost indentation - argument not starting in the same line as  method call. But maybe it's ok?

ErikSchierboom:
@GKotfis Hmmm, that's almost correct! There is an Indent function somewhere, but I don't know if you can easily use that in your code. If not, I think this is fine.

ErikSchierboom:
Closed by #450

PR #450, opened 2017-10-02
Add OCR Numbers Test Generator

GKotfis:
This reference #417 issue.

### DIFF
--- a/exercises/ocr-numbers/Example.cs
+++ b/exercises/ocr-numbers/Example.cs
@@ -10,8 +10,39 @@ public static class OcrNumbers
     public static string Convert(string input)
     {
         var lines = input.Split('\n');
-        
+
+        if (lines.Length > CharacterHeight)
+            lines = GetTransformedInput(lines);
+
         return Positions(lines).Aggregate("", (str, pos) => str + ConvertCharacter(lines, pos.Item1, pos.Item2));
+    }
+
+    /// <summary>
+    /// Transform multilines input into one line digits separate by comma
+    /// </summary>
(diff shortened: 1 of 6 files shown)


GKotfis:
I see that some test cases are not covered. expected value =-1 should be generated as error case.

ErikSchierboom:
@GKotfis There are some existing exercises that do something similar. You need to do something like `canonicalDataCase.ExceptionThrown = canonicalDataCase.Expected is long ? typeof(ArgumentException) : null;`.

GKotfis:
Yep after changes still three tests failed because of wrong implementation in Example (I guess Example.cs is used on CI to run tests?)

1. Input_with_a_number_of_columns_that_is_not_a_multiple_of_three_raises_an_error - return "?" instead of raising error
2. Input_with_a_number_of_columns_that_is_not_a_multiple_of_three_raises_an_error - return empty string instead of raising error
3. Numbers_separated_by_empty_lines_are_recognized_lines_are_joined_by_commas_ - throws IndexOutOfRangeException

ErikSchierboom:
@GKotfis Ah, you then need to also modify the example implementation. This is due to the canonical data being newer than the one the exercise was originally based on.

You can verify the example implementation by doing the CI build locally, but only for one specific exercise: `.\build.ps1 --exercise=ocr-numbers` (or `./build.sh --exercise=ocr-numbers` on Linux/macOS). This will execute the same steps as the CI server, but for only the specified exercise.

ErikSchierboom:
Great work @GKotfis ! Merged. Thanks! :tada:

PR #451, opened 2017-10-02
Add bowling test generator

bangsholt:
This is the initial attempt at making the generator for the bowling test.

I've annotated the Bowling.cs with a few comments, so I hope you'll have a quick look and advise where I should look to figure out what to do.

This is in relation to issue #402 

Thank you in advance,
Jesper


bangsholt:
Alright, so the build is failing - I'll have a look at this tomorrow again.

generators/Exercises/Bowling.cs
```diff
@@ -0,0 +1,84 @@
...
+            sb.AppendLine(base.RenderTestMethodBodyAssert(testMethodBody));
+
+            return sb.ToString();
+        }
+
+        protected override void UpdateCanonicalData(CanonicalData canonicalData)
+        {
+            foreach (var canonicalDataCase in canonicalData.Cases)
+            {
+                canonicalDataCase.SetConstructorInputParameters();
```
ErikSchierboom: Is this line necessary? I believe the default is empty

exercises/bowling/BowlingTest.cs
```diff
@@ -1,207 +1,350 @@
...
+    public void Rolls_cannot_score_negative_points()
+    {
+        var previousRolls = new int[] {};
+        var sut = new BowlingGame();
+        foreach (var roll in previousRolls)
+        {
+            sut.Roll(roll);
+        }
+        sut.Roll(-1);
+        Assert.Null(sut.Score());
```
ErikSchierboom: I would like the cases where the the canonical data indicates an error to throw an exception. This means you have to add something like this:

`canonicalDataCase.ExceptionThrown = canonicalDataCase.Expected is long ? null : typeof(ArgumentException);`

It probably also means removing the code explicitly setting the tested property to `Score` and the expected value to `null`.

bangsholt: Aha, I'll update it to do this. I was following how the hand made test cases were. I'll do the update for this, it should make the code nicer.

jpreese:
@bangsholt a new override was added to the `Exercise` abstract class, `string[] RenderAdditionalMethods()` which you can use to add helper methods for your test suites.

bangsholt:
I messed up this one - new one incoming soon.

Issue #416, opened 2017-09-06
Add Nucleotide Count Test Generator


jpreese:
This issue is part of an overall initiative to complete #195.

The documentation on how to add generators can be found [here](https://github.com/exercism/csharp/blob/master/docs/GENERATORS.md). If you get stuck or have any questions about adding this generator, please do not hesitate to reach out!

InKahootz:
I'll hop in with this one! Looking over the files, it seems to be a easy starter :smiley:

ErikSchierboom:
@InKahootz Good observation! The test generators should just follow the canonical data as specific in the [problem-specification repo](https://github.com/exercism/problem-specifications), so if that means losing a couple of test, that is completely fine.

ErikSchierboom:
Closed by #456

PR #456, opened 2017-10-04
Add Nucleotide Count Test Generator #416

InKahootz:
This was/is a good learning experience. I look forward to feedback and doing more work.
I tried to become as familiar as possible with the code base to come up with an elegant solution, though I didn't see a way to easily test instance properties. 

As explained further in my commit message, I took the liberty to remove the Count method from the exercise. It was previously needed in he the old unit testing but with the canonical data based unit test, it was pruned.

### DIFF
--- a/generators/Exercises/NucleotideCount.cs
+++ b/generators/Exercises/NucleotideCount.cs
@@ -1,0 +1,61 @@
+﻿using Generators.Input;
+using Generators.Output;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Generators.Exercises
+{
+    class NucleotideCount : Exercise
+    {
+        protected override void UpdateCanonicalData(CanonicalData canonicalData)
+        {
+            foreach (var canonicalDataCase in canonicalData.Cases)
+            {
+                if (!((Dictionary<string, object>)canonicalDataCase.Expected).ContainsKey("error"))
+                {
+                    canonicalDataCase.UseVariableForExpected = true;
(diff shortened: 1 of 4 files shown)


jpreese:
Awesome @InKahootz! We always welcome the help 👍 

As a heads up, if you haven't figured out already, the broken build is due to the fact that the class name was changed and the `Example.cs` was not changed as well. That should get Travis to go green for you.

You can also run powershell .\build.ps1 (for windows) or build.sh (for linux) to test your build before pushing to your branch if you so desire.

InKahootz:
Ah, I forgot `Example.cs`. Doesn't show up in VS. I'll get that fixed right quick.

generators/Exercises/NucleotideCount.cs
```diff
@@ -0,0 +1,61 @@
+﻿using Generators.Input;
+using Generators.Output;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Generators.Exercises
+{
+    class NucleotideCount : Exercise
```
ErikSchierboom: Could you add the `public` modifier (for consistency)?

InKahootz:
Thanks! I'll start looking for my next issue.
One question. Is there no way to test instance/static properties on a class without writing a custom template?

ErikSchierboom:
@InKahootz There should be. I'll look into it later. I'll merge when the CI is green.

ErikSchierboom:
Merged! Thanks a lot @InKahootz! :tada:

PR #458, opened 2017-10-04
423 triangle generator

### LINKED ISSUE #423: Add Triangle Test Generator

bmeverett:
Adds the triangle test generator. I modified the class to better take into account the canonical data as suggested by @ErikSchierboom [here](https://github.com/exercism/csharp/issues/423#issuecomment-333786737)

Closes #423

### DIFF
--- a/exercises/triangle/Triangle.cs
+++ b/exercises/triangle/Triangle.cs
@@ -9,7 +9,22 @@ public enum TriangleKind
 
 public static class Triangle
 {
-    public static TriangleKind Kind(decimal side1, decimal side2, decimal side3)
+    public bool IsScalene(decimal side1, decimal side2, decimal side3)
+    {
+        throw new NotImplementedException("You need to implement this function.");
+    }
+
+    public bool IsIsosceles(decimal side1, decimal side2, decimal side3) 
+    {
+        throw new NotImplementedException("You need to implement this function.");
+    }
+
+    public bool IsEquilaterial(decimal side1, decimal side2, decimal side3) 
+    {
+        throw new NotImplementedException("You need to implement this function.");
+    }
+    
+    private static TriangleKind Kind(decimal side1, decimal side2, decimal side3)
     {
         throw new NotImplementedException("You need to implement this function.");
     }
--- a/exercises/triangle/Triangle.cs
+++ b/exercises/triangle/Triangle.cs
@@ -9,22 +9,22 @@ public enum TriangleKind
 
 public static class Triangle
 {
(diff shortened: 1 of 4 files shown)


exercises/triangle/Triangle.cs
```diff
@@ -9,7 +9,22 @@ public enum TriangleKind
 
 public static class Triangle
 {
-    public static TriangleKind Kind(decimal side1, decimal side2, decimal side3)
+    public static bool IsScalene(double side1, double side2, double side3)
```
ErikSchierboom: Just to be consistent, we should keep the types used here and in `Example.cs` the same.

ErikSchierboom:
Just one small nit, it already looks great!

bmeverett:
@ErikSchierboom fixed! However, it doesn't seem to like the double as parameters. It complained when I ran it locally but I think the actual test is different, so I should probably change back to decimal. Is there way to run tests locally or can I just run the `build.ps1` script?

ErikSchierboom:
@bmeverett You should use the `build.ps1` script, but you can speed things up significantly by specifying the exercise you want to build/test: `build.ps1 --exercise=triangle`.

ErikSchierboom:
Merged. Great work @bmeverett! Thanks. :tada:

robkeim:
Thanks @bmeverett for taking the time to do this, much appreciated!

I know I'm late here, but I have a couple of comments (@ErikSchierboom feel free to chime in here too since you reviewed this originally):
- It would be more natural C# to have spaces between function arguments `IsScalene(1, 2, 3)` instead of `IsScalene(1,2,3)`. That seems like a one line fix
- In `Triangle.cs` shouldn't we remove the `Kind` method which is leading the students to a solution in the same direction? I'd personally be in favor of dropping that and enum in that file and letting students chose if they wanted to solve the problem that way or not.

ErikSchierboom:
Two excellent points @robkeim. The first issue was one that I felt could easily be corrected later, the second one I missed.

robkeim:
Thanks @ErikSchierboom for the input. @bmeverett would you like to make a new PR to fix these issues? If not, no worries I can do it.

bmeverett:
@robkeim @ErikSchierboom I can take a look at it. Do you want the Kind method removed from `Triangle.cs` and `Example.cs` or just `Triangle.cs`?

ErikSchierboom:
@bmeverett It's fine in `Example.cs`, as it is actually used there. It only needs to be removed from `Triangle.cs`.

robkeim:
Changes look good, thanks @bmeverett 🎉

PR #459, opened 2017-10-04
add PascalsTriangle generator

### LINKED ISSUE #418: Add Pascals Triangle Test Generator

bmeverett:
Adds the PascalsTriangle generator. Pretty straight forward, except for the exception cases. I modified the Example to get around the yield return so the exception would be thrown. And the final test case, where the value is null,  I converted that back to -1 because you can't run a test with the incorrect number of parameters as the solution doesn't build. 

Closes #418

### DIFF
--- a/exercises/pascals-triangle/Example.cs
+++ b/exercises/pascals-triangle/Example.cs
@@ -1,15 +1,24 @@
 ﻿using System.Collections.Generic;
+using System;
 
 public static class PascalsTriangle
 {
     public static IEnumerable<IEnumerable<int>> Calculate(int rows)
+    {
+        if (rows < 0)
+            throw new ArgumentOutOfRangeException();
+
+        return ItterateRows(rows);
+    }
+    
+    private static IEnumerable<IEnumerable<int>> ItterateRows(int rows)
     {
         for (var i = 1; i <= rows; i++)
         {
             yield return Row(i);
         }
     }
-    
+
     private static IEnumerable<int> Row(int row)
     {
         yield return 1;
--- a/exercises/pascals-triangle/Example.cs
+++ b/exercises/pascals-triangle/Example.cs
@@ -8,10 +8,10 @@ public static class PascalsTriangle
         if (rows < 0)
             throw new ArgumentOutOfRangeException();
 
-        return ItterateRows(rows);
+        return IterateRows(rows);
     }
     
-    private static IEnumerable<IEnumerable<int>> ItterateRows(int rows)
+    private static IEnumerable<IEnumerable<int>> IterateRows(int rows)
     {
         for (var i = 1; i <= rows; i++)
         {
--- a/exercises/pascals-triangle/PascalsTriangle.cs
+++ b/exercises/pascals-triangle/PascalsTriangle.cs
@@ -7,4 +7,14 @@ public static class PascalsTriangle
     {
         throw new NotImplementedException();
     }
+
+    private static IEnumerable<int> Row(int row)
+    {
+        throw new NotImplementedException();
+    }
+
+    private static IEnumerable<IEnumerable<int>> ItterateRows(int rows) 
+    {
+        throw new NotImplementedException();
+    }
 }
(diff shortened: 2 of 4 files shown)


generators/Exercises/PascalsTriangle.cs
```diff
@@ -0,0 +1,23 @@
...
+        {
+            foreach (var dataCases in canonicalData.Cases)
+            {
+                dataCases.UseVariableForExpected = true;
+                dataCases.Property = "calculate";
+                if (!(dataCases.Expected is JArray))
+                    dataCases.ExceptionThrown = typeof(ArgumentOutOfRangeException);
+                if (dataCases.Properties["count"] == null)
+                    dataCases.Properties["count"] = -1;
+            }
```
jpreese: Should we be converting this to a -1? It should be entirely possible to pass in null (int?), or pass in nothing with a default param.

exercises/pascals-triangle/Example.cs
```diff
@@ -1,15 +1,24 @@
...
     public static IEnumerable<IEnumerable<int>> Calculate(int rows)
     {
+        if (rows < 0)
+            throw new ArgumentOutOfRangeException();
+
+        return ItterateRows(rows);
+    }
+    
+    private static IEnumerable<IEnumerable<int>> ItterateRows(int rows)
+    {
```
jpreese: ItterateRows should be IterateRows

generators/Exercises/PascalsTriangle.cs
```diff
@@ -0,0 +1,23 @@
...
+using Newtonsoft.Json.Linq;
+
+namespace Generators.Exercises
+{
+    public class PascalsTriangle : Exercise
+    {
+        protected override void UpdateCanonicalData(CanonicalData canonicalData)
+        {
+            foreach (var dataCases in canonicalData.Cases)
+            {
```
jpreese: I would change dataCases to dataCase (most use canonicalDataCase). Since the variable is a single case.

jpreese:
Just some small nits @bmeverett, looks good though!

generators/Exercises/PascalsTriangle.cs
ErikSchierboom: @jpreese Well, I think the canonical data is at fault here. In what way would passing null make sense for calculating pascal's triangle? I've created [an issue](https://github.com/exercism/problem-specifications/issues/925) to discuss removing this strange case. In the mean time, I'm in favor of manually removing this error case from the canonical data to make sure we don't output it.

ErikSchierboom: @jpreese  @bmeverett I've created a [PR](https://github.com/exercism/problem-specifications/pull/926) to fix the canonical data.

exercises/pascals-triangle/PascalsTriangle.cs
```diff
@@ -7,4 +7,14 @@ public static IEnumerable<IEnumerable<int>> Calculate(int rows)
     {
         throw new NotImplementedException();
     }
+
+    private static IEnumerable<int> Row(int row)
```
ErikSchierboom: Could you remove this function? The stub implementation only needs the public methods to allow it to compile.

exercises/pascals-triangle/PascalsTriangle.cs
```diff
@@ -7,4 +7,14 @@ public static IEnumerable<IEnumerable<int>> Calculate(int rows)
...
+    {
+        throw new NotImplementedException();
+    }
+
+    private static IEnumerable<IEnumerable<int>> IterateRows(int rows)
```
ErikSchierboom: Could you remove this function? The stub implementation only needs the public methods to allow it to compile.

ErikSchierboom:
Merged. Thanks a lot! :tada:

generators/Exercises/PascalsTriangle.cs
jpreese: Yeah, I'd agree. I know there was a PR over in the problem-specs repo about wanting to test edge cases (e.g. null). I'm not sure if this is a result of that or where that got left off.

Issue #452, opened 2017-10-03
Fix line endings to newline


ErikSchierboom:
To prevent cross-platform issues, we should use `\n` for all our source code files, both for exercises and generators, as well as generated test files.

jpreese:
I played with this a little bit on a local branch, and the change itself is pretty straightforward. Though after the fact, every stage you perform, you're going to get warnings about CRLF to LF conversions.

I'd be curious to see how typical it is to deviate away from text=auto, which should handle it for us.

ErikSchierboom:
@jpreese Hmmm, if it proves to be more of a nuisance than a boon, we should just skip it.

jpreese:
Yeah I'd need to look into it more, but it's my understanding that `auto` will do the transform for us, without the warning. Linux users shouldn't have a problem as they're using LF anyway. I'm not honestly sure why it was impacting @felix91gr.

I then looked at a random sampling of repositories and really couldn't find any that explicitly set LF.

ErikSchierboom:
Okay, let's close this then.

felix91gr:
The problem was that my git changed some files that had CRLF instead of LF. Some Windows user must’ve accidentally created those. 

My point is, if it happened once, maybe it’ll happen again. Better enforce the LF everywhere so that we don’t get spurious file changes when git notices CRLF and decides to normalize it.

ErikSchierboom:
Well, I've tried to reproduce this on my Mac but have failed. So I suspect it might be some environment-specific issue with your machine. If more people have this issue, we can always re-open.

InKahootz:
I actually encountered the same problem @felix91gr did with those two specific files in the `largest-series-product` exercise folder. I tried a whole heap of solutions, but nothing would allow me to discard those changes.

I tried changing my global config autoclrf (which was originally set to `true`)
```
git config --global core.autoclrf false 
git config --global core.autoclrf true  
git config --global core.autoclrf input  
```
Every time I changed the setting, I removed the directory and recloned it, but as soon as a cloned it would say I had changed the line endings in those two files. I made sure it was only line endings comparing the `git diff` and `git diff --ignore-all-space`

The only thing I remember doing to solve it (my fork was 3 commits behind at the time since it was based off of `6677ba9`) was to delete my fork, refork, and clone. Somehow this fixed it.

My environment is windows. Using VSCode and VS2017.3.

felix91gr:
### About `.gitattributes`

[...]

PR #460, opened 2017-10-05
Fix line endings with `CRLF` in `largest-series-product` folder

felix91gr:
As per #452 

Now they have just `LF`


felix91gr:
Ignore this one. I managed to make it into one commit without having to hack too much around with `git` (remember that my `git` really hates files that end in `CRLF`)

PR #463, opened 2017-10-05
Add Saddle Points test generator

GKotfis:
This reference #421 issue.
1. fix file name (missing 's') SaddlePointsTest
2. add SaddlePoint test generator
3. add multidimensional value formater
 What do you think about those changes - extension of ValueFormatter? :) This switch(val) in Format method gets bigger and bigger so maybe some refactoring needs to be done here (at least split to methods).


generators/Exercises/SaddlePoints.cs
```diff
@@ -0,0 +1,52 @@
...
+                canonicalDataCase.Property = "Calculate";
+                canonicalDataCase.SetConstructorInputParameters("input");
+                canonicalDataCase.UseVariablesForConstructorParameters = true;
+                canonicalDataCase.UseVariablesForInput = true;
+                canonicalDataCase.UseVariableForTested = true;
+                canonicalDataCase.UseVariableForExpected = true;
+
+                canonicalDataCase.Properties["input"] = (canonicalDataCase.Properties["input"] as JArray).ToObject<int[,]>();
+
+                var array = (canonicalDataCase.Expected as Array);
```
ErikSchierboom: You can lose the parentheses here

exercises/saddle-points/SaddlePointsTest.cs
```diff
@@ -0,0 +1,75 @@
...
+using System;
+
+public class SaddlePointsTest
+{
+    [Fact]
+    public void Can_identify_single_saddle_point()
+    {
+        var input = new[,]
+        {
+        	{ 9, 8, 7 },
```
ErikSchierboom: The indentation is slight off (also in the other cases), I would expect the following:

```csharp
var input = new[,]
{
    { 9, 8, 7 },
};
```

GKotfis:
I'll do refactoring in separate PR

ErikSchierboom:
Merged. Thanks again! :tada:

PR #457, opened 2017-10-04
Add Tournament test generator

GKotfis:
This reference #422 issue

PR #464, opened 2017-10-06
add connect generator and modify class to accept array input

### LINKED ISSUE #404: Add Connect Test Generator

bmeverett:
Adds the Connect generator. Minor modifications to the class to accept an array as input.

Resolves #404

### DIFF
--- a/exercises/connect/ConnectTest.cs
+++ b/exercises/connect/ConnectTest.cs
@@ -1,7 +1,6 @@
 // This file was auto-generated based on version 1.0.0 of the canonical data.
 
 using Xunit;
-using System;
 
 public class ConnectTest
 {
@@ -9,14 +8,14 @@ public class ConnectTest
     public void An_empty_board_has_no_winner()
     {
         var board = new [] 
-                        { 
-                            ". . . . .",
-                            " . . . . .",
-                            "  . . . . .",
-                            "   . . . . .",
-                            "    . . . . ."
-                            
-                        } ;
+        { 
+            ". . . . .",
+            " . . . . .",
+            "  . . . . .",
+            "   . . . . .",
+            "    . . . . ."
+             
+        };
         var sut = new Connect(board);
         Assert.Equal(ConnectWinner.None, sut.Result());
     }
@@ -25,9 +24,9 @@ public class ConnectTest
     public void X_can_win_on_a_1x1_board()
     {
         var board = new [] 
-                        { 
-                            "X"
-                        } ;
+        { 
+            "X" 
+        };
         var sut = new Connect(board);
         Assert.Equal(ConnectWinner.Black, sut.Result());
     }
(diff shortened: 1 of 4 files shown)


exercises/connect/ConnectTest.cs
```diff
@@ -1,124 +1,86 @@
...
+using Xunit;
 
 public class ConnectTest
 {
-    private static string MakeBoard(string[] board)
+    [Fact]
+    public void An_empty_board_has_no_winner()
     {
-        return string.Join("\n", board.Select(x => x.Replace(" ", "")));
+        var board = new[] { ".....", ".....", ".....", ".....", "....." };
```
ErikSchierboom: For the Connect case, it is quite important that the input strings are aligned nicely under each other, to allow users to see the pattern that needs to be connected. By default, the generator framework does not support this. but perhaps you can "borrow" from [this generator](https://github.com/exercism/csharp/pull/457/files#diff-2de73a257c3f6aa9810bdd1be46c7ab4R37)?

generators/Exercises/Connect.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        protected override void UpdateCanonicalData(CanonicalData canonicalData)
+        {
+            foreach (var canonicalDataCase in canonicalData.Cases)
+            {
+                canonicalDataCase.UseVariablesForConstructorParameters = true;
+                canonicalDataCase.SetConstructorInputParameters("board");
+
+                //remove whitespace
+                var trimBoard = new List<string>();
```
ErikSchierboom: Why trim the white space? The original implementation also doesn't trim it and this makes it look better IMHO. Could you remove the trimming?

bmeverett: If I don't trim it the tests where failing because I think it's looping through every character and taking the spaces as an empty space. I could be wrong, I'll look into it.

bmeverett:
@ErikSchierboom I pushed up some changes. I aligned the arrays better, but I couldn't figure out how to get rid of the empty line at the end of the array. Besides that, hopefully I addressed your comments.

exercises/connect/ConnectTest.cs
```diff
@@ -1,124 +1,160 @@
...
 
 public class ConnectTest
 {
-    private static string MakeBoard(string[] board)
+    [Fact]
+    public void An_empty_board_has_no_winner()
     {
-        return string.Join("\n", board.Select(x => x.Replace(" ", "")));
+        var board = new [] 
+                        {
```
ErikSchierboom: Could you perhaps use less indentation? Something like this looks better IMHO:

```csharp
var board = new [] 
{ 
    ". . . . .",
    " . . . . .",
    "  . . . . .",
    "   . . . . .",
    "    . . . . ."
};
```

ErikSchierboom:
Merged. Thanks a lot. 🎉

PR #470, opened 2017-10-09
Added Bowling Test Generator

bangsholt:
The second attempt at making the bowling test generator.

Thanks to the added `RenderAdditionalMethods` it has come out somewhat prettier - There's still a few things I'd like input on before I would consider it finished :)

### DIFF
--- a/exercises/bowling/BowlingTest.cs
+++ b/exercises/bowling/BowlingTest.cs
@@ -12,7 +12,7 @@ public class BowlingTest
         var previousRolls = new int[] {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
         var sut = new BowlingGame();
         DoRoll(previousRolls, sut);
-        Assert.Equal(0, actual);
+        Assert.Equal(0, sut.Score());
     }
 
     [Fact(Skip = "Remove to run test")]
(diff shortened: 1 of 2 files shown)


jpreese:
Looks like not all of the tests passed, @bangsholt. Is this a WIP? Did you need any help with anything?

bangsholt:
I didn't realise the example.cs was used to run the tests.

I guess I'll have a look in some of the other coding projects to be inspired how to make a scoring class.

ErikSchierboom:
@bangsholt You might get away with copying a lot from the [Java example implementation](https://github.com/exercism/java/blob/master/exercises/bowling/.meta/src/reference/java/BowlingGame.java).

bangsholt:
It seems to be the same implementation as the example.cs, it just throws instead of returning null, which solves some of the tests. The rest seems to be needing a bit more logic.

ErikSchierboom:
@bangsholt Okay, good luck!

bangsholt:
Hey @ErikSchierboom , I'm sorry but I don't find it enjoyable at all to try to find a working solution for the bowling test. If that means that this won't be merged, so be it.

ErikSchierboom:
@bangsholt No problem. It does mean this won't be merged, as the example implementation does not pass the tests, but we can branch off where you left off. Thanks for the hard work!

bmeverett:
@ErikSchierboom I can take a look at fixing this up

jpreese:
Awesome, thanks @bmeverett for your work on this 👍

ErikSchierboom:
Closed due to #480 being merged.

PR #471, opened 2017-10-10
Added DifferenceOfSquares generator

felix91gr:
**Work I have left to do** is written on #442. I messed up that branch too much, so I deleted my fork and started it again.

### DIFF
--- a/generators/Exercises/DifferenceOfSquares.cs
+++ b/generators/Exercises/DifferenceOfSquares.cs
@@ -1,0 +1,25 @@
+using Generators.Input;
+using System;
+
+namespace Generators.Exercises
+{
+    public class DifferenceOfSquares : Exercise
+    {
+        protected override void UpdateCanonicalData(CanonicalData canonicalData)
+        {
+            foreach (var canonicalDataCase in canonicalData.Cases)
+                switch(canonicalDataCase.Property)
+                {
+                    case "squareOfSum":
+                        canonicalDataCase.Property = "CalculateSquareOfSum";
+                        break;
+                    case "sumOfSquares":
+                        canonicalDataCase.Property = "CalculateSumOfSquares";
+                        break;
+                    case "differenceOfSquares":
+                        canonicalDataCase.Property = "CalculateDifferenceOfSquares";
+                        break;
+                }
+        }
+    }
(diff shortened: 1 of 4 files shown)


felix91gr:
I finished testing this on my machine.

@ErikSchierboom and @jpreese please review this :)

felix91gr:
BTW, is there a way to run only one exercise's test suite? To wait less when I'm debugging my changes to one of them? (right now, I'm working on the `largest-series-product` one)

generators/Exercises/DifferenceOfSquares.cs
```diff
@@ -0,0 +1,28 @@
...
+            			canonicalDataCase.Property = "CalculateSquareOfSum";
+            			break;
+            		case "sumOfSquares":
+            			canonicalDataCase.Property = "CalculateSumOfSquares";
+            			break;
+            		case "differenceOfSquares":
+            			canonicalDataCase.Property = "CalculateDifferenceOfSquares";
+            			break;
+                    default:
+                        canonicalDataCase.Property = "Default_This_Shouldnt_Happen";
```
ErikSchierboom: Could you throw an exception here?

ErikSchierboom: The formatting of this switch is off (see the diff in GitHub).

felix91gr: Oh. Um, that `default` shouldn't even exist, actually. I just put it there to debug my changes to the property names. If another property comes, and has a new name, it shouldn't make the generator throw. Since this loop is an override of the original names, it should just ignore new names that come up.

felix91gr: I'll change that up

felix91gr:
There we go. I removed the default. And about this...

> The formatting of this switch is off (see the diff in GitHub).

I think I know what you mean. I've just found the diff (I was gonna ask you about it). Lemme fix it.

felix91gr:
There we go. It was tabs v/s spaces once again.

@ErikSchierboom: it's ready for revision :)

ErikSchierboom:
Great, thanks a lot! 🎉

Issue #412, opened 2017-09-06
Add Largest Series Product Test Generator


jpreese:
This issue is part of an overall initiative to complete #195.

The documentation on how to add generators can be found [here](https://github.com/exercism/csharp/blob/master/docs/GENERATORS.md). If you get stuck or have any questions about adding this generator, please do not hesitate to reach out!

felix91gr:
I'll tackle this one as well

felix91gr:
I finished this one as well :)
I'll wait for #471 to get merged before I open a PR though, because I branched off of that commit.

felix91gr:
It's here: #474

felix91gr:
Now that the PR's been merged, this issue can be closed :)

jpreese:
Awesome! Thanks, @felix91gr.

PR #472, opened 2017-10-11
Added Minesweeper generator

felix91gr:
> - Added generator itself
> - Updated generated Test file
> - Fixed Example and Minesweeper classes to recieve and return the right type on method Annotate.
  Was: string -> string. Now it is: string[] -> string[]

@ErikSchierboom please tell me I didn't screw up by changing `Annotate` from a `string -> string` function to a `string[] -> string[]` one? I don't know how else to interpret the Canonical Data [files](https://github.com/exercism/problem-specifications/blob/master/exercises/minesweeper/canonical-data.json).

If that's allright, then this commit is pretty much cooked afaik.

### DIFF
--- a/exercises/minesweeper/MinesweeperTest.cs
+++ b/exercises/minesweeper/MinesweeperTest.cs
@@ -1,142 +1,240 @@
-﻿using Xunit;
+// This file was auto-generated based on version 1.0.0 of the canonical data.
+
+using Xunit;
 
 public class MinesweeperTest
 {
     [Fact]
-    public void Zero_sized_board()
+    public void No_rows()
     {
-        var input = "";
-        var expected = "";
-
+        var input = new string[] 
+        { 
+             
+        };
+        var expected = new string[] 
+        { 
+             
+        };
         Assert.Equal(expected, Minesweeper.Annotate(input));
     }
 
     [Fact(Skip = "Remove to run test")]
-    public void Empty_board()
+    public void No_columns()
     {
-        var input = FormatInput(new[]
-        {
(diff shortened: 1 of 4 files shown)


felix91gr:
I branched this one off of `master`, so it's clean to be merged before #471 and #412

felix91gr:
I'm done with number 1.

> Format the arrays such that they are neatly aligned on different lines.

Now I'm on to this.

felix91gr:
@ErikSchierboom this one's ready as well ;)

exercises/minesweeper/MinesweeperTest.cs
```diff
@@ -1,142 +1,240 @@
...
 
 public class MinesweeperTest
 {
     [Fact]
-    public void Zero_sized_board()
+    public void No_rows()
     {
-        var input = "";
-        var expected = "";
+        var input = new []
```
ErikSchierboom: I think no rows should mean an empty array, not an array with an empty string. Could you replace that (also for the expected value)?

felix91gr: Sure!

felix91gr:
Okay, now I've fixed and tested it. The arrays it creates now are actually better, because they are strictly `new string[]`. 

Creating just `new []` is not a good idea imo, if we can avoid it. In this case, we _know_ that the input is exactly a `string` array, so I changed that as well.

Thanks for the feedback.
Tell me what you think of this one :)

ErikSchierboom:
Merged. Thanks! :tada:

PR #473, opened 2017-10-11
Added FlattenArray generator

felix91gr:
I have a problem here. The `null` values in the [Canonical Data](https://github.com/exercism/problem-specifications/blob/master/exercises/flatten-array/canonical-data.json) file for this exercise aren't rendered correctly at the moment of generating the test. In fact, they are ommited:

```json
{
      "description": "all values in nested list are null",
      "property": "flatten",
      "input": [null, [[[null]]], null, null, [[null, null], null], null],
      "expected": []
}
```
gets rendered as:

```csharp
public void All_values_in_nested_list_are_null()
{
        Assert.Empty(FlattenArray.Flatten(new[] { , new[] { new[] { new[] {  } } }, , , new[] { new[] { ,  },  },  }));
}
```

Is there a way to `UpdateCanonicalData` the nulls into this one?

Assuming there isn't, I digged around in the code base and the only thing I could find that might be amiss is the `ValueFormatter` class. That's the one in charge of rendering, but that's as far as I can go today.

Pls halp.

### DIFF
--- a/exercises/flatten-array/Example.cs
+++ b/exercises/flatten-array/Example.cs
@@ -1,6 +1,6 @@
 ﻿using System.Collections;
 
-public static class Flattener
+public static class FlattenArray
 {
     public static IEnumerable Flatten(IEnumerable input)
     {
--- a/exercises/flatten-array/FlattenArray.cs
+++ b/exercises/flatten-array/FlattenArray.cs
@@ -1,7 +1,7 @@
 ﻿using System;
 using System.Collections;
 
-public static class Flattener
+public static class FlattenArray
 {
     public static IEnumerable Flatten(IEnumerable input)
     {
--- a/exercises/flatten-array/FlattenArrayTest.cs
+++ b/exercises/flatten-array/FlattenArrayTest.cs
@@ -7,36 +7,140 @@ public class FlattenArrayTest
     [Fact]
     public void No_nesting()
     {
-        Assert.Equal(new[] { 0, 1, 2 }, FlattenArray.Flatten(new[] { 0, 1, 2 }));
+        var input = new[] { 0, 1, 2 };
+        var expected = new[] { 0, 1, 2 };
+        Assert.Equal(expected, FlattenArray.Flatten(input));
     }
 
     [Fact(Skip = "Remove to run test")]
     public void Flattens_array_with_just_integers_present()
     {
-        Assert.Equal(new[] { 1, 2, 3, 4, 5, 6, 7, 8 }, FlattenArray.Flatten(new[] { 1, new[] { 2, 3, 4, 5, 6, 7 }, 8 }));
+        var input = new object[] {
+          1,
+          new object[] {
+            2,
+            3,
+            4,
+            5,
+            6,
+            7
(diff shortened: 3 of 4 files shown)


ErikSchierboom:
Well, the canonical data often uses `null` to denote a missing value, which is why we don't actually render it. You could see what happens if you add it to the value formatter and re-run all the generators to see which ones break. If too much breaks, I would suggest converting the `null` values to `new EscapedValue("null")` instances.

felix91gr:
> I would suggest converting the null values to new EscapedValue("null") instances.

You mean `UnescapedValue`? Like I did in #472?

felix91gr:
Okay so, I tried this...
> You could see what happens if you add it to the value formatter and re-run all the generators to see which ones break

And I only broke 1 test on 1 exercise. That's nice. The bad thing is, now _this_ test doesn't compile XD

I'll investigate. Counsel is very much appreciated, because I'm only now understanding the `Exercise` codebase.

felix91gr:
An alternative could be to use `List<T>`, maybe... :thinking: That's how the original test is written in `master`:

```csharp
var nestedList = new List<object>
{
    null,
    new List<object>
    {
        null,
        new List<object> { null },
        new List<object> { new List<object> { new List<object> { null } } }
    },
    null
};
```

felix91gr:
Now that I look closely at the `two-bucket` exercise, it doesn't appear to have been broken by this commit. It includes absolutely _no nulls whatsoever_.

Was it broken on another commit?

ErikSchierboom:
@felix91gr I wouldn't worry about the two-bucket exercise, it is most likely due to some change in the canonical data. Just make sure you don't add that file to this PR.

I'm also perfectly fine with using `List<object>`, you can choose what you like.

felix91gr:
Okay, I reverted all the tests I generated in that run, except for the `flatten-array` one.

The problem seems to be that the type checker can't figure out a type for the arrays I'm giving it. That's why I suggested that it should probably be a `List`.

On the other hand... I just tested changing the arrays by hand from: `new[]` to `new object[]` and **it actually type-checks and passes the tests!**

Do you know how I can do this from the generator? @ErikSchierboom

ErikSchierboom:
Hmmm, I don't think there is an easy way to do that at the moment. I think you'll have to manually create the rendered string.

felix91gr:
I should use a template, right? Like for the Minesweeper generator. But I
don’t really understand those yet — the MS one needed to be almost exactly
the same as the one you sent me.

Where can I read about the syntax and capabilities of the templates?

If at all possible, I’d like to have a recursive one so that I can examine
the array thoroughly.

ErikSchierboom:
@felix91gr For templating we use [dotliquid](https://github.com/dotliquid/dotliquid/wiki/DotLiquid-for-Developers). However, as you want to do a recursive rendering, I think you should just do some string manipulation.

felix91gr:
I think I finally understood how `UnescapedValues` work. I've finished the generator now :)

ErikSchierboom:
Merged. Awesome!

PR #474, opened 2017-10-11
Added Largest Series Product Generator

felix91gr:
- Changed name of Property to `GetLargestProduct`, to fit with example and stub implementation
- Fixed issues in example implementation, that didn't handle the newest tests' requirements. In particular, erroneous input crashed the example, while the testing suite **expected it to return -1**. Now the example behaves properly.
- Updated `eol` to canonical (unintended, but the generator created it).

### DIFF
--- a/exercises/largest-series-product/LargestSeriesProductTest.cs
+++ b/exercises/largest-series-product/LargestSeriesProductTest.cs
@@ -1,171 +1,97 @@
-using System;
+// This file was auto-generated based on version 1.0.0 of the canonical data.
+
 using Xunit;
+using System;
 
 public class LargestSeriesProductTest
 {
     [Fact]
+    public void Finds_the_largest_product_if_span_equals_length()
+    {
+        Assert.Equal(18, LargestSeriesProduct.GetLargestProduct("29", 2));
+    }
+
+    [Fact(Skip = "Remove to run test")]
     public void Can_find_the_largest_product_of_2_with_numbers_in_order()
     {
-        const string digits = "0123456789";
-        const int span = 2;
-        const int expected = 72;
-
-        Assert.Equal(expected, LargestSeriesProduct.GetLargestProduct(digits, span));
+        Assert.Equal(72, LargestSeriesProduct.GetLargestProduct("0123456789", 2));
     }
 
     [Fact(Skip = "Remove to run test")]
     public void Can_find_the_largest_product_of_2()
     {
-        const string digits = "576802143";
-        const int span = 2;
-        const int expected = 48;
-
-        Assert.Equal(expected, LargestSeriesProduct.GetLargestProduct(digits, span));
-    }
-
-    [Fact(Skip = "Remove to run test")]
-    public void Finds_the_largest_product_if_span_equals_length()
-    {
-        const string digits = "29";
-        const int span = 2;
-        const int expected = 18;
-
-        Assert.Equal(expected, LargestSeriesProduct.GetLargestProduct(digits, span));
+        Assert.Equal(48, LargestSeriesProduct.GetLargestProduct("576802143", 2));
     }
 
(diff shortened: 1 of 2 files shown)


exercises/largest-series-product/Example.cs
```diff
@@ -22,6 +36,7 @@ private static long ParseDigit(char c)
     {
         if (!char.IsDigit(c))
         {
+            Console.WriteLine("Invalid digit: " + c);
```
ErikSchierboom: Could you remove this line?

exercises/largest-series-product/LargestSeriesProductTest.cs
```diff
@@ -1,171 +1,96 @@
...
+    [Fact(Skip = "Remove to run test")]
+    public void Reports_1_for_nonempty_string_and_empty_product_0_span_()
+    {
+        Assert.Equal(1, LargestSeriesProduct.GetLargestProduct("123", 0));
+    }
+
+    [Fact(Skip = "Remove to run test")]
+    public void Rejects_empty_string_and_nonzero_span()
+    {
+        Assert.Equal(-1, LargestSeriesProduct.GetLargestProduct("", 1));
```
ErikSchierboom: The `-1` is an error case, and should throw an exception. Same goes for the other `-1` cases.

generators/Exercises/LargestSeriesProduct.cs
```diff
@@ -0,0 +1,13 @@
+using Generators.Input;
+
+namespace Generators.Exercises
+{
+    public class LargestSeriesProduct : Exercise
+    {
+        protected override void UpdateCanonicalData(CanonicalData canonicalData)
+        {
+            foreach (var canonicalDataCase in canonicalData.Cases)
+                canonicalDataCase.Property = "GetLargestProduct";
```
ErikSchierboom: Maybe `CalculateLargestSeriesProduct` would be a better name.

exercises/largest-series-product/Example.cs
felix91gr: Yes!

exercises/largest-series-product/LargestSeriesProductTest.cs
felix91gr: So, if I'm understanding this, a `-1` in the `canonical-data.json` represents an exception?

generators/Exercises/LargestSeriesProduct.cs
felix91gr: Hmm. Maybe?

I interpret `GetLargestProduct` as *I wanna get the largest product in this series*, which fits the exercise description. Basically the same as saying `CalculateLargestSeriesProduct`, but more precise. I would leave it, but regardless, both convey the correct meaning.

exercises/largest-series-product/LargestSeriesProductTest.cs
felix91gr: Ah, yes. Indeed it does:

negative numbers in canonical data

exercises/largest-series-product/Example.cs
felix91gr: I reverted all of my changes to `Example.cs` :)

exercises/largest-series-product/LargestSeriesProductTest.cs
felix91gr: I added the `Assert.Throws`!

felix91gr:
I think I've fixed all of the issues. Thanks for the feedback!

generators/Exercises/LargestSeriesProduct.cs
ErikSchierboom: Well argued. I agree

ErikSchierboom:
Thanks!

felix91gr:
:boat: :D

PR #475, opened 2017-10-12
Fixing `ErrorHandling` exercise's asserts

felix91gr:
Every time I ran `./build.sh`, I got the following warnings:

> ErrorHandlingTest.cs(23,9): warning xUnit2003: Do not use Assert.Equal() to check for null value.
> ErrorHandlingTest.cs(33,9): warning xUnit2004: Do not use Assert.Equal() to check for boolean conditions.
> ErrorHandlingTest.cs(37,9): warning xUnit2004: Do not use Assert.Equal() to check for boolean conditions.
> ErrorHandlingTest.cs(59,9): warning xUnit2004: Do not use Assert.Equal() to check for boolean conditions.

This PR addresses those warnings, using respectively:
- `Assert.Null()`
- `Assert.True()`
- `Assert.False()`
- `Assert.True()`

### DIFF
--- a/exercises/error-handling/ErrorHandlingTest.cs
+++ b/exercises/error-handling/ErrorHandlingTest.cs
@@ -20,7 +20,7 @@ public class ErrorHandlingTest
         Assert.Equal(1, successfulResult);
 
         var failureResult = ErrorHandling.HandleErrorByReturningNullableType("a");
-        Assert.Equal(null, failureResult);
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
Wonderful!

felix91gr:
^^

robkeim:
Good catch, thanks for cleaning this up @felix91gr!

PR #479, opened 2017-10-17
Fix queen-attack generator after canonical-data change

HarrisonMc555:
A recent change to the queen-attack canonical data broke the test generator. This fixes it.

As a note, we might want to consider transitioning from `x` and `y` to `row` and `column`, as this is what the canonical data converged to.


ErikSchierboom:
@HarrisonMc555 Excellent! Feel free to send another PR to update from `x` and `y` to `row` and `column`.

PR #480, opened 2017-10-20
Finishes Bowling Test Generation

### LINKED ISSUE #402: Add Bowling Test Generator

bmeverett:
Finishes up #470 
Closes #402 

The `RenderTestMethodBodyAct` is a little messier then I'd like, but it gets the job done.

### DIFF
--- a/exercises/bowling/BowlingTest.cs
+++ b/exercises/bowling/BowlingTest.cs
@@ -1,6 +1,7 @@
 // This file was auto-generated based on version 1.0.1 of the canonical data.
 
 using Xunit;
+using System;
 using System.Collections.Generic;
 
 public class BowlingTest
@@ -151,9 +152,7 @@ public class BowlingTest
         var sut = new BowlingGame();
         var previousRolls = new int[] {};
         DoRoll(previousRolls, sut);
-        sut.Roll(-1);
-        var actual = sut.Score();
-        Assert.Null(actual);
+        Assert.Throws<ArgumentException>(() => sut.Roll(-1));
     }
 
     [Fact(Skip = "Remove to run test")]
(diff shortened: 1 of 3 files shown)


exercises/bowling/BowlingTest.cs
```diff
@@ -1,207 +1,303 @@
...
+
+    [Fact(Skip = "Remove to run test")]
+    public void Rolls_cannot_score_negative_points()
+    {
+        var sut = new BowlingGame();
+        var previousRolls = new int[] {};
+        DoRoll(previousRolls, sut);
+        sut.Roll(-1);
+        var actual = sut.Score();
+        Assert.Null(actual);
```
ErikSchierboom: I would prefer the methods that have erroneous input (see the [canonical-data](https://github.com/exercism/problem-specifications/blob/master/exercises/bowling/canonical-data.json#L133)) to throw an exception, instead of returning `null`.

generators/Exercises/Bowling.cs
```diff
@@ -0,0 +1,87 @@
...
+    {
+        private const string PreviousRolls = "previous_rolls";
+
+        protected override void UpdateCanonicalData(CanonicalData canonicalData)
+        {
+            foreach (var canonicalDataCase in canonicalData.Cases)
+            {
+                if (!(canonicalDataCase.Expected is long))
+                {
+                    canonicalDataCase.Expected = null;
```
ErikSchierboom: See comment above, I would like to throw an exception in this case.

generators/Exercises/Bowling.cs
```diff
@@ -0,0 +1,87 @@
...
+                }
+                canonicalDataCase.SetInputParameters();
+                canonicalDataCase.UseVariableForTested = true;
+                canonicalDataCase.Property = "score";
```
ErikSchierboom: When we switch to using exceptions, this line should be removed as the actual property being tested is then not `score`, but `roll`.

ErikSchierboom:
@bmeverett By the way, the older [PR](https://github.com/exercism/csharp/pull/470/files) for this generator shows you how to use exceptions.

bmeverett:
@ErikSchierboom I implemented your fixes. I changed the test cases to throw exceptions, and in the case of roll I implemented the function to check for valid input so that it would throw exceptions for invalid input.

ErikSchierboom:
@bmeverett Looking great! I'll wait for CI to become green and then I'll merge.

ErikSchierboom:
Merged. Thanks! 🎉

PR #481, opened 2017-10-20
Add RenderAdditionalMethods() example to GENERATORS.md

deyshin:
I found an example that utilizes `RenderAdditionalMethods()`. 
Having this example in the doc will make it easier for the readers to understand

### DIFF
--- a/docs/GENERATORS.md
+++ b/docs/GENERATORS.md
@@ -87,6 +87,8 @@ Allow additional methods to be added to the test suite.
 
 There may exist cases where a suite of unit tests will need to reuse the same logic in each of the tests. Rather than duplicating code, this method allows you to provide helper methods for the tests.
 
+An example of this is the [Tournament](https://github.com/exercism/csharp/blob/master/generators/Exercises/Tournament.cs)
+
 Additional methods added using this override will be added to the bottom of the test suite.
 
 ## Updating Existing Files


docs/GENERATORS.md
```diff
@@ -87,6 +87,8 @@ Allow additional methods to be added to the test suite.
 
 There may exist cases where a suite of unit tests will need to reuse the same logic in each of the tests. Rather than duplicating code, this method allows you to provide helper methods for the tests.
 
+An example of this is the [Tournament](https://github.com/exercism/csharp/blob/master/generators/Exercises/Tournament.cs)
```
ErikSchierboom: Good addition. Two small points.

1. Could you link to the line where the method is overridden? The link is: https://github.com/exercism/csharp/blob/master/generators/Exercises/Tournament.cs#L45
1. Could you add " generator." to the end of the line, so that is reads "An example of this is the Tournament generator."?

robkeim:
Thanks for taking the time to improve the documentation @deyshin... I know that tends to not be the most thrilling part of development for most people, but it's very important! Changes look good to me with @ErikSchierboom's edits.

deyshin:
Thanks for the feedback, @ErikSchierboom! 
@robkeim, I applied the requested changes.

ErikSchierboom:
Looks good! Merged. Thanks for taking the time to do this!

deyshin:
My pleasure, @ErikSchierboom!

PR #483, opened 2017-10-24
two-fer: "unlock_by": null

rpottsoh:
Not "unlock_by": "null"  This crashes `configlet tree .`.  See exercism/configlet#102

### DIFF
--- a/config.json
+++ b/config.json
@@ -64,7 +64,7 @@
         "optional_values",
         "strings"
       ],
-      "unlocked_by": "null",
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
Good catch!

ErikSchierboom:
But you did make the effort of creating a PR, so thanks for that! And thanks to @nywilken for finding it.

rpottsoh:
You're welcome!  Happy to help!

Issue #486, opened 2017-11-12
Add reverse-string exercise


robkeim:
Description can be found here:
https://github.com/exercism/problem-specifications/blob/master/exercises/reverse-string/description.md

Canonical-data here:
https://github.com/exercism/problem-specifications/blob/master/exercises/reverse-string/canonical-data.json

robkeim:
I'll take this one

ErikSchierboom:
Closed by #490

PR #490, opened 2017-11-12
Add reverse string exercise

robkeim:
For number #486 

The only thing I wasn't certain on was regarding the track:
- Should this be in the main track?
- Should this unlock any other exercises?


ErikSchierboom:
I think it is fine as it is now: a non-core exercise that is unlocked by submitting the first exercise.

ErikSchierboom:
Merged. Thanks!

PR #491, opened 2017-11-13
Single exercise option

### LINKED ISSUE #434: Add the exercises without canonical data to test generator
Here's what I shared about this in #396:

Right now we have to manually check to see if an exercise gets canonical data in the future. I'd like to have it integrated into the generator with some sort of flag saying "this doesn't have data" and then it can start throwing exceptions in the future if the data is created so we know we'll have to implement it (keeping my fingers crossed that all exercises will have canonical data someday)

ErikSchierboom:
This PR adds functionality to fix #434 (show status of all exercises). When running the test generator, all the track's exercises are checked to see what to do. There are four options:

1. There is no canonical data for the exercise. A warning is printed.
1. There is canonical data and no generator is available. An error is printed (just for display purposes, as it makes it easier to discern between the various types).
1. There is canonical data and a generator is available. The generator is run.
1. There is canonical data and a generator is available, but the generator inherits from `CustomExercise`. In this case, we assume that the tests are crafter manually and don't run a generator.

I've also added an option to filter on the above mentioned statuses, which you can do by providing a command-line option. For example, `dotnet run -s Unimplemented` would show only those exercises with canonical data but without a generator.


generators/Exercises/Diamond.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Generators.Exercises
+{
+    public class Diamond : CustomExercise
+    {
```
robkeim: It's probably worth adding a comment here on why we decided not to follow the canonical data for this exercise (here's it's because we wanted to do property based testing). That way if something changes in the future and we want to revisit that decision we have the info here without digging through the issues.

generators/Program.cs
```diff
@@ -35,30 +35,69 @@ private static void SetupLogger()
...
-                CacheCanonicalData = options.CacheCanonicalData
-            };
-            return new CanonicalDataParser(canonicalDataOptions);
+                case GeneratorExercise generatorExercise:
+                    var canonicalData = canonicalDataParser.Parse(exercise.Name);
+                    generatorExercise.Regenerate(canonicalData);
+
+                    Log.Information("{Exercise}: tests generated", exercise.Name);
+                    break;
+                case UnimplementedExercise _:
```
robkeim: Can't you remove the underscores here since they're not used?

robkeim:
Awesome work @ErikSchierboom... you took what I had in mind even a step further that's great! I added a couple of minor comments, but feel free to merge yourself even if you make changes for my comments as they're minor.

generators/Program.cs
ErikSchierboom: Nope, unfortunately not. I tried that at first, but then it isn't a valid pattern match anymore.

generators/Exercises/Diamond.cs
ErikSchierboom: Good point. I've updated the PR.

generators/Program.cs
robkeim: Hmm ok I thought you could, thanks for checking!

robkeim:
Thanks @ErikSchierboom!

PR #492, opened 2017-11-13
Update exercises

### LINKED ISSUE #488: Update exercises with latest canonical data
I started the first batch of this work in #487 but there are still three exercises remaining (as of Nov 12):
- AllYourBase (changed to using "error" structure)
- BeerSong (changed to returning an array which isn't currently being formatted multi-line. That means the test validating the entire song is a reaaalllllyyyyy long line :) I think we should update the array to be able to be formatted multiline)
- NthPrime (changed to using "error" structure)

@ErikSchierboom FYI

ErikSchierboom:
Closes #488


robkeim:
Changes look good to me @ErikSchierboom. Waiting for CI to finish before merging.

PR #496, opened 2017-11-19
Update dominoes README

### LINKED ISSUE #190: dominoes: ask for Chain rather than CanChain OR modify README to ask for CanChain
There are two possible ways to interpret the dominoes problem, as [the canonical data](https://github.com/exercism/x-common/blob/master/exercises/dominoes/canonical-data.json) tells us.

1. Only ask for `CanChain`, a boolean telling us whether the given dominoes can be chained. The tests only need to check whether the boolean matches the expectation.
2. Ask for `Chain`, which will be (I assume for C#) a valid chain if one exists, or `null` if one does not. The tests would need to check the various properties noted in [the canonical data](https://github.com/exercism/x-common/blob/master/exercises/dominoes/canonical-data.json#L27-L33)

This track asks for `CanChain` only. **I assumed this was an intentional decision**, so I did not question it. But Erik asked me to file an issue. So y'all can reopen the issue in case y'all want to ask for `Chain`.

ErikSchierboom:
Closes #190

### DIFF
--- a/exercises/dominoes/README.md
+++ b/exercises/dominoes/README.md
@@ -1,24 +1,15 @@
 # Dominoes
 
-Make a chain of dominoes.
-
-Compute a way to order a given set of dominoes in such a way that they form a
+Compute whether there exists a way to order a given set of dominoes in such a way that they form a
 correct domino chain (the dots on one half of a stone match the dots on the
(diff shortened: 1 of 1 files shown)


robkeim:
Changes look good to me thanks @ErikSchierboom!

PR #503, opened 2017-12-30
Add back deleted, deprecated exercises

kytrinyx:
When migrating the data from v1 to v2, we will lose all the solutions submitted to
deprecated exercises, unless they exist in the track configuration.

This adds the deprecated binary, trinary, octal, and hexadecimal exercises back,
leaving them deprecated so that they don't actively show up as part of the track.

We can delete them from the track again after we launch v2, if we want to.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1240,6 +1240,30 @@
       ],
       "unlocked_by": "markdown",
       "uuid": "a5794706-58d2-48f7-8aab-78639d7bce77"
+    },
+    {
+      "core": false,
+      "deprecated": true,
+      "slug": "binary",
+      "uuid": "cef7deef-54ce-4201-b263-7cd2098533f8"
+    },
+    {
+      "core": false,
+      "deprecated": true,
+      "slug": "trinary",
+      "uuid": "c7dd8467-87e2-4997-a96e-a04cb8b891e8"
+    },
+    {
+      "core": false,
+      "deprecated": true,
+      "slug": "octal",
+      "uuid": "c8555f37-9976-4f52-a5db-6a680ec8d53b"
+    },
+    {
+      "core": false,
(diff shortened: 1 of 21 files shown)


ErikSchierboom:
@kytrinyx Do they have to be actually working exercises? As of now, they don't follow the newly introduced project format. That is at the least inconsistent, but possibly also erroneous. Should people that once worked on these exercises be allowed to work on them again?

iHiD:
> Do they have to be actually working exercises?

Yes in terms of they need the files etc. No in terms of them needing all the same level of config (e.g. topics, difficulty, etc).

I'm 👍 to this @kytrinyx, but let's check it in a sample ETL run to be sure.

kytrinyx:
> As of now, they don't follow the newly introduced project format.

Hm. Yeah, they do need to actually work. I'll read through the track documentation and get this to conform.

kytrinyx:
@ErikSchierboom Would you take a look to see if this is properly following the project structure?

kytrinyx:
The build is failing because I've not yet converted the test suite. I'll give that a go next.

ErikSchierboom:
Thanks for doing this! I'll take a look tomorrow. By the way, you can verify the build by running build.sh or build.cmd (also works for the F# track).

kytrinyx:
> you can verify the build by running build.sh or build.cmd (also works for the F# track).

Ah! That's helpful thanks!

ErikSchierboom:
Oh, and you can test build single exercises using --exercise=<slug>

ErikSchierboom:
That is --exercise=slug

kytrinyx:
Sweet, thanks!

ErikSchierboom:
Thanks a lot!

PR #504, opened 2018-01-03
Updated test to remove false positives

Monomachus:
* Had a case that wasn't in the requirement
* For a question we had a false response

### DIFF
--- a/exercises/bob/BobTest.cs
+++ b/exercises/bob/BobTest.cs
@@ -43,7 +43,7 @@ public class BobTest
     [Fact(Skip = "Remove to run test")]
     public void Talking_forcefully()
     {
-        Assert.Equal("Whatever.", Bob.Response("Let's go make out behind the gym!"));
+        Assert.Equal("Whoa, chill out!", Bob.Response("Let's go make out behind the gym!"));
     }
 
     [Fact(Skip = "Remove to run test")]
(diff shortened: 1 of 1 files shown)


robkeim:
Hello @Monomachus thanks for your PR!

We actually share all of our test cases across the different tracks and you can find the test cases for this exercise here:
https://github.com/exercism/problem-specifications/blob/master/exercises/bob/canonical-data.json

I think what's happening here is the description is out-of-date and should be updated with the one that's located here:
https://github.com/exercism/problem-specifications/blob/master/exercises/bob/description.md

If you read that description does the existing solution make sense? If so, would you mind updating this PR to update the Readme.md for the Bob exercise instead?
https://github.com/exercism/csharp/tree/master/exercises/bob

Monomachus:
This case is not correct in my opinion:
Statement: "Let's go make out behind the gym!"
Response: "Whatever."

If you yell at him, response -  'Whoa, chill out!'

What do you think?

ErikSchierboom:
@Monomachus As @robkeim indicated, the test cases for this exercise are generated from the shared problem specifications. You can find the canonical data for the bob exercise here: https://github.com/exercism/problem-specifications/blob/master/exercises/bob/canonical-data.json

Could you verify that the issue you are having is only in the C#? If it is also present in the canonical data, you should try opening an issue in the problem-specifications repository: https://github.com/exercism/problem-specifications/issues When things are fixed there, we can easily update our code.

PR #509, opened 2018-01-09
Extended and corrected the about section on C#

Warlaan:
Some claims about C# were imho incorrect, others were not on point (for example C#'s type inference is certainly one of its weakest features).

### DIFF
--- a/docs/ABOUT.md
+++ b/docs/ABOUT.md
@@ -1,5 +1,7 @@
-C# is a modern, object oriented language developed and maintained by Microsoft. The learning curve is small, reduced even if the developer has experience with other 'C-Style' languages.
+C# is a multiparadigm language with lots of features. While at its core it's still an object oriented language that's modeled after Java it supports first class functions, SQL-like syntax to access containers, automatic getters and setters, pattern matching, direct memory access in an otherwise garbage collected environment, statically typed, dynamically typed and (to a certain degree) type inferred programming and many other features in a constantly growing feature set.
 
-C# is a carefully designed language with lots of great features, such as type-inference and async/await. The official [documentation](https://docs.microsoft.com/en-us/dotnet/csharp/) is very extensive and well-written.
+C# is typically compiled to Microsoft's IL code, which means that it can interact seamlessly with other such languages like F#, Visual Basic or derivates of otherwise non-IL-compatible languages (https://en.wikipedia.org/wiki/List_of_CLI_languages).
 
-Although .NET used to be Windows-only, with the release of [.NET Core](https://www.microsoft.com/net/core) you can also use C# on Mac or Unix-based systems.
+That IL code is run in a virtual machine, and while the default implementation by Microsoft (the ".NET framework runtime") is supplied for Windows only, the open source [Mono project](http://www.mono-project.com/) and the new [.NET Core](https://www.microsoft.com/net/core) project by Microsoft allow C# to be used on other platforms like MacOS or Unix-based systems.
+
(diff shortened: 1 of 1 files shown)


docs/ABOUT.md
```diff
@@ -1,5 +1,7 @@
...
 
-C# is a carefully designed language with lots of great features, such as type-inference and async/await. The official [documentation](https://docs.microsoft.com/en-us/dotnet/csharp/) is very extensive and well-written.
+C# is typically compiled to Microsoft's IL code, which means that it can interact seamlessly with other such languages like F#, Visual Basic or derivates of otherwise non-IL-compatible languages (https://en.wikipedia.org/wiki/List_of_CLI_languages).
 
-Although .NET used to be Windows-only, with the release of [.NET Core](https://www.microsoft.com/net/core) you can also use C# on Mac or Unix-based systems.
+That IL code is run in a virtual machine, and while the default implementation by Microsoft (the ".NET framework runtime") is supplied for Windows only, the open source [Mono project](http://www.mono-project.com/) and the new [.NET Core](https://www.microsoft.com/net/core) project by Microsoft allow C# to be used on other platforms like MacOS or Unix-based systems.
```
ErikSchierboom: I don't think mentioning IL code offers much. Maybe rephrase to say that C# runs on the CLR, on which F# and Visual Basic also target. The point to make here is that the language is compatible with or CLR languages, which you do make, but I would prefer to have it slightly more concise (and without the IL code detail).

ErikSchierboom: I don't really think the .NET Framework is the default implementation anymore. .NET Core is what is being pushed at the default. Note that this document is intended so sell people on diving into the C# track on Exercism. It is not a goal to be as complete as possible.

docs/ABOUT.md
```diff
@@ -1,5 +1,7 @@
-C# is a modern, object oriented language developed and maintained by Microsoft. The learning curve is small, reduced even if the developer has experience with other 'C-Style' languages.
+C# is a multiparadigm language with lots of features. While at its core it's still an object oriented language that's modeled after Java it supports first class functions, SQL-like syntax to access containers, automatic getters and setters, pattern matching, direct memory access in an otherwise garbage collected environment, statically typed, dynamically typed and (to a certain degree) type inferred programming and many other features in a constantly growing feature set.
```
ErikSchierboom: I think multiparadigm should be multi-paradigm.

ErikSchierboom: Although it was indeed modelled after Java, it has long since surpassed that language. Hence, I think mentioning that it is a 'C-Style language' is a better wording.

ErikSchierboom: Should we mention dynamic typing? I hardly see anyone using it, so is this really a big advantage?

ErikSchierboom: Maybe reword the type inference a bit to something like: "and supports basic type inference"?

ErikSchierboom: I think we should definitely add a mention of async/await in some way, for example by mentioning that C# has an elegant, simple way to handling asynchrony.

ErikSchierboom: I'm not sure automatic getters and setters should be in the list. They seem to be a relatively minor feature.

robkeim:
I feel this is a step in the right direction because it's offering more information about C#, but I agree with @ErikSchierboom that it doesn't make a great sales pitch. Reading through this, I don't think it would convince anyone that C# would be a track that they'd want to learn.

ErikSchierboom:
Thanks a lot for the hard work. There were some great ideas here, but this PR has been superseded by #512.

Issue #367, opened 2017-08-19
Update about information to make it more detailed


ErikSchierboom:
With Exercism V2, a language track's about page can be quite detailed. With the help of markdown, the resulting page can look beautiful and have quite a bit of information. Our about page currently renders as follows on the V2 website: https://v2.exercism.io/tracks/csharp

It's not bad, but consider the Haskell track's page: https://v2.exercism.io/tracks/haskell
Much better IMHO! I think we should follow Haskell's lead here and be more verbose on what makes C# great, by showcasing a couple of C#'s best features (tooling, LINQ, async/await, cross-platform, etc.)

The about information can be found in `docs/ABOUT.md`.

robkeim:
I just checked out v2 of the site for the first time... it's looking awesome!

ErikSchierboom:
It is, isn't it? If you decide to check it out and find any issues, you can post them [here](https://github.com/exercism/v2-feedback/issues).

Warlaan:
The about page seriously needs an update. There are a lot of good things to be said about C#, but half of the things currently on the about page I strongly disagree with.

"The learning curve is small, reduced even if the developer has experience with other 'C-style' languages." I have no idea how C# managed to pull off that popular misconception, but it's just plain wrong.

I would say: "C# is designed to resemble other 'C-style' languages, allowing you to be productive very soon.". Saying that the learning curve was small is just plain wrong. Just try to explain the difference between classes and structs, or list and explain the different access modifiers. There's just more to learn in C# than in almost any other language. Which is actually the one thing that is so great about that language - if there is a feature that could be implemented it probably already is. First class functions? Check. Writing unmanaged code in a managed language? Check. Language interoperability? Check. 

So if there are so many good things to be said there's no need for false advertising. And there certainly are better features to be listed than C#'s "type inference" (try F# as a comparison - what C# has is not type inference, it's just 'not typing the type name twice'). And async/await is technically not a feature but a pair of keywords.

Also if you mention .Net Core you should also mention that the Mono project has made C# available cross platform for several years now.

I don't have time right now, but I'll come back later and post a suggestion for new version of the text.

ErikSchierboom:
@Warlaan Thanks for the feedback! Looking forward to your suggested changes.

Warlaan:
I created a PR with my suggested text: https://github.com/exercism/csharp/pull/509

martinfreedman:
[...]

PR #512, opened 2018-01-10
Update ABOUT.md

martinfreedman:
In reference to issue #367 here are my suggestions. Fully proof read now!

### DIFF
--- a/docs/ABOUT.md
+++ b/docs/ABOUT.md
@@ -1,4 +1,4 @@
-C# is a multi-paradigm statically-typed programming language with object-oriented, declarative, functional, generic, lazy integrated querying features and type inference. 
+C# is a multi-paradigm, statically-typed programming language with object-oriented, declarative, functional, generic, lazy, integrated querying features and type inference. 
 
 __Statically-typed__ means that identifiers have a type set at compile time--like those in Java, C++ or Haskell--instead of holding data of any type like those in Python, Ruby or JavaScript.
 
@@ -8,16 +8,16 @@ __Declarative__ means programming what is to be done, as opposed to how it is do
 
 __Functional__ means that functions are first-class data types that can be passed as arguments to and returned from other functions.
 
-__Generic__ means that algorithms are written in terms of types to-be-specified-later that are then instantiated when needed for specific types provided as parameters. 
+__Generic__ means that algorithms are written in terms of types to-be-specified-later that are then instantiated, when needed, for the specific types provided as parameters. 
 
-__Lazy__ (a.k.a "non-strict") means that the compiler will put off evaluating a thing until absolutely neccessary. This lets you safely do weird stuff like operating on an infinite list--the language will only create it up to the last value you actually use.
+__Lazy__ (a.k.a "deferred execution") means that the compiler will put off evaluating an item until required. This lets one safely do weird stuff like operating on an infinite list--the language will only create the list up to the last value needed.
 
-__Integrated Querying__ means the language feature called [LINQ "Language-Integrated Query"](https://msdn.microsoft.com/en-us/library/bb308959.aspx), which enables querying directly within the language, not only its own objects but, also, external data sources through formats such as XML, Json, SQL, NoSQL DBs and event streams.  
+__Integrated Querying__ means the language feature called [LINQ "Language-Integrated Query"](https://msdn.microsoft.com/en-us/library/bb308959.aspx), which enables lazy querying directly within the language, not only its own objects but, also, external data sources through formats such as XML, JSON, SQL, NoSQL DBs and event streams.  
 
 __Type inference__ means that the compiler will often figure out the type of an identifier by itself so you don't have to specify it. Scala and F# both do this.
 
 __Syntax__ is similar to that of other C-style languages such as C, C++ and Java.
 
-__.NET__ is the managed environment within which C# runs and, so, you get access to the entire .NET ecosystem, including all packages on [nuget.org](http://www.nuget.org). .NET used to be Windows-only but, with the release of [.NET Core](https://www.microsoft.com/net/core) -- as well as [MONO](http://www.mono-project.com/) -- you can also use C# on Mac, Linux or Unix-based systems. 
+__.NET__ is the managed environment within which C# runs and, so, you get access to the entire .NET ecosystem, including all packages on [nuget.org](http://www.nuget.org). .NET used to be Windows-only but, with the release of [.NET Core](https://www.microsoft.com/net/core) -- as well as [Mono](http://www.mono-project.com/) -- you can also use C# on Mac, Linux or Unix-based systems and on mobile platforms too. 
 
 C# also has features, amongst others, to make programming multiple threads, multiple processors, parallelisation, asynchrony, unmanaged code in a managed environment and language interoperability easier. It is developed and maintained by Microsoft, who provides the official [documentation](https://docs.microsoft.com/en-us/dotnet/csharp/).
--- a/docs/ABOUT.md
+++ b/docs/ABOUT.md
@@ -20,4 +20,4 @@ __Syntax__ is similar to that of other C-style languages such as C, C++ and Java
 
 __.NET__ is the managed environment within which C# runs. This used to be Windows-only but, with the release of [.NET Core](https://www.microsoft.com/net/core) -- as well as [MONO](http://www.mono-project.com/) -- you can also use C# on Mac, Linux or Unix-based systems. 
 
-C# also has features, amongst others, to enable programming multiple threads, multiple processes, parallelisation, asynchrony, unmanaged code in a managed environment and language interoperability easier. It is developed and maintained by Microsoft, who provides the official [documentation](https://docs.microsoft.com/en-us/dotnet/csharp/).
+C# also has features, amongst others, to make programming multiple threads, multiple processors, parallelisation, asynchrony, unmanaged code in a managed environment and language interoperability easier. It is developed and maintained by Microsoft, who provides the official [documentation](https://docs.microsoft.com/en-us/dotnet/csharp/).
--- a/docs/ABOUT.md
+++ b/docs/ABOUT.md
@@ -20,4 +20,4 @@ __Syntax__ is similar to that of other C-style languages such as C, C++ and Java
 
 __.NET__ is the managed environment within which C# runs. This used to be Windows-only but, with the release of [.NET Core](https://www.microsoft.com/net/core) -- as well as [MONO](http://www.mono-project.com/) -- you can also use C# on Mac, Linux or Unix-based systems.
 
-C# is developed and maintained by Microsoft and provides the official [documentation](https://docs.microsoft.com/en-us/dotnet/csharp/).
+C# is developed and maintained by Microsoft, who provides the official [documentation](https://docs.microsoft.com/en-us/dotnet/csharp/).
--- a/docs/ABOUT.md
+++ b/docs/ABOUT.md
@@ -18,6 +18,6 @@ __Type inference__ means that the compiler will often figure out the type of an 
 
 __Syntax__ is similar to that of other C-style languages such as C, C++ and Java.
 
-__.NET__ is the managed environment within which C# runs and, so, you get access to the entire .NET ecosystem, including all packages on [nuget.org](http://www.nuget.org). .NET used to be Windows-only but, with the release of [.NET Core](https://www.microsoft.com/net/core) -- as well as [Mono](http://www.mono-project.com/) -- you can also use C# on Mac, Linux or Unix-based systems and on mobile platforms too. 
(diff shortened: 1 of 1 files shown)


docs/ABOUT.md
```diff
@@ -1,5 +1,23 @@
-C# is a modern, object oriented language developed and maintained by Microsoft. The learning curve is small, reduced even if the developer has experience with other 'C-Style' languages.
+C# is a multi-paradigm statically-typed programming language with object-oriented, declarative, functional, generic, lazy integrated querying features and type inference.
```
ErikSchierboom: Please add a comma between "multi-paradigm" and "statically-typed"

ErikSchierboom: Please add a comma between "lazy" and "integrated querying"

docs/ABOUT.md
```diff
@@ -1,5 +1,23 @@
...
+
+__Declarative__ means programming what is to be done, as opposed to how it is done (which is an implementation detail which can distract from the domain or business logic).
+
+__Functional__ means that functions are first-class data types that can be passed as arguments to and returned from other functions.
+
+__Generic__ means that algorithms are written in terms of types to-be-specified-later that are then instantiated when needed for specific types provided as parameters. 
+
+__Lazy__ (a.k.a "non-strict") means that the compiler will put off evaluating a thing until absolutely neccessary. This lets you safely do weird stuff like operating on an infinite list--the language will only create it up to the last value you actually use.
+
+__Integrated Querying__ means the language feature called [LINQ "Language-Integrated Query"](https://msdn.microsoft.com/en-us/library/bb308959.aspx), which enables querying directly within the language, not only its own objects but, also, external data sources through formats such as XML, Json, SQL, NoSQL DBs and event streams.
```
ErikSchierboom: Json -> JSON

docs/ABOUT.md
```diff
@@ -1,5 +1,23 @@
...
+
+__Integrated Querying__ means the language feature called [LINQ "Language-Integrated Query"](https://msdn.microsoft.com/en-us/library/bb308959.aspx), which enables querying directly within the language, not only its own objects but, also, external data sources through formats such as XML, Json, SQL, NoSQL DBs and event streams.  
+
+__Type inference__ means that the compiler will often figure out the type of an identifier by itself so you don't have to specify it. Scala and F# both do this.
+
+__Syntax__ is similar to that of other C-style languages such as C, C++ and Java.
+
+__.NET__ is the managed environment within which C# runs and, so, you get access to the entire .NET ecosystem, including all packages on [nuget.org](http://www.nuget.org). .NET used to be Windows-only but, with the release of [.NET Core](https://www.microsoft.com/net/core) -- as well as [MONO](http://www.mono-project.com/) -- you can also use C# on Mac, Linux or Unix-based systems.
```
ErikSchierboom: MONO -> Mone

ErikSchierboom: "C# runs and, so, you get access" ->"C# runs, so you get access"

docs/ABOUT.md
```diff
@@ -1,5 +1,23 @@
...
+
+__Type inference__ means that the compiler will often figure out the type of an identifier by itself so you don't have to specify it. Scala and F# both do this.
+
+__Syntax__ is similar to that of other C-style languages such as C, C++ and Java.
+
+__.NET__ is the managed environment within which C# runs and, so, you get access to the entire .NET ecosystem, including all packages on [nuget.org](http://www.nuget.org). .NET used to be Windows-only but, with the release of [.NET Core](https://www.microsoft.com/net/core) -- as well as [MONO](http://www.mono-project.com/) -- you can also use C# on Mac, Linux or Unix-based systems. 
+
+C# also has features, amongst others, to make programming multiple threads, multiple processors, parallelisation, asynchrony, unmanaged code in a managed environment and language interoperability easier. It is developed and maintained by Microsoft, who provides the official [documentation](https://docs.microsoft.com/en-us/dotnet/csharp/).
```
ErikSchierboom: "programming multiple threads, multiple processors" -> "programming with multiple threads/processors"

robkeim:
I really like this setup and I think the individual sections could be strengthened by adding links for people that want to learn more (although that could easily come in a future PR).

If I didn't know what functional programming was for example the first thing I'd do after reading this is probably Google it. We could save the reader that step by linking them somewhere that made sense.

ErikSchierboom:
> I think the individual sections could be strengthened by adding links for people that want to learn more

Excellent point! @martinfreedman would you be willing to add links to the various sections? There are also some open comments I made, but they are mostly trivial.

docs/ABOUT.md
martinfreedman: and will add to "enables querying" "enables lazy querying"?

martinfreedman: Added deferred instead of non-strict and a couple of minor tweaks to the language (only on bits copied from the Haskell page)

martinfreedman:
Done, (not sure how to automatically indicate that PR has been updated here).

docs/ABOUT.md
martinfreedman: First comma yes, the second was deliberate as it is only LINQ that is lazy (although one could add Tasks etc. I suppose).

martinfreedman:
I probably did  not integrate your PR comments the right way but they should all be there now

ErikSchierboom:
Thanks a lot @martinfreedman! We should probably open a separate issue to add links to the about page.

Warlaan:
It's definitely a lot better now!

But - I hope I don't sound like I just like to complain, but I think the essence of C# could still be described better. The thing is that C#'s strength is that it sits in between the chairs, so listing paradigms will leave you with a wrong impression.

For example lazy evaulation - it's true that C# has it, but only when using yield or LINQ. Compare that to Haskell where everything is evaluated lazily by default, or F# where you can wrap any code in "lazy (...)" and receive a result of Lazy<T>.

Or take type inference - in languages like F# or Haxe you can write whole classes without specifying any type, because everything is inferred. In C# there's only one specific case where you can have the type inferred, and that's when you are declaring a variable and initializing it to a value in the same line.

And declarative - if someone disagrees with me I'd really be interested in their opinion, but does C# really have significantly more declarative elements than other imperative languages? Sure, it's got LINQ, but apart from that it's the same combination of declarative class structures with imperative method bodies that you have in any similar language. Compare that to true declarative languages like HTML or the visual languages used in many game engines to build scenes and there's just a huge difference.

So yes, C# has these elements, but it's the fact that it does borrow from other paradigms that characterizes it. The paradigms itself aren't really used enough to affect the general "tone" of the language.

Last but not least one small thing: what are "imperative class-based objects"? Class declarations are declarative (hence "declarations") and method bodies are imperative. If anything C#'s objects are less imperative than those of prototype based languages like Lua or JavaScript, since in those languages classes are not declared but created imperatively (does not apply to 100% to JavaScript).

So I hope I don't spoil the mood, I just wanted to give my feedback in case someone wants to improve the text further and agrees with me.

martinfreedman:
@Warlaan I have no issue with any of your points. Play with it

robkeim:
@Warlaan thanks for your detailed feedback!

Before going and making more changes, I'd take a step back and ask the questions:
1. What is the purpose of the ABOUT.md?
2. Who is the target audience for the ABOUT.md?

I think the desired result is going to vary significantly based on the answers to those questions. The second one might not be easy to answer since Exercism targets several different personas (a total noobie, someone trying to learn a new language, and someone that's trying to improve their skills in a language they alreayd know). If we're trying to target noobies or someone with little programming experience even the way that it's currently written we're going to make their heads spin by the time they reach the end of the first sentence.

@Warlaan's additions are more detailed and push even further into the realm of targeting people that have a fair amount of programming knowledge including diverse experience since there are many comparisons between other languages.

What do you guys feel should be the purpose this document and the target audience? @ErikSchierboom  @martinfreedman @Warlaan

ErikSchierboom:
Well, I think Exercism caters to people with *some* coding experience, but that can actually be very little. I would prefer to keep things simple and not necessarily go into the nitty-gritty details. Remember, this is just a document to entice people to start working on the language track, not a page with a detailed analysis of the language. I also feel that if another language does things better (e.g. laziness of type-inference), that doesn't mean we cannot list it as a feature.

I'll respond to some of the comments in more detail tomorrow.

martinfreedman:
I agree with @ErikSchierboom . It is not a detailed technical argument. It is a brief and broad about the language.

ErikSchierboom:
> For example lazy evaulation - it's true that C# has it, but only when using yield or LINQ. Compare that to Haskell where everything is evaluated lazily by default, or F# where you can wrap any code in "lazy (...)" and receive a result of Lazy.

Yes Haskell does it better, but C#'s enumerable and `Lazy<T>` do help when dealing with lazy evaluation.

> Or take type inference - in languages like F# or Haxe you can write whole classes without specifying any type, because everything is inferred. In C# there's only one specific case where you can have the type inferred, and that's when you are declaring a variable and initializing it to a value in the same line.

Yes, there are languages that have much better and extensive type-inference, but that doesn't mean it isn't a nice feature of C#. Try doing some Java or C(++) code, and it quickly becomes a selling point.

And declarative - if someone disagrees with me I'd really be interested in their opinion, but does C# really have significantly more declarative elements than other imperative languages? Sure, it's got LINQ, but apart from that it's the same combination of declarative class structures with imperative method bodies that you have in any similar language. Compare that to true declarative languages like HTML or the visual languages used in many game engines to build scenes and there's just a huge difference.

I think having LINQ is enough reason to have it listed as being declarative. `async/await` is also quite declarative in that it hides concurrency implementation details and let's you focus on the general design.

> So yes, C# has these elements, but it's the fact that it does borrow from other paradigms that characterizes it. The paradigms itself aren't really used enough to affect the general "tone" of the language.

It may not be enough the characterize C#, but it does mean that those features are what makes C# be C#, and as such they warrant to be listed I think.

> Last but not least one small thing: what are "imperative class-based objects"? Class declarations are declarative (hence "declarations") and method bodies are imperative. If anything C#'s objects are less imperative than those of prototype based languages like Lua or JavaScript, since in those languages classes are not declared but created imperatively (does not apply to 100% to JavaScript).

I think you're right. Would you be willing to send a PR?

Warlaan:
My apologies, I simply wasn't aware of the Lazy<T>-type in C# - a bit silly, since I knew about the type in F# and it's obviously a .NET type, so F# and C# do have the same feature, F# just has a little syntactic sugar coating for it.

About the other points: I can only imagine two situations in which people use a page like exercism: 
1. They want to improve their skills in a specific language. In that case they jump to the exercises directly and don't read the about page.
2. They want to choose a language to look into. In that case they want a comparison, not marketing. 
I really don't understand the idea of "selling C#". It has been mentioned that C#'s type inference was a "selling point" or that earlier versions of the text weren't good "advertising". Why does C# need advertising? 

I love to look into new languages, I think it's a very valuable method to improve your coding skills. And when I do I search for some characteristics of the language first and then look into the most interesting features first.
Now if I decided to try a language with type inference C# would simply be a bad choice. If I wanted to see what declarative programming looked like using C# would in no way confront me with that thinking.

C# is the right language to have a look at if someone wants to see what it looks like to be able to use one language for lots of different programming styles.
That tiny bit of type inference and the recent addition of the dynamic type don't change how a typical desktop or backend programmer thinks about the language, but it allows to write C# code that is about as concise as the kind of code that you typically find in scripting languages. And that is what allows C# to be used in places where other backend languages would be inappropriate, e.g. as a scripting language in game engines.

LINQ doesn't make C# significantly more of a declarative or lazily evaluated language, but it allows C# to be used in a typical database-style workflow where queries are created first but evaluated at the time they are executed.

Just look at F#. Technically it's very similar to C#, there's basically nothing that you can do in F# that's technically impossible in C#, but the functional first approach confronts you with functional thinking. Technically the type inference is just as optional as it is in C#, but due to the different syntax it becomes a core feature that you will see in the very first tutorials you find.

So I am not criticising how much the current version is or isn't going into detail, I just think it's missing the point. C# excels at making it possible to write domain-typical code of different domains in one language. Thanks to LINQ and lazy evaluation you can use database-typical code to access different collections or databases. Thanks to keywords like var or dynamic you can write script-like code. Thanks to the unsafe keyword you can write code that closely resembles languages like C++. Thanks to the property syntax, reflection, attributes, additional visibility modifiers etc. it's easier to create a custom API (some would even call it "write a DSL") than in other languages.

Imho that's what people who look at this page should know. Idiomatic C# isn't type inferred, lazily evaluated or more declarative than similar code in other C-style languages. But it does change significantly with the workflow. 
Instead of saying "if you are looking for a language with type inference, lazy evaluation and declarative features have a look at C#" this page should say "if you are looking for an exceptionally versatile langauge that allows you to write in different styles more than any other language then have a look at C#".

martinfreedman:
I think this is a great opening line ""if you are looking for an exceptionally versatile language that allows you to write in different styles more than any other language then have a look at C#"." that would resolve most issues raised here to the current readme. Anyway it is easier to modify that one than the previous one which was too sparse.

Anyway I don't think it matters much. The chances of someone deciding to do a track based on a well (or better) written readme is probably zero. 

 I wrote it more to help the c# track be found by search engines, so that if someone wants to try an online c# course they are more likely to find exercism and join here. 

That should be the real goal of the readme.

ErikSchierboom:
I agree with @martinfreedman.

PR #519, opened 2018-02-04
Inline the track-wide exercise README insert into the exercise template

kytrinyx:
Since we now have a track-wide exercise template, I've inlined the markdown README insert in the docs directory into the template.

The first three commits do general cleanup so that the diff when inlining is minimal.

After inlining, I've got a commit that tweaks the inlined text to make it refer to the actual exercise directory name, since the template has access to that value at generation time.

The final commit regenerates all the READMEs using the updated template.

See https://github.com/exercism/meta/issues/94

### DIFF
--- a/config/exercise_readme.go.tmpl
+++ b/config/exercise_readme.go.tmpl
@@ -4,10 +4,12 @@
 {{- with .Hints }}
 {{ . }}
 {{ end }}
-{{- with .TrackInsert }}
-{{ . }}
-{{ end }}
-{{- with .Spec.Credits -}}
+### Submitting Exercises
+
+Note that, when trying to submit an exercise, make sure the exercise file that you're submitting is in the `exercism/csharp/<exerciseName>` directory.
+
+For example, if you're submitting `bob.cs` for the Bob exercise, the submit command would be something like `exercism submit <path_to_exercism_dir>/csharp/bob/bob.cs`.
+{{ with .Spec.Credits }}
 ## Source
 
 {{ . }}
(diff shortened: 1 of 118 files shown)


ErikSchierboom:
Looking good! We as the C# track's maintainers could have a discussion to determine if the track-specific bit about the submit directory adds any real value. @robkeim @jpreese what do you think?

robkeim:
Thanks @kytrinyx! @ErikSchierboom I'm not sure how much additional value that submit directory is adding and if we felt there was a lot of value add I'd argue we should try to get it integrated for all of the tracks. What are your thoughts?

ErikSchierboom:
I don't think it adds much value. Personally, I'm in favor of just removing it. The Exercism CLI should actually help the user with these kinds of errors I feel.

kytrinyx:
> The Exercism CLI should actually help the user with these kinds of errors I feel.

Agreed! I'll remove it and push this back up.

kytrinyx:
Oh. Didn't realize this was merged. I'll submit a new PR :)

kytrinyx:
Actually, I'll let you all sort it out now that I think of it. Sorry for the noise 🌷

ErikSchierboom:
@kytrinyx One question: is there a tool with which we can regenerate the docs?

kytrinyx:
> is there a tool with which we can regenerate the docs?

Shoot, yes! 100% absolutely yes :)

TL;DR:

```
configlet generate . --spec-path=../path/to/problem-specifications
```

Long version:
https://github.com/exercism/docs/blob/master/maintaining-a-track/regenerating-exercise-readmes.md

ErikSchierboom:
Wonderful, thanks!

Issue #420, opened 2017-09-06
Add Robot Simulator Test Generator


jpreese:
This issue is part of an overall initiative to complete #195.

The documentation on how to add generators can be found [here](https://github.com/exercism/csharp/blob/master/docs/GENERATORS.md). If you get stuck or have any questions about adding this generator, please do not hesitate to reach out!

nikibobi:
I will take this task

GKotfis:
I have a solution but need to resolve one problem with the same description for two test. If you take a look at [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/robot-simulator/canonical-data.json) description _"does not change the position"_ appears two times. So it renders two methods with the same name. Two ideas come to my mind:
1. Update canonical data
2. Do something on generator side to make unique method names by adding some suffix

ErikSchierboom:
@GKotfis There already is such an option in the `CanonicalDataCase` class: `UseFullDescriptionPath`.

ErikSchierboom:
Closed by #521.

PR #521, opened 2018-02-21
Add Robot Simulator Test Generator

GKotfis:
Reference: #420


ErikSchierboom:
This is looking good! There is a build failure though, which is probably due to the rename of the enum not being applied to the example file.

ErikSchierboom:
Thanks a lot!

Issue #414, opened 2017-09-06
Add Meetup Test Generator


jpreese:
This issue is part of an overall initiative to complete #195.

The documentation on how to add generators can be found [here](https://github.com/exercism/csharp/blob/master/docs/GENERATORS.md). If you get stuck or have any questions about adding this generator, please do not hesitate to reach out!

InKahootz:
Started on this one.

jpreese:
Were you still wanting to complete this one @InKahootz ?

PR #522, opened 2018-02-22
Upgrade meetup test generator to the latest canonical data

GKotfis:
Reference issue: #414 

I took this one because of it's still open. But I see that there is already added meetup test generator. So this changes upgrade generator to the latest canonical-data.


ErikSchierboom:
Thanks a lot! 🎉

Issue #535, opened 2018-04-16
Don't list deprecated exercises as unimplemented


ErikSchierboom:
The generator can list unimplemented exercises by using `dotnet run -s Unimplemented`, however the unimplemented list contains deprecated exercises, which should be excluded.

ErikSchierboom:
Closed by #559.

PR #559, opened 2018-04-21
Skip deprecated exercise listing

ShamilS:
Solution for ['Don't list deprecated exercises as unimplemented #535' issue](https://github.com/exercism/csharp/issues/535)

### DIFF
--- a/generators/ExerciseCollection.cs
+++ b/generators/ExerciseCollection.cs
@@ -29,7 +29,6 @@ namespace Generators
 
         private IEnumerable<Exercise> GetExercises()
         {
-            //ss: foreach (var exerciseName in ConfigFile.GetExerciseNames())
             foreach (var exercise in ConfigFile.GetExercises())
             {
                 string exerciseName = exercise.Name;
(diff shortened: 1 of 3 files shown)


generators/Input/ConfigFile.cs
```diff
@@ -2,28 +2,36 @@
...
 
         private class Config
         {
             public ConfigExercise[] Exercises { get; set; }
         }
 
-        private class ConfigExercise
+        public class ConfigExercise
         {
+            public string Name { get { return this.Slug.ToExerciseName(); } }
```
ErikSchierboom: You can simplify this a little bit: `public string Name => Slug.ToExerciseName();`

generators/ExerciseCollection.cs
```diff
@@ -29,9 +29,12 @@ public ExerciseCollection(CanonicalDataFile canonicalDataFile)
 
         private IEnumerable<Exercise> GetExercises()
         {
-            foreach (var exerciseName in ConfigFile.GetExerciseNames())
+            foreach (var exercise in ConfigFile.GetExercises())
             {
-                if (HasNoCanonicalData(exerciseName))
+                string exerciseName = exercise.Name;
```
ErikSchierboom: Could replace `string` with `var`?

ErikSchierboom:
Great work! Thanks a lot 🎉

PR #561, opened 2018-04-22
'Proverb' problem generator and solution v.1

ShamilS:
Done.

### DIFF
--- a/generators/Exercises/Proverb.cs
+++ b/generators/Exercises/Proverb.cs
@@ -1,0 +1,16 @@
+﻿using Generators.Output;
+
+namespace Generators.Exercises
+{
+    public class Proverb : GeneratorExercise
+    {
+        protected override string RenderTestMethodBodyAssert(TestMethodBody testMethodBody)
+        {
+            if  (testMethodBody.CanonicalDataCase.Properties["input"]["strings"] as string[] == null)
+            {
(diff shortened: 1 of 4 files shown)


generators/Exercises/Proverb.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Generators.Output;
+
+namespace Generators.Exercises
+{
+    public class Proverb : GeneratorExercise
+    {
+        protected override string RenderTestMethodBodyAssert(TestMethodBody testMethodBody)
+        {
+            if  (testMethodBody.CanonicalDataCase.Properties["input"]["strings"] as string[] == null)
```
ErikSchierboom: One space too much after the `if` statement.

generators/Exercises/Proverb.cs
```diff
@@ -0,0 +1,16 @@
...
+            {
+                return TemplateRenderer.RenderInline("Assert.Empty(Proverb.Recite(new string[] { }));", new { });
```
ErikSchierboom: This shouldn't be necessary if I recall correctly, but I assume you wouldn't use it otherwise I guess. What happens when you omit these lines?

ShamilS: Compile time error for the first test method.

PR #562, opened 2018-04-23
'sublist' problem test generator v.1

ShamilS:
Done.

### DIFF
--- a/exercises/sublist/SublistTest.cs
+++ b/exercises/sublist/SublistTest.cs
@@ -14,96 +14,96 @@ public class SublistTest
     [Fact(Skip = "Remove to run test")]
     public void Empty_list_within_non_empty_list()
     {
-        Assert.Equal(SublistType.Sublist, Sublist.Classify(new List<int>() { }, new List<int>() {1, 2, 3}));
+        Assert.Equal(SublistType.Sublist, Sublist.Classify(new List<int>() { }, new List<int>() { 1, 2, 3 }));
     }
 
     [Fact(Skip = "Remove to run test")]
     public void Non_empty_list_contains_empty_list()
     {
-        Assert.Equal(SublistType.Superlist, Sublist.Classify(new List<int>() {1, 2, 3}, new List<int>() { }));
+        Assert.Equal(SublistType.Superlist, Sublist.Classify(new List<int>() { 1, 2, 3 }, new List<int>() { }));
     }
 
     [Fact(Skip = "Remove to run test")]
     public void List_equals_itself()
     {
-        Assert.Equal(SublistType.Equal, Sublist.Classify(new List<int>() {1, 2, 3}, new List<int>() {1, 2, 3}));
+        Assert.Equal(SublistType.Equal, Sublist.Classify(new List<int>() { 1, 2, 3 }, new List<int>() { 1, 2, 3 }));
     }
 
     [Fact(Skip = "Remove to run test")]
     public void Different_lists()
     {
-        Assert.Equal(SublistType.Unequal, Sublist.Classify(new List<int>() {1, 2, 3}, new List<int>() {2, 3, 4}));
+        Assert.Equal(SublistType.Unequal, Sublist.Classify(new List<int>() { 1, 2, 3 }, new List<int>() { 2, 3, 4 }));
     }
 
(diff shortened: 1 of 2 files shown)


generators/Exercises/Sublist.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            var template = (list != null) ? string.Join(", ", Array.ConvertAll<int, string>(list, (x) => { return $"{x}"; })) : " ";
+            return new UnescapedValue($"new List<int>() {{{template}}}");
+        }
+        protected override void UpdateCanonicalData(CanonicalData canonicalData)
+        {
+            foreach (var canonicalDataCase in canonicalData.Cases)
+            {              
+                var input = canonicalDataCase.Properties["input"] as System.Collections.Generic.Dictionary<string, object>;
+                input["listOne"] = inputValues (input["listOne"] as int[]);
```
ErikSchierboom: There is an unneeded space after `inputValues`

exercises/sublist/SublistTest.cs
```diff
@@ -1,150 +1,109 @@
...
     }
 
     [Fact(Skip = "Remove to run test")]
-    public void Empty_is_a_sublist_of_anything()
+    public void Empty_list_within_non_empty_list()
     {
-        var list1 = new List<int>();
-        var list2 = new List<int> { 1, 2, 3, 4 };
-        Assert.Equal(SublistType.Sublist, Sublist.Classify(list1, list2));
+        Assert.Equal(SublistType.Sublist, Sublist.Classify(new List<int>() { }, new List<int>() {1, 2, 3}));
```
ErikSchierboom: Could you make it such that there is a space after the opening `{` and before the closing `}`? (also goes for the other tests).

generators/Exercises/Sublist.cs
```diff
@@ -0,0 +1,34 @@
...
+using Humanizer;
+using System.Linq;
+
+namespace Generators.Exercises
+{
+    public class Sublist : GeneratorExercise
+    {
+        protected override HashSet<string> AddAdditionalNamespaces() => new HashSet<string>() { typeof(System.Collections.Generic.IList<int>).Namespace };
+
+        private UnescapedValue inputValues(int[] list)
```
robkeim: By convention C# methods start with upper case letters so this method should be named `InputValues` instead of `inputValues`.

There are also some minor simplifications you can do to make this less verbose resulting in:
```
        private static UnescapedValue InputValues(int[] list)
        {
            var template = list != null ? string.Join(", ", list) : "";
            return new UnescapedValue($"new List<int> {{ {template} }}".Replace("  ", " "));
        }
```

generators/Exercises/Sublist.cs
```diff
@@ -0,0 +1,34 @@
...
+using Generators.Input;
+using Generators.Output;
+using Humanizer;
+using System.Linq;
+
+namespace Generators.Exercises
+{
+    public class Sublist : GeneratorExercise
+    {
+        protected override HashSet<string> AddAdditionalNamespaces() => new HashSet<string>() { typeof(System.Collections.Generic.IList<int>).Namespace };
```
robkeim: It looks like you've already included the `System.Collections.Generic` namespace so `typeof(IList<int>).Namespace` should work here.

generators/Exercises/Sublist.cs
```diff
@@ -0,0 +1,34 @@
...
+        private UnescapedValue inputValues(int[] list)
+        {
+            var template = (list != null) ? string.Join(", ", Array.ConvertAll<int, string>(list, (x) => { return $"{x}"; })) : "";
+            return new UnescapedValue($"new List<int>() {{ {template} }}".Replace("  ", " "));
+        }
+        protected override void UpdateCanonicalData(CanonicalData canonicalData)
+        {
+            foreach (var canonicalDataCase in canonicalData.Cases)
+            {              
+                var input = canonicalDataCase.Properties["input"] as System.Collections.Generic.Dictionary<string, object>;
```
robkeim: Same thing here `Dictionary<string, object>` should be sufficient.

robkeim:
Thanks @ShamilS :tada:

Issue #552, opened 2018-04-16
Add twelve-days generator


ErikSchierboom:
If you want to implement this exercise, please note that you need to ensure the test cases are generated based on the input.

Here's the exercise description:
https://github.com/exercism/problem-specifications/blob/master/exercises/twelve-days/description.md

Here's the test case data:
https://github.com/exercism/problem-specifications/blob/master/exercises/twelve-days/canonical-data.json

ShamilS:
I will make it.

ErikSchierboom:
Closed by #564.

PR #564, opened 2018-04-29
twelve-days generator and solution v.1

ShamilS:
Done #552 - "Add twelve-days generator"

### DIFF
--- a/generators/Exercises/TwelveDays.cs
+++ b/generators/Exercises/TwelveDays.cs
@@ -22,11 +22,6 @@ namespace Generators.Exercises
                 {
                     properties["input"] = input["startVerse"];
                 }
-                else if ((long)input["startVerse"] == 1 && (long)input["endVerse"] == 12)
-                {
-                    canonicalDataCase.Property = "sing";
-                    properties["input"] = null;
-                }
             }
         }
     }
--- a/generators/Exercises/TwelveDays.cs
+++ b/generators/Exercises/TwelveDays.cs
@@ -1,0 +1,50 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Generators.Input;
+using Generators.Output;
+using System.Linq;
+using Humanizer;
+
+namespace Generators.Exercises
+{
+    public class TwelveDays : GeneratorExercise
+    {
+        protected override void UpdateCanonicalData(CanonicalData canonicalData)
+        {
+            canonicalData.Exercise = "twelve-days-song";
+            foreach (var canonicalDataCase in canonicalData.Cases)
+            {
+                canonicalDataCase.UseVariableForExpected = true;
+                var properties = canonicalDataCase.Properties as Dictionary<string, object>;
+                var input = properties["input"] as Dictionary<string, object>;
+                if ((long)input["startVerse"] == (long)input["endVerse"])
+                {
+                    canonicalDataCase.Property = "verse";
+                    properties["input"] = input["startVerse"];
+                }
+                else canonicalDataCase.Property = "verses";
+            }
+        }
+
+        protected override string RenderTestMethodBodyArrange(TestMethodBody testMethodBody)
+        {
+            string expectedValue = "";
+            var properties = testMethodBody.CanonicalDataCase.Properties as Dictionary<string, object>;
+            var expectedValues = properties["expected"] as string[];
+            if (expectedValues.Length == 1)
+            {
+                expectedValue = $"\"{expectedValues[0]}\\n\"";
+            }
+            else
+            {
+                for (int i = 0; i < expectedValues.Length; i++)
+                {
+                    expectedValues[i] = $"\"{expectedValues[i]}\\n\\n\"" + (i < expectedValues.Length - 1? " +" : "");
+                }
+                expectedValue = expectedValues.Aggregate("", (x, y) => x + System.Environment.NewLine + new string(' ', 4) + y);
+            }
+            return $"var expected = {expectedValue};";
+        }
+    }
+}
--- a/generators/Exercises/TwelveDays.cs
+++ b/generators/Exercises/TwelveDays.cs
@@ -12,39 +12,22 @@ namespace Generators.Exercises
     {
         protected override void UpdateCanonicalData(CanonicalData canonicalData)
         {
-            canonicalData.Exercise = "twelve-days-song";
             foreach (var canonicalDataCase in canonicalData.Cases)
             {
                 canonicalDataCase.UseVariableForExpected = true;
+                canonicalDataCase.Expected = ConvertHelper.ToMultiLineString(canonicalDataCase.Expected);
                 var properties = canonicalDataCase.Properties as Dictionary<string, object>;
                 var input = properties["input"] as Dictionary<string, object>;
                 if ((long)input["startVerse"] == (long)input["endVerse"])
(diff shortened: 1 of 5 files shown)


generators/Exercises/TwelveDays.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Linq;
+using Humanizer;
+
+namespace Generators.Exercises
+{
+    public class TwelveDays : GeneratorExercise
+    {
+        protected override void UpdateCanonicalData(CanonicalData canonicalData)
+        {
+            canonicalData.Exercise = "twelve-days-song";
```
ErikSchierboom: Why the changed exercise name?

generators/Exercises/TwelveDays.cs
```diff
@@ -0,0 +1,50 @@
...
+            foreach (var canonicalDataCase in canonicalData.Cases)
+            {
+                canonicalDataCase.UseVariableForExpected = true;
+                var properties = canonicalDataCase.Properties as Dictionary<string, object>;
+                var input = properties["input"] as Dictionary<string, object>;
+                if ((long)input["startVerse"] == (long)input["endVerse"])
```
ErikSchierboom: I understand you did this to keep the current implementation the same, but I think we should adhere to the canonical data. That means having just one property (`recite`) with start and end verse parameters. If we adhere to the canonical data, this customization can all go away.

generators/Exercises/TwelveDays.cs
```diff
@@ -0,0 +1,50 @@
...
+                {
+                    canonicalDataCase.Property = "verse";
+                    properties["input"] = input["startVerse"];
+                }
+                else canonicalDataCase.Property = "verses";
+            }
+        }
+
+        protected override string RenderTestMethodBodyArrange(TestMethodBody testMethodBody)
```
ErikSchierboom: This is not the correct method to override, as this is really a core method and you should rarely have to change its behavior. What you are looking for (formatting the multiline string) should be easy to do, see [this generator](https://github.com/exercism/csharp/blob/master/generators/Exercises/FoodChain.cs#L11).

ShamilS: @ErikSchierboom I will make the changes as you requested - my previous comments were just to clarify my first version code's background.

generators/Exercises/TwelveDays.cs
ErikSchierboom: I understand, but it is in fact the other way around: we should be following the canonical data whenever possible.

generators/Exercises/TwelveDays.cs
ErikSchierboom: I understand, but it is in fact the other way around: we should be following the canonical data whenever possible.

generators/Exercises/TwelveDays.cs
ErikSchierboom: Well, it *is* still a valid extension point, albeit one that should seldom be used. Thanks for the explanation.

generators/Exercises/TwelveDays.cs
ShamilS: @ErikSchierboom OK, should I then remove

    public static string Sing()
    ...

from Example.cs and TwelveDays.cs ?

generators/Exercises/TwelveDays.cs
ShamilS: @ErikSchierboom OK, [this generator](https://github.com/exercism/csharp/blob/master/generators/Exercises/FoodChain.cs#L11]) doesn't add trailing new line - \n - to the end of verses' strings. Should I then remove the trailing new line - \n - from the Example.cs cases?:

            case 1:
              return "On the first day of Christmas my true love gave to me, a Partridge in a Pear Tree.\n";
          ...

If I will remove the trailing new line - \n - from the end of verses' strings then [twelve-days problems description](https://github.com/exercism/problem-specifications/blob/master/exercises/twelve-days/description.md) should be also corrected AFAIU.

OK. I will make all these related corrections.

generators/Exercises/TwelveDays.cs
ErikSchierboom: Correct, that method should be remove in both files.

generators/Exercises/TwelveDays.cs
ErikSchierboom: That is actually a good point! I think you may have found an issue with the problem description, which implies that there should be an empty lines between each line. The canonical data however states otherwise, and in this case I would like to follow the canonical data (so no blank lines between lines, and no trailing newline for a single line.

generators/Exercises/TwelveDays.cs
```diff
@@ -0,0 +1,28 @@
...
+        {
+            foreach (var canonicalDataCase in canonicalData.Cases)
+            {
+                canonicalDataCase.UseVariableForExpected = true;
+                canonicalDataCase.Expected = ConvertHelper.ToMultiLineString(canonicalDataCase.Expected);
+                var properties = canonicalDataCase.Properties as Dictionary<string, object>;
+                var input = properties["input"] as Dictionary<string, object>;
+                if ((long)input["startVerse"] == (long)input["endVerse"])
+                {
+                    properties["input"] = input["startVerse"];
```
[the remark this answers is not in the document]
ShamilS: @ErikSchierboom Thanks. OK, I will wait for @robkeim answer on your question. As for me I'd leave overloaded recite methods.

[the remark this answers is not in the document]
robkeim: From my point of view, in order to call for one verse this looks a little strange as it seems you're passing redundant information:
`TwelveDays.Recite(1, 1);`

How do you guys feel about a single method with the following signature:
`public static string Recite(int start, int? end = null)`

Then you would call it as follows:
- One verse => `TwelveDays.Recite(1);`
- Two verses => `TwelveDays.Recite(1, 2);`

This adds a small additional complexity for the student but exposes them to nullable values and default parameters.

@ErikSchierboom and @ShamilS how do you feel about that approach?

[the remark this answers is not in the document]
ErikSchierboom: I like it!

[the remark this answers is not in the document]
ShamilS: @robkeim and @ErikSchierboom I'd note that currently the second parameter of the Recite(...) method means the ending verse number not the quantity of verses. I'd also note that introducing two concepts, optional parameters and nullable types, could be considered by some folks as methodically incorrect. Therefore if you wanted to have just one Recite(...) method without overloads then I'd propose to have it as the following:

     public static string Recite(int startingVerseNumber, int versesCount = 1)

IOW, introducing concept of optional parameters here in this twelve-days exercise, and leaving introduction to the nullable types to some other exercise(s).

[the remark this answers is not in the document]
ErikSchierboom: I agree with your assessment @ShamilS. Let's go with the two overloads then.

ErikSchierboom:
Looking good! Waiting for the OK from @robkeim.

generators/Exercises/TwelveDays.cs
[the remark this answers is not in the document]
robkeim: Sounds good to me!

robkeim:
Thanks @ShamilS! 🎉

PR #565, opened 2018-05-01
matrix generator #542

ShamilS:


ShamilS:
**Note to my current PR**

Existing matrix.cs class interface to implement

    - constructor Matrix(string input)
    - public int Rows
    - public int Cols
    - public int[] Row(int row) 
    - public int[] Column(int col)

doesn't correspond to the canonical data json. I have made the first version of the matrix generator to generate tests corresponding to the above (obsolete?) interface. 

    public class Matrix : GeneratorExercise
    {
        protected override string RenderTestMethodBodyAssert(TestMethodBody testMethodBody)
        {
            var canonicalDataCase = testMethodBody.CanonicalDataCase;
            var input = canonicalDataCase.Properties["input"] as System.Collections.Generic.Dictionary<string, object>;
            var template = "Assert.Equal(new[] { {{ExpectedData}} }, new Matrix(\"{{ConstructorData}}\").{{MethodName}}({{MethodCallArgumentValue}}));";
            var templateParams = new
            {
                ExpectedData = string.Join(", ", canonicalDataCase.Properties["expected"] as int[]),
                ConstructorData = input["string"].ToString().EscapeSpecialCharacters(),
                MethodName = canonicalDataCase.Property.Dehumanize(),
                MethodCallArgumentValue = input["index"]
            };
            return TemplateRenderer.RenderInline(template, templateParams);
        }
    }

If I will change matrix.cs interface to

    - static int[] Row(string input, int row) 
    - static int[] Column(string input, int col)

then the matrix generator would just become

    public class Matrix : GeneratorExercise
    {
    }

I suppose I should to the latter and to submit the new PR?

P.S. (Rethoric) question/remark: How is comes that there sometimes exists such a big discrepancy between json canonical data and an existing interface/implementation of the corresponding exercise? In the case of the existing matrix.cs implementation almost all its interface should be ruined to make the new implementation corresponding to the canonical json data - I'm (always) feeling some resistance to crush and to rebuild (from scratch) the existing implementations interfaces - I thought I have buried my NIH-syndrome forever but here it has to be digged-up and to be applied fiercely  ;)

P.P.S. OOP with static methods only as in the case of the matrix.cs canoninal data looks like a heavy degradation/migration to the functional programming style, isn't it? What are the reasoning of such approach of exercism.io?

robkeim:
@ShamilS regarding this:

> P.S. (Rethoric) question/remark: How is comes that there sometimes exists such a big discrepancy between json canonical data and an existing interface/implementation of the corresponding exercise? In the case of the existing matrix.cs implementation almost all its interface should be ruined to make the new implementation corresponding to the canonical json data - I'm (always) feeling some resistance to crush and to rebuild (from scratch) the existing implementations interfaces - I thought I have buried my NIH-syndrome forever but here it has to be digged-up and to be applied fiercely ;)

The reason that the existing implementation might differ largely from the canonical data is that historically we didn't have canonical data so each track implemented their own version of the exercise with corresponding tests. Recently there's been a push to get more consistency across the different language tracks, hence the canonical data. Now we need to retrofit our existing implementations to the canonical data. In some (most?) of the cases, the needed changes aren't that big, but in some cases more dramatic changes are required

ShamilS:
@ErikSchierboom and @robkeim I have fixed the generator code. The resulting code is simple

    public class Matrix : GeneratorExercise
    {
        protected override void UpdateCanonicalData(CanonicalData canonicalData)
        {
            foreach (var canonicalDataCase in canonicalData.Cases)
            {
                canonicalDataCase.Properties["string"] = canonicalDataCase.Properties["input"]["string"];
                canonicalDataCase.SetConstructorInputParameters("string");
                canonicalDataCase.Properties["index"] = canonicalDataCase.Properties["input"]["index"];
                canonicalDataCase.SetInputParameters("index");
            }
        }
    }

but this final code simplicity costed  quite some time to get to because of the bug(?)/feature(?) in the [CanonicalDataCase.cs](https://github.com/exercism/csharp/blob/master/generators/Input/CanonicalDataCase.cs) and/or in some other generators - see #566 and #567 .

The bug/feature is appearing in some other generators when one the following two  [CanonicalDataCase.cs](https://github.com/exercism/csharp/blob/master/generators/Input/CanonicalDataCase.cs) code lines are called:

        [JsonIgnore]
        public IReadOnlyDictionary<string, dynamic> Input
            => _inputParameters.ToDictionary(parameter => parameter, parameter => Properties[parameter]);

        [JsonIgnore]
        public IReadOnlyDictionary<string, dynamic> ConstructorInput 
            => _constructorInputParameters.ToDictionary(parameter => parameter, parameter => Properties[parameter]);

PR #593, opened 2018-06-10
pov: Add generator

### LINKED ISSUE #544: Add pov generator
If you want to implement this exercise, please note that you need to ensure the test cases are generated based on the input.

Here's the exercise description:
https://github.com/exercism/problem-specifications/blob/master/exercises/pov/description.md

Here's the test case data:
https://github.com/exercism/problem-specifications/blob/master/exercises/pov/canonical-data.json

ErikSchierboom:
Closes #544.

This PR also removes the generic-ness of the exercise, as I think the subject matter at hand is already complex enough without the generics.

### DIFF
--- a/exercises/pov/Example.cs
+++ b/exercises/pov/Example.cs
@@ -2,75 +2,72 @@
 using System.Collections.Generic;
 using System.Linq;
 
-public class Graph<T> : IEquatable<Graph<T>>
+public class Tree : IEquatable<Tree>
 {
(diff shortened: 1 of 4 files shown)


robkeim:
Thanks @ErikSchierboom! Agreed with the removing of generics... we could always re-add that complexity later if we wanted/needed, or even consider having other exercises that focus on the illustrating the power of generics.

PR #601, opened 2018-06-25
Generators refactoring

ErikSchierboom:
This is to test the newly enabled [WIP plugin](https://github.com/apps/wip) (see https://github.com/exercism/discussions/issues/217#issuecomment-399756092). And it will also give insight into how my refactoring attempt is progressing.


ErikSchierboom:
Finally, after much work, I'm done with my huge generators refactoring. It has become even bigger that the F# equivalent, and it contains the following improvements:

- Simplified extension points available to generators to three virtual methods.
- Removed as many of the classes the user should know about.
- Greatly simplified rendering code.
- Make data rendering code generic (handle any dictionary/array/list instead of only a couple of hardcoded ones).
- Many places in which the generator code has been simplified.
- Fixed a bug where the test methods were not always in the correct order when using nested canonical data cases.
- Made canonical data and canonical data case classes immutable.
- Allow testing of properties and constructors.
- Use `ValueTuple<T>` where possible.
- Add `MultilineString` class to allow generators to be explicit on how they want the data to be output.
- Add helper method to render enumerations.
- Remove unused code.
- Updated documentation with more examples of common use cases.
- Restructured namespaces to make the code have a better structure.
- And just overall many, many improvements to make the code more readable.

CC @robkeim @jpreese

ErikSchierboom:
@robkeim @jpreese If this PR is too big too review, no sweat. If you do want to review, it might be best to just check out this branch and examine the code locally.

robkeim:
@ErikSchierboom I'm feeling brave... I'll take a look :)

docs/GENERATORS.md
```diff
@@ -32,71 +34,215 @@ namespace Generators.Exercises
...
-There may exist cases where a suite of unit tests will need to reuse the same logic in each of the tests. Rather than duplicating code, this method allows you to provide helper methods for the tests.
+Note that `ArgumentException` type's namespace will be automatically added to the list of namespaces used in the test class.
 
-An example of this is the [Tournament](https://github.com/exercism/csharp/blob/master/generators/Exercises/Tournament.cs#L45) generator.
+#### Custom input/constructor parameters
 
-Additional methods added using this override will be added to the bottom of the test suite.
+In some case, you might want to override the parameters that are used as input parameters.
+
+An example of this is the [two fer]() generator, which does not use any input parameters when the `"name"` input parameter is set to `null`:
```
robkeim: Missing url here

docs/GENERATORS.md
```diff
@@ -32,71 +34,215 @@ namespace Generators.Exercises
...
+  // [...]
+
+  if (testMethod.Input["name"] is null)
+    testMethod.InputParameters = Array.Empty<string>();
+}
+```
+
+If a test method tests an instance method, you can also specify which parameters to use as constructor parameters (the others will be input parameters, unless specified otherwise).
+
+An example of this is the [matrix]() generator, which specifies that the `"string"` parameter should be passed as a constructor parameter:
```
robkeim: Here as well

docs/GENERATORS.md
```diff
@@ -32,71 +34,215 @@ namespace Generators.Exercises
...
+```
 
-More advanced tests may need to leverage a `template`. A template allows you to add additional code to a test and assert more complex statements.
+#### Test for an exception being thrown
 
-An example of this is the [RunLengthEncoding](https://github.com/exercism/csharp/blob/master/generators/Exercises/RunLengthEncoding.cs) test.
+Some test methods want to verify that an exception is being thrown.
 
-Here the **Assert** is being overridden. The assert needs to call additional functions, but only if the property is `consistency`. Otherwise, render the assert as usual.
+An example of this is the [RNA transcription]() generator, which defines that some test methods should throws an `ArgumentException`:
```
robkeim: Missing url here

ErikSchierboom:
@robkeim Great! I'm really interested in what you think.

docs/GENERATORS.md
```diff
@@ -32,71 +34,215 @@ namespace Generators.Exercises
...
+}
+```
+
+Note that as mentioned before, the namespace of any thrown exception types are automaticall added to the list of namespaces.
+
+### Method 3: UpdateTestClass(TestClass testClass)
+
+This method allows you to customize the output of the test class. Only in rare cases would you want to override this method. The most common use case to override this method, is to add additional (helper) methods to the test suite.
+
+An example of this is the [Tournament](https://github.com/exercism/csharp/blob/master/generators/Exercises/Tournament.cs) generator, which adds a helper method to the test suite:
```
robkeim: Minor: tournament should be lowercase to be consistent with the other test names

exercises/all-your-base/AllYourBaseTest.cs
```diff
@@ -89,7 +89,7 @@ public void Number_15_bit_integer()
     public void Empty_list()
     {
         var inputBase = 2;
-        var digits = new int[0];
+        var digits = Array.Empty<int>();
```
robkeim: 👍

exercises/bowling/BowlingTest.cs
```diff
@@ -60,7 +60,30 @@ public void Consecutive_spares_each_get_a_one_roll_bonus()
     public void A_spare_in_the_last_frame_gets_a_one_roll_bonus_that_is_counted_once()
     {
         var sut = new BowlingGame();
-        var previousRolls = new [] { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 7, 3, 7 };
+        var previousRolls = new[]
```
robkeim: Why are these all on separate lines but the previous tests aren't? IMO it's probably better to have it on one line

robkeim: I guess there are a few other occurrences in the rest of this test file.

exercises/go-counting/Example.cs
```diff
@@ -37,45 +37,45 @@ private static Owner[][] ParseBoard(string input)
...
         var col = coordinate.Item1;
 
         var coords = new[]
         {
-            new ValueTuple<int, int>(col,   row - 1),
-            new ValueTuple<int, int>(col-1, row),
-            new ValueTuple<int, int>(col+1, row),
-            new ValueTuple<int, int>(col,   row+1)
+            (col,   row - 1),
+            (col-1, row),
```
robkeim: Super minor: missing spaces around the `-` here and the `+` below

exercises/minesweeper/MinesweeperTest.cs
```diff
@@ -1,235 +1,160 @@
...
     public void No_columns()
     {
-        var minefield = new string[] 
-        { 
-            "" 
-        };
-        var expected = new string[] 
-        { 
-            "" 
+        var minefield = new[]
```
robkeim: Minor: the fact that each row is a separate line here vs in the expected it's on one line makes this definitely look like it wasn't handwritten.

exercises/sieve/SieveTest.cs
```diff
@@ -35,7 +35,177 @@ public void Limit_is_prime()
     [Fact(Skip = "Remove to run test")]
     public void Find_primes_up_to_1000()
     {
-        var expected = new[] { 2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47, 53, 59, 61, 67, 71, 73, 79, 83, 89, 97, 101, 103, 107, 109, 113, 127, 131, 137, 139, 149, 151, 157, 163, 167, 173, 179, 181, 191, 193, 197, 199, 211, 223, 227, 229, 233, 239, 241, 251, 257, 263, 269, 271, 277, 281, 283, 293, 307, 311, 313, 317, 331, 337, 347, 349, 353, 359, 367, 373, 379, 383, 389, 397, 401, 409, 419, 421, 431, 433, 439, 443, 449, 457, 461, 463, 467, 479, 487, 491, 499, 503, 509, 521, 523, 541, 547, 557, 563, 569, 571, 577, 587, 593, 599, 601, 607, 613, 617, 619, 631, 641, 643, 647, 653, 659, 661, 673, 677, 683, 691, 701, 709, 719, 727, 733, 739, 743, 751, 757, 761, 769, 773, 787, 797, 809, 811, 821, 823, 827, 829, 839, 853, 857, 859, 863, 877, 881, 883, 887, 907, 911, 919, 929, 937, 941, 947, 953, 967, 971, 977, 983, 991, 997 };
+        var expected = new[]
```
robkeim: This could probably be on one line or a few lines instead of one line per number

jpreese:
haha, oh man.. where to start. Yeah, let me check this out locally and try to make sense of it all. Should hopefully have some time later in the week.

exercises/minesweeper/MinesweeperTest.cs
ErikSchierboom: Fixed

exercises/bowling/BowlingTest.cs
ErikSchierboom: Well, this is me trying to be clever. Many of the input arrays have a very limited number of elements (e.g. 1 or 2), in which case rendering them over multiple lines would be overkill IMHO. The code that does this check can be found [here](https://github.com/exercism/csharp/pull/601/files#diff-186289da388a45f86733f30d74d153dcR75). It essentially checks if it is not an array of arrays and if the rendered single line output would be less than or equal to 72 characters. If so, it is output as a single line. 

This is of course completely up for debate. I'd be happy to hear your thoughts on this.

exercises/sieve/SieveTest.cs
ErikSchierboom: See comment above.

exercises/bowling/BowlingTest.cs
ErikSchierboom: What I could do, is detect how many lines would end up being created. If there are more than a given number, say 8, then I could put multiple elements on a single line. Would that be an option?

ErikSchierboom:
Although I do think the commits are really useful to see the gradual evolution of the refactoring, I think that squashing this will be the way to go when merging this PR.

exercises/bowling/BowlingTest.cs
ErikSchierboom: @robkeim I've given it some more thought, and I think I was attempting to be a bit too clever. I've removed the automatic multi-line rendering of single-dimensional arrays. Exercises can choose to opt-in themselves, should they want to do multi-line array initialization.

jpreese:
That review destroyed by browser.

ErikSchierboom:
@jpreese Thanks for the review. I've updated the docs.

ErikSchierboom:
Thanks a ton for reviewing @robkeim and @jpreese!

PR #606, opened 2018-07-16
allergies: Use enumeration for allergens

### LINKED ISSUE #605: allergies: Use enum for allergens
The current `allergies` exercise represents allergens as strings. As they are a fixed set of values, we should be using an enum instead.

ErikSchierboom:
Closed #605

### DIFF
--- a/generators/Exercises/Generators/Allergies.cs
+++ b/generators/Exercises/Generators/Allergies.cs
@@ -1,5 +1,8 @@
-﻿using System.Text;
+﻿using System;
+using System.Linq;
+using System.Text;
 using Exercism.CSharp.Output;
+using Exercism.CSharp.Output.Rendering;
 
 namespace Exercism.CSharp.Exercises.Generators
 {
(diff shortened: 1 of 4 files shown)


generators/Exercises/Generators/Allergies.cs
```diff
@@ -11,19 +14,34 @@ protected override void UpdateTestMethod(TestMethod testMethod)
...
+
+        private UnescapedValue[] ConvertExpected(dynamic expected)
+        {
+            if (expected is string[] allergens)
+                return allergens.Select(RenderAllergen).ToArray();
+
+            return Array.Empty<UnescapedValue>();
+        }
+
+        private UnescapedValue RenderAllergen(dynamic input) => Render.Enum("Allergen", input);
```
robkeim: Minor: does `nameof(Allergen)` work here?

robkeim:
Thanks @ErikSchierboom 👍

generators/Exercises/Generators/Allergies.cs
ErikSchierboom: Nope, because the `Allergen` enumeration is not defined in the generators project :)

ErikSchierboom:
Thanks for merging!

generators/Exercises/Generators/Allergies.cs
robkeim: *face palm* yup makes sense :)

PR #612, opened 2018-07-30
[Leap] Make message to ignore built-in library method C# specific

### LINKED ISSUE #610: Consider calling out DateTime.IsLeapYear() for Leap exercise
Was doing some mentoring and noticed our Leap instructions just say 

>If your language provides a method in the standard library that does this look-up, pretend it doesn't exist and implement it yourself.

I think since we have the ability to be specific, since the README exists within the C# repo, we can explicitly call out DateTime.IsLeapYear().  I think it'd be a good learning opportunity, highlighting this function exists in the first place. As well as makes it a bit more personable to the C# track.

robkeim:
Closes #610

@jpreese FYI

### DIFF
--- a/exercises/leap/.meta/hints.md
+++ b/exercises/leap/.meta/hints.md
@@ -1,4 +1,4 @@
-## Hints
+## Notes
 
 The DateTime class in C# provides a built-in [IsLeapYear](https://msdn.microsoft.com/en-us/library/system.datetime.isleapyear(v=vs.110).aspx) method
 which you should pretend doesn't exist for the purposes of implementing this exercise.
--- a/exercises/leap/.meta/hints.md
+++ b/exercises/leap/.meta/hints.md
@@ -1,0 +1,4 @@
+## Hints
+
+The DateTime class in C# provides a built-in [IsLeapYear](https://msdn.microsoft.com/en-us/library/system.datetime.isleapyear(v=vs.110).aspx) method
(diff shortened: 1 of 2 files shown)


ErikSchierboom:
Just waiting for the CI to become green!

ErikSchierboom:
I don't know if there are other ways to add to the README. Maybe @kytrinyx knows?

And so far, we haven't yet been regenerating the README's, but we should be able to do that using the `configlet` command.

ErikSchierboom:
@robkeim According to [the source](https://github.com/exercism/configlet/blob/master/track/exercise_readme.go), it looks like we should still be using the `hints.md` file (which name is a bit unfortunate in this case). Could you add the new text to the `hints.md` file, and then regenerate the README's using configlet?

robkeim:
@ErikSchierboom yeah I'll go ahead and make that change

exercises/leap/.meta/hints.md
```diff
@@ -0,0 +1,4 @@
+## Hints
```
ErikSchierboom: I was thinking that perhaps this should be something else in this case, maybe "Notes"  or something like that?

robkeim: Makes sense to me... fixed!

ErikSchierboom:
Waiting for the CI.

ErikSchierboom:
Thanks!

PR #616, opened 2018-08-07
Update to Idomatic Nomenclature

ilivewithian:
C# is an object oriented language, therefore 'method' should be used in place of 'function'.

It's nitpicking, but for the 101 stuff it seems sensible to nudge people in the right direction.

### DIFF
--- a/exercises/alphametics/README.md
+++ b/exercises/alphametics/README.md
@@ -1,6 +1,6 @@
 # Alphametics
 
-Write a function to solve alphametics puzzles.
+Write a method to solve alphametics puzzles.
 
 [Alphametics](https://en.wikipedia.org/wiki/Alphametics) is a puzzle where
 letters in words are replaced with numbers.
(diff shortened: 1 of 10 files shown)


ErikSchierboom:
Hi. Thanks for the PR! The exercise README's are automatically generated based on the README defined in the [problem-specifications](https://github.com/exercism/problem-specifications) repository. I do agree that using "method" is preferrable to "function", as it is indeed more idiomatic. The downside to doing customizations is that regenerating the README's becoming a bit more problematic, as we have to re-do the customizations.

What do you feel about this @robkeim and @jpreese?

Regardless whether this is merged, thanks a lot for going through the trouble to create a PR!

robkeim:
@ErikSchierboom do you know how many times we've updated these README's in the past to align with "new" versions? This is the second time we've recently hit an issue where we wanted to modify the README (#612).

If we want to diverge, maybe we could keep track of the changes we've made somewhere so if we did update the README we know which changes that we'd have to re-apply. I'm all in favor of sharing where sharing makes sense but in this case it feels like we're sacrificing the quality of the README for the sole purpose of being able to share it between the tracks.

ErikSchierboom:
Having given this some thought, I think we should be able to solve this using git patch files. Does anyone have any experience with git patch files?

ErikSchierboom:
@ilivewithian I've just opened PR #618 to allow us to support patching README files. As an example, I've also included the changes in this PR. Could you take a look at PR #618 to see how it looks like? Once you feel you understand what is needed (creating a .patch file in the config/patches directory), I could remove the patch from the other PR, merge it and you could then update this PR using the specified patch method. This would allow you to get this PR merged. Would that work for you?

ErikSchierboom:
Closed by #618. Thanks for the PR and bringing this to our attention @ilivewithian!

PR #622, opened 2018-08-29
Patch to fix CustomSet exercise

ChrisPritchard:
The CustomSet default class has a method named Insert, but the tests seem to expect this to be Add. I renamed it to Add because normally Insert would have an index as well - that is, I aligned the base class with the tests rather than the tests with the class.

Sorry to raise a PR for such a trivial issue - if this is against guidelines for contribution please let me know.


ChrisPritchard:
Sure, I've made the change, cheers

ErikSchierboom:
Thanks a lot!

PR #623, opened 2018-09-10
Removed untested constructor on Kindergarten-Garden exercise

ChrisPritchard:
There is a second constructor on the start class, which some students implement, some ignore, and some ask questions about as they are confused. As this constructor is untested, I've removed it.

Sorry about the random merge commits - unsure how to squash those on github


ErikSchierboom:
Thanks for fixing this!

PR #629, opened 2018-09-28
[Accumulate] Remove out of scope test

### LINKED ISSUE #628: Add laziness instructions to accumulate README
Accumulate is not per se about laziness, and this is reflected in the README. The tests suite contains a test that verifies that the implementation is evaluated lazily. We should add a section to the README to mention the laziness property that is being tested.

Note that to update the README, one should create a patch file [here](https://github.com/exercism/csharp/tree/master/config/patches) and then run the `update-docs.sh` command in the root.

robkeim:
Fixes #628 

This one doesn't have a generator so the fix is straightforward.

### DIFF
--- a/exercises/accumulate/AccumulateTest.cs
+++ b/exercises/accumulate/AccumulateTest.cs
@@ -45,17 +45,6 @@ public class AccumulateTest
     }
 
     [Fact(Skip = "Remove to run test")]
-    public void Accumulate_is_lazy()
-    {
-        var counter = 0;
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
Oh, this is awkward. There has been some more discussion between mentors and the concensus now seems to be that the laziness property is useful to teach to people. I've updated the issue to reflect this change, as it now requires a documentation change. Sorry about that!

robkeim:
No worries I'll go ahead and abandon this PR then.

ErikSchierboom:
Once again sorry about this!

PR #659, opened 2018-10-02
Add laziness instructions to accumulate readme

### LINKED ISSUE #628: Add laziness instructions to accumulate README

B1tF8er:
Fix #628


exercises/accumulate/README.md
```diff
@@ -33,6 +33,10 @@ This exercise requires you to write an extension method. For more information, s
 
 To run the tests, run the command `dotnet test` from within the exercise directory.
 
+### Laziness test
+
+Since `accumulate` returns an IEnumerable it's execution can be deferred until `ToList()` it is called on it and this is tested with the `Accumulate_is_lazy` method
```
ErikSchierboom: Some tiny nits:
- Maybe enclose `IEnumerable` in backticks? (to have it render as code)
- I think I would prefer a comma after `IEnumerable`.
- `can be deferred` -> `is deferred`
- `called on it and this is` -> `called on it, which is`

B1tF8er: sure, let me fix those

ErikSchierboom: Great!

config/patches/0004-Add-laziness-instructions-accumulate-readme.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/exercises/accumulate/README.md b/exercises/accumulate/README.md
```
ErikSchierboom: Could you perhaps merge this patch file with the previous one, so that there is only one patch file? That's the last :)

B1tF8er: yeah, just let me do some research on how to do that, and I will do it

ErikSchierboom:
Merged! Thanks for doing this!

B1tF8er:
> Merged! Thanks for doing this!

You're welcome. Glad I could help.

PR #660, opened 2018-10-02
fixes #656 updates clock tests

### LINKED ISSUE #656: Update test suite for exercise: clock
The `clock` exercise's [test suite](https://github.com/exercism/csharp/blob/master/exercises/clock/ClockTest.cs) is automatically generated using its [test generator](https://github.com/exercism/csharp/blob/master/generators/Exercises/Generators/Clock.cs). This test generator converts the `clock`'s [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/clock/canonical-data.json) into the test suite just mentioned.

The canonical data for the `clock` exercise has been updated since the test suite was last generated. We should be updating the test suite to the latest version. To do this, one should:
- Run the test generator for the exercise. This is done by opening a command prompt and running: `dotnet run -e clock`.
- Finally, please make sure that the [example implementation](https://github.com/exercism/csharp/blob/master/exercises/clock/Example.cs) passes all the tests. You can verify this by opening a command prompt in the root directory and running `./build.sh --exercise=clock` or `.\build.ps1 -Exercise clock`, depending on your operating system.

bmeverett:


ErikSchierboom:
Lovely! Thanks for doing this 🎉

PR #663, opened 2018-10-03
Bob: generated new test data

### LINKED ISSUE #643: Update test suite for exercise: bob
The `bob` exercise's [test suite](https://github.com/exercism/csharp/blob/master/exercises/bob/BobTest.cs) is automatically generated using its [test generator](https://github.com/exercism/csharp/blob/master/generators/Exercises/Generators/Bob.cs). This test generator converts the `bob`'s [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/bob/canonical-data.json) into the test suite just mentioned.

The canonical data for the `bob` exercise has been updated since the test suite was last generated. We should be updating the test suite to the latest version. To do this, one should:
- Run the test generator for the exercise. This is done by opening a command prompt in the `generators` directory and running: `dotnet run -e bob`.
- Finally, please make sure that the [example implementation](https://github.com/exercism/csharp/blob/master/exercises/bob/Example.cs)passes all the tests. You can verify this by opening a command prompt in the root directory and running `./build.sh --exercise=bob` or`.\build.ps1 -Exercise bob`, depending on your operating system.

tomschluter:
Closes #643

### DIFF
--- a/exercises/bob/BobTest.cs
+++ b/exercises/bob/BobTest.cs
@@ -1,4 +1,4 @@
-// This file was auto-generated based on version 1.2.0 of the canonical data.
+// This file was auto-generated based on version 1.4.0 of the canonical data.
 
 using Xunit;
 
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
Nice. Thanks! 🎉

PR #665, opened 2018-10-03
Generated test suite.

abecerramatias:
Result of run: dotnet run -e binary-search
No errors:
[12:00:21 INF] Cloning repository...
[12:00:25 INF] Repository cloned.
[12:00:25 INF] Updating repository to latest version...
[12:00:26 INF] Updated repository to latest version.
[12:00:26 INF] Re-generating test classes...
[12:00:27 INF] binary-search: tests generated
[12:00:27 INF] Re-generated test classes.

### DIFF
--- a/exercises/binary-search/BinarySearchTest.cs
+++ b/exercises/binary-search/BinarySearchTest.cs
@@ -58,7 +58,7 @@ public class BinarySearchTest
     {
         var array = new[] { 1, 3, 4, 6, 8, 9, 11 };
         var sut = new BinarySearch(array);
-        Assert.Equal(new Dictionary<string, object> { ["error"] = "value not in array" }, sut.Find(7));
+        Assert.Equal(-1, sut.Find(7));
     }
 
     [Fact(Skip = "Remove to run test")]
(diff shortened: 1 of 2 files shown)


exercises/binary-search/BinarySearchTest.cs
```diff
@@ -58,30 +58,30 @@ public void Identifies_that_a_value_is_not_included_in_the_array()
     {
         var array = new[] { 1, 3, 4, 6, 8, 9, 11 };
         var sut = new BinarySearch(array);
-        Assert.Equal(-1, sut.Find(7));
+        Assert.Equal(new Dictionary<string, object> { ["error"] = "value not in array" }, sut.Find(7));
```
ErikSchierboom: This code will not compile. The problem is that the canonical data format has changed (see https://github.com/exercism/problem-specifications/commit/e0ffb0085591467aa946962310d3da6fbbbf9483#diff-c24381733233cf7c7744c5c45cf840e0R80). This means that you will have to change the test generator to handle the changed canonical data. 

Usually, if the canonical data's expected value is an error object (which it is now), we convert that to an error. However, in this case I think it makes sense to just use `-1` as the expected value instead.

abecerramatias: Thanks for the tip, I will take a look.

ErikSchierboom:
Thanks! 🎉

PR #668, opened 2018-10-03
Update test suite for exercise: isogram

yvincent:
As per instructed:
 - Regenerated test class.
 - No need to change the generator.
 - Tests passed.


ErikSchierboom:
Thanks for doing this! 🎉

PR #673, opened 2018-10-03
update queen attack tests

### LINKED ISSUE #649: Update test suite for exercise: queen-attack
The `queen-attack` exercise's [test suite](https://github.com/exercism/csharp/blob/master/exercises/queen-attack/QueenAttackTest.cs) is automatically generated using its [test generator](https://github.com/exercism/csharp/blob/master/generators/Exercises/Generators/QueenAttack.cs). This test generator converts the `queen-attack`'s [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/queen-attack/canonical-data.json) into the test suite just mentioned.

The canonical data for the `queen-attack` exercise has been updated since the test suite was last generated. We should be updating thetest suite to the latest version. To do this, one should:
- Run the test generator for the exercise. This is done by opening a command prompt in the `generators` directory and running: `dotnet run -e queen-attack`.
- Finally, please make sure that the [example implementation](https://github.com/exercism/csharp/blob/master/exercises/queen-attack/Example.cs) passes all the tests. You can verify this by opening a command prompt in the root directory and running `./build.sh --exercise=queen-attack` or `.\build.ps1 -Exercise queen-attack`, depending on your operating system.

bmeverett:
Fixes #649

### DIFF
--- a/generators/Exercises/Generators/QueenAttack.cs
+++ b/generators/Exercises/Generators/QueenAttack.cs
@@ -17,7 +17,7 @@ namespace Exercism.CSharp.Exercises.Generators
 
         private static void UpdateTestMethodForCreateProperty(TestMethod testMethod)
         {
-            if (testMethod.Expected < 0)
+            if (testMethod.Expected is Dictionary<string, object> || testMethod.Expected < 0)
             {
                 testMethod.ExceptionThrown = typeof(ArgumentOutOfRangeException);
             }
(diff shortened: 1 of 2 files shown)


generators/Exercises/Generators/QueenAttack.cs
```diff
@@ -17,7 +17,7 @@ protected override void UpdateTestMethod(TestMethod testMethod)
 
         private static void UpdateTestMethodForCreateProperty(TestMethod testMethod)
         {
-            if (testMethod.Expected < 0)
+            if (testMethod.Expected is Dictionary<string, object> || testMethod.Expected < 0)
```
ErikSchierboom: I think the `testMethod.Expected < 0` can be removed.

ErikSchierboom:
Great work! 🎉

PR #675, opened 2018-10-03
FORTH: update to generate InvalidOperationException from canonical data

bjorkqvist:
As per instructed in #648 :

Regenerated test class.
Changed generator to generate same testfile.
Tests passed.

### DIFF
--- a/generators/Exercises/Generators/Forth.cs
+++ b/generators/Exercises/Generators/Forth.cs
@@ -1,5 +1,5 @@
-﻿using System;
-using Exercism.CSharp.Output;
+﻿using Exercism.CSharp.Output;
+using System;
 
 namespace Exercism.CSharp.Exercises.Generators
 {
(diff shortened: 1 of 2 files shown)


generators/Exercises/Generators/Forth.cs
```diff
@@ -9,7 +9,7 @@ protected override void UpdateTestMethod(TestMethod testMethod)
         {
             testMethod.TestMethodName = testMethod.TestMethodNameWithPath;
 
-            if (testMethod.Expected == null)
+            if (testMethod.Expected == null || string.Join(" ", testMethod.Expected).StartsWith("[error"))
```
ErikSchierboom: Is the `null` check still necessary? I think you could replace the if-expression with `testMethod.Expected is Dictionary<string, object>`, which is how other generators also do this check.

bjorkqvist:
Sure, i will give it a try. 😊

Den tors 4 okt. 2018 08:05Erik Schierboom <notifications@github.com> skrev:

bjorkqvist:
ok, so Emazculator is an old account of mine. Don't understand  why it got the latest commit. The commit should be fine, but i'm deleting my old account to avoid future confusion.

ErikSchierboom:
Merged. Thanks a lot! 🎉

PR #677, opened 2018-10-03
Add generator for exercise: pythagorean-triplet

joostas:
Implemented test generation based on canonical data. But it required to add additional methods to PythagoreanTriplet.cs implementation for item comparison.
By the way. Last generated test method with given example implementation runs too long.


ErikSchierboom:
I think that in this case, the changed canonical has [changed so much](https://github.com/exercism/problem-specifications/blob/master/exercises/pythagorean-triplet/canonical-data.json), that we should consider restructuring the exercise. Having looked at the new canonical data, I think I would prefer the following changes:

- Convert the class name from `Triplet` to `PythagoreanTriplet`.
- Make the class a `static` class, with a single, static `TripletsWithSum` method.
- Return a tuple instead of the class itself, as this is a perfect use case for using tuples. As tuples have built-in equality, this also makes the tests easier to write. For more information on tuples, check: https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7#tuples

Would you be comfortable with making these changes? If not, feel free to let me know.

joostas:
Will try.

exercises/pythagorean-triplet/PythagoreanTriplet.cs
```diff
@@ -1,28 +1,9 @@
...
-        throw new NotImplementedException("You need to implement this function.");
-    }
-
-    public bool IsPythagorean()
-    {
-        throw new NotImplementedException("You need to implement this function.");
-    }
-
-    public static IEnumerable<Triplet> Where(int maxFactor, int minFactor = 1, int sum = 0)
+    public static IEnumerable<(int a, int b, int c)> TripletWithSum(int sum)
```
ErikSchierboom: `TripletWithSum` should be `TripletsWithSum`

joostas:
HI, I see strange error message from Travis CI build:
`No executable found matching command "dotnet-/home/travis/build/exercism/csharp/tools/Cake.0.27.1/Cake.dll"`
Maybe you could help to find how to solve it?

ErikSchierboom:
> Maybe you could help to find how to solve it?

I've restarted the Travis build, which occasionally fails.

generators/Exercises/Generators/PythagoreanTriplet.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Collections.Generic;
+using System.Text;
+using Exercism.CSharp.Exercises;
+using Exercism.CSharp.Output;
+using Exercism.CSharp.Output.Rendering;
+using Newtonsoft.Json.Linq;
+
+namespace generators.Exercises.Generators
+{
+    public class PythagoreanTriplet : GeneratorExercise
```
ErikSchierboom: You can simplify this generator quite a bit, using the existing generator framework methods:

```csharp
public class PythagoreanTriplet : GeneratorExercise
{
    {
        testMethod.Expected = ConvertExpected(testMethod.Expected);
    }

    private dynamic ConvertExpected(dynamic value)
    {
        int[][] values = value.ToObject<int[][]>();
        var triplets = values.Select(ints => (ints[0], ints[1], ints[2])).ToArray();
        return new UnescapedValue(Render.ArrayMultiLine(triplets));
    }

    protected override void UpdateNamespaces(ISet<string> namespaces)
    {
        namespaces.Add(typeof(Array).Namespace);
    }
}
```

This generator should behave the same way as your version and is a bit simpler.

joostas: Nice! It seems much simpler.
I missed the point that Actual value can be left untouched.
Jagged array also fits very nice here.

PR #682, opened 2018-10-04
Changing test generator for specific exercise (Wordy) - Fix for Issue #637

deniscapeto:
Now if you generate the test you can run it right away. For Wordy exercise the answer must be an integer. If the answer in canonical data is not int the exercise should throw an exception. This fix Issue #637

### DIFF
--- a/exercises/wordy/WordyTest.cs
+++ b/exercises/wordy/WordyTest.cs
@@ -1,4 +1,4 @@
-// This file was auto-generated based on version 1.1.0 of the canonical data.
+// This file was auto-generated based on version 1.2.0 of the canonical data.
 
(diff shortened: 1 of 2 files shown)


ErikSchierboom:
The change is looking good! Did you re-run the test generator (the issue comment describes how to do this)? I would expect the canonical data version in the `WordTest.cs` file to be updated. If you didn't, could you run the generator to regenerate the test file?

deniscapeto:
You are right. The  `WordTest.cs` is being generated correctly but I didn't update it. I'll do it.

ErikSchierboom:
Merged, thanks! 🎉

PR #683, opened 2018-10-04
Fixed #654 - Updated Anagram Tests

### LINKED ISSUE #654: Update test suite for exercise: anagram
The `anagram` exercise's [test suite](https://github.com/exercism/csharp/blob/master/exercises/anagram/AnagramTest.cs) is automatically generated using its [test generator](https://github.com/exercism/csharp/blob/master/generators/Exercises/Generators/Anagram.cs). This test generator converts the `anagram`'s [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/anagram/canonical-data.json) into the test suite just mentioned.

The canonical data for the `anagram` exercise has been updated since the test suite was last generated. We should be updating the testsuite to the latest version. To do this, one should:
- Run the test generator for the exercise. This is done by opening a command prompt in the `generators` directory and running: `dotnet run -e anagram`.
- Finally, please make sure that the [example implementation](https://github.com/exercism/csharp/blob/master/exercises/anagram/Example.cs) passes all the tests. You can verify this by opening a command prompt in the root directory and running `./build.sh --exercise=anagram` or `.\build.ps1 -Exercise anagram`, depending on your operating system.

irrationalRock:
This PR addresses https://github.com/exercism/csharp/issues/654.

I changed `anagrams` to `FindAanagrams` in both `Example.cs` and `Anagram.cs`.

This is because I noticed that the property was changed in https://github.com/exercism/problem-specifications/blob/master/exercises/anagram/canonical-data.json.

It also passed all of the tests when running `./build.sh --exercise=anagram`.

Thanks for considering this request.

### DIFF
--- a/exercises/anagram/Anagram.cs
+++ b/exercises/anagram/Anagram.cs
@@ -7,7 +7,7 @@ public class Anagram
         throw new NotImplementedException("You need to implement this function.");
     }
 
-    public string[] Anagrams(string[] potentialMatches)
+    public string[] FindAnagrams(string[] potentialMatches)
     {
         throw new NotImplementedException("You need to implement this function.");
     }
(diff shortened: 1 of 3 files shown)


robkeim:
Thanks @irrationalRock! These changes look good to me.

PR #674, opened 2018-10-03
Add function to check outdated exercises

EarthlingRich:
This PR lets users run ```dotnet run -s Outdated``` to output exercises that use outdated canonical data.

See issue #631.

PR #684, opened 2018-10-04
Fix generating exercises

EarthlingRich:
Generating exercises with ```dotnet run -e <exercise>``` currently does not work. Because of the Outdated implementation of #674. This PR fixes this. As an added bonus the re-generating text will only be printed if the -e option is used.

### DIFF
--- a/generators/Exercises/ExerciseCollection.cs
+++ b/generators/Exercises/ExerciseCollection.cs
@@ -15,7 +15,7 @@ namespace Exercism.CSharp.Exercises
         private readonly CanonicalDataFile _canonicalDataFile;
         private readonly Dictionary<string, Type> _exerciseTypesByName;
 
-        public ExerciseCollection(CanonicalDataFile canonicalDataFile)
+        public ExerciseCollection(CanonicalDataFile canonicalDataFile, Options options)
             => (_canonicalDataFile, _exerciseTypesByName) = (canonicalDataFile, GetExerciseTypesByName());
 
         private static Dictionary<string, Type> GetExerciseTypesByName()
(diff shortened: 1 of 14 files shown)


generators/Exercises/ExerciseCollection.cs
```diff
@@ -35,13 +36,14 @@ private IEnumerable<Exercise> GetExercises()
...
                 var exerciseName = exercise.Slug.ToExerciseName();
+
                 if (exercise.Deprecated)
                     yield return new DeprecatedExercise(exerciseName);
                 else if (HasNoCanonicalData(exerciseName))
                     yield return new MissingDataExercise(exerciseName);
                 else if (IsNotImplemented(exerciseName))
                     yield return new UnimplementedExercise(exerciseName);
-                else if (IsOutdated(exerciseName))
+                else if (IsOutdated(exerciseName) && !_options.ShouldGenerate)
```
ErikSchierboom: If you replace `!_options.ShouldGenerate` with ` && _options.Status == GeneratorStatus.Outdated`, it will ensure that outdated exercises are only returned when the outdated status is requested. The current code works when specifying a single exercise, but it does not work when just doing `dotnet run` to regenerate all exercises. In that case, the outdated exercises are listed as outdated and are not updated.

Another possible approach we could consider is adding a `Outdated` property to the `GeneratorExercise` class, and then add some additional logic in the `RegenerateTestClass` method to handle this case. I think I would actually prefer this latter approach, as an exercise can semantically be both a generator exercise _and_ an outdated exercise. What do you think?

ErikSchierboom: I've merged a [temporary fix](#686) by the way, so take your time to implement this.

EarthlingRich:
No problem. Experimenting with different solutions sometimes is necessary. I think the current implementation should be ok.

ErikSchierboom:
Thanks! 🎉

Issue #632, opened 2018-10-02
Implement new exercise: rest-api


ErikSchierboom:
There is a new exercise, [rest-api](https://github.com/exercism/problem-specifications/blob/master/exercises/rest-api/description.md), which data can be found here: https://github.com/exercism/problem-specifications/tree/master/exercises/rest-api

To implement the `rest-api` exercise, the following needs to be done:

- Create a directory for the new exercise in the `exercises` directory (note: this has to match the exact exercise name).
- Add a new entry to the [config.json](https://github.com/exercism/csharp/blob/master/config.json) file.
- Create a generator to automatically convert the [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/rest-api/canonical-data.json) to a test file. For more information on how to do this, check the [generators docs](https://github.com/exercism/csharp/blob/master/docs/GENERATORS.md).
- Create a stub implementation file that contains the code necessary to compile (which are the methods called by the test file), but lacks an implementation. (see [this example](https://github.com/exercism/csharp/blob/master/exercises/two-fer/TwoFer.cs)).
- Create an implementation file that passes all the tests.
- Create a project file for the project (you can probably best copy and modify one of the existing projects).

yvincent:
I'll try this one.

ErikSchierboom:
That would be great!

ErikSchierboom:
Closed by #687.

PR #687, opened 2018-10-05
RestAPI exercise

yvincent:
PR for issue #632 
- Created new project in Exercise solution with Test class, empty implementation, README and implementation example.
- Created new generator for REST api exercise.
- Added entry in config.json file (used default values, not sure what to use).

### DIFF
--- a/config.json
+++ b/config.json
@@ -1293,9 +1293,12 @@
       "slug": "rest-api",
       "uuid": "4c89b6a8-0e89-46c9-5a7e-31bfe6a57007",
       "core": false,
-      "unlocked_by": null,
-      "difficulty": 0,
-      "topics": null,
+      "unlocked_by": "markdown",
+      "difficulty": 6,
+      "topics": [
+        "json",
+        "parsing"
+      ],
       "deprecated": false
     },
     {
--- a/config.json
+++ b/config.json
@@ -1290,6 +1290,15 @@
       ]
     },
     {
+      "slug": "rest-api",
+      "uuid": "4c89b6a8-0e89-46c9-5a7e-31bfe6a57007",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 0,
+      "topics": null,
+      "deprecated": false
+    },
+    {
       "slug": "binary",
       "uuid": "cef7deef-54ce-4201-b263-7cd2098533f8",
       "core": false,
--- a/exercises/rest-api/Example.cs
+++ b/exercises/rest-api/Example.cs
@@ -1,0 +1,126 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using Newtonsoft.Json;
+
+public class User
+{
+    public string name { get; }
+    public IDictionary<string, double> owes { get; private set; }
+    public IDictionary<string, double> owed_by { get; private set; }
+    public double balance => owed_by.Sum(x => x.Value) - owes.Sum(x => x.Value);
+
+    public User(string name)
+    {
+        this.name = name;
+        owes = new Dictionary<string, double>();
+        owed_by = new Dictionary<string, double>();
+    }
+
+    public void Lend(User borrower, double amount)
+    {
+        var remaining = amount;
+        if (owes.ContainsKey(borrower.name))
+        {
+            remaining = owes[borrower.name] - amount;
+            if (remaining > 0)
+            {
+                owes[borrower.name] = remaining;
+                return;
+            }
+
+            owes.Remove(borrower.name);
+            remaining *= -1;
+        }
+
+        if (remaining > 0)
+        {
+            if (owed_by.ContainsKey(borrower.name))
+            {
+                owed_by[borrower.name] += remaining;
+            }
+            else
+            {
+                owed_by.Add(borrower.name, remaining);
+                owed_by = owed_by.OrderBy(x => x.Key).ToDictionary(x => x.Key, x => x.Value);
+            }
+        }
+    }
+
+    public void Borrow(User lender, double amount)
+    {
+        var remaining = amount;
+        if (owed_by.ContainsKey(lender.name))
+        {
+            remaining = owed_by[lender.name] - amount;
+            if (remaining > 0)
+            {
+                owed_by[lender.name] = remaining;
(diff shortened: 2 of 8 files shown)


exercises/rest-api/Example.cs
```diff
@@ -0,0 +1,126 @@
...
+        else if (url == "/iou")
+        {
+            var values = JsonConvert.DeserializeObject<Dictionary<string, object>>(payload);
+            var lender = users.First(x => x.name.Equals(values["lender"]));
+            var borrower = users.First(x => x.name.Equals(values["borrower"]));
+            var amount = (double)values["amount"];
+            lender.Lend(borrower, amount);
+            borrower.Borrow(lender, amount);
+
+            // Replace to workaround inconsistency in specs
```
ErikSchierboom: Could you perhaps create an issue at https://github.com/exercism/problem-specifications/issues for any inconsistencies in the canonical data?

exercises/rest-api/RestApi.cs
```diff
@@ -0,0 +1,38 @@
...
+    }
+}
+
+public class RestApi
+{
+    private IList<User> users;
+
+    public RestApi(string database)
+    {
+        users = JsonConvert.DeserializeObject<List<User>>(database);
```
ErikSchierboom: Could replace this line with `throw new NotImplementedException("You need to implement this function.");`?

exercises/rest-api/RestApi.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        Name = name;
+        Owes = new Dictionary<string, double>();
+        Owed_By = new Dictionary<string, double>();
+    }
+}
+
+public class RestApi
+{
+    private IList<User> users;
```
ErikSchierboom: Could you remove this field? It is not necessary for the public test API and a user should be free to choose the underlying representation he/she chooses.

exercises/rest-api/RestApi.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+using System.Collections.Generic;
+using Newtonsoft.Json;
+
+public class User
```
ErikSchierboom: Could you remove this class? It is not necessary for the public test API and a user should be free to choose the underlying representation he/she chooses.

config.json
```diff
@@ -1289,6 +1289,15 @@
         "searching"
       ]
     },
+    {
+      "slug": "rest-api",
+      "uuid": "4c89b6a8-0e89-46c9-5a7e-31bfe6a57007",
+      "core": false,
+      "unlocked_by": null,
```
ErikSchierboom: I don't think this should be a core exercise at the moment, which means that it has to be unlocked by a core exercise. Let's first come up with a difficulty and some topics, as they will help in finding a suitable core exercise.

config.json
```diff
@@ -1289,6 +1289,15 @@
...
+      "difficulty": 0,
+      "topics": null,
```
ErikSchierboom: Do you have any suggestions for topics? There is a reference list to be found [here](https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt). From that list, I think not much is relevant here, but we might use a new topic here: "json". What do you think?

generators/Exercises/Generators/RestApi.cs
```diff
@@ -0,0 +1,30 @@
+﻿using Exercism.CSharp.Output;
+using Newtonsoft.Json;
+
+namespace Exercism.CSharp.Exercises.Generators
+{
+    public class RestApi : GeneratorExercise
+    {
+        protected override void UpdateTestMethod(TestMethod testMethod)
+        {
+            testMethod.UseVariableForTested = true;
```
ErikSchierboom: I think it might be useful to also use variables for the expected and input:

```
testMethod.UseVariablesForInput = true;
testMethod.UseVariableForTested = true;
testMethod.UseVariableForExpected = true;
```

Could you create a version that has the above code to see if that reads a bit better?

yvincent:
Last commit in master added a new constructor parameter 'option' for ExerciseCollection and my branch wasn't updated with that.
However the parameter is not used in the constructor, so I wonder if it should be there in the first place?

config.json
```diff
@@ -1289,6 +1289,15 @@
         "searching"
       ]
     },
+    {
+      "slug": "rest-api",
+      "uuid": "4c89b6a8-0e89-46c9-5a7e-31bfe6a57007",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 0,
```
ErikSchierboom: What do you consider the difficulty of this exercise to be? To me, the difficulty is mainly due to keeping an administration of users and dealing with JSON. My gut feeling would put it at difficulty 5 or 6, due to the relative large amount of code one has to write (compared to other exercises).

exercises/rest-api/Example.cs
yvincent: Done ([https://github.com/exercism/problem-specifications/issues/1353](https://github.com/exercism/problem-specifications/issues/1353)).

config.json
yvincent: Markdown could probably be a good core exercise to unlock this one (because of the 'parsing' topic).

config.json
yvincent: json seems appropriate for this exercise. I would probably use parsing as well from the reference list.

config.json
yvincent: Not exactly sure of the difficulty compared to different exercise. When I look at Markdown exercise, I would probably make Rest Api difficulty a little bit harder than markdown, so a 6 seems good.

generators/Exercises/Generators/RestApi.cs
yvincent: Done.

ErikSchierboom: I like the effect the change has! Maybe also use `testMethod.UseVariablesForConstructorParameters = true;`?

ErikSchierboom:
There was a build error, but it has nothing to do with this PR. Thanks a ton for doing this! 🎉

yvincent:
No problem, that was fun. Cheers.

PR #688, opened 2018-10-05
add Affine cipher

davidelettieri:
add affine cipher exercise, info for config file is taken from ruby project

### DIFF
--- a/config.json
+++ b/config.json
@@ -1329,8 +1329,8 @@
       "slug": "affine-cipher",
       "uuid": "2462D9AE-6483-40C6-955A-79CB2AC25B34",
       "core": false,
-      "unlocked_by": null,
-      "difficulty": 3,
+      "unlocked_by": "bob",
+      "difficulty": 4,
       "topics": [
         "cryptography",
         "math",
--- a/config.json
+++ b/config.json
@@ -1324,6 +1324,18 @@
       "difficulty": 0,
       "topics": null,
       "deprecated": true
+    },
+    {
+      "slug": "affine-cipher",
+      "uuid": "2462D9AE-6483-40C6-955A-79CB2AC25B34",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 3,
+      "topics": [
+        "cryptography",
+        "math",
+        "strings"
+      ]
     }
   ]
 }
--- a/exercises/affine-cipher/AffineCipher.cs
+++ b/exercises/affine-cipher/AffineCipher.cs
@@ -1,0 +1,14 @@
+﻿using System;
+
+public class AffineCipher
+{
+    public string Encode(string plainText, int a, int b)
+    {
+        throw new NotImplementedException("You need to implement this function.");
+    }
+
+    public string Decode(string cipheredText, int a, int b)
+    {
+        throw new NotImplementedException("You need to implement this function.");
+    }
+}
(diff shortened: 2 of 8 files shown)


exercises/affine-cipher/README.md
```diff
@@ -0,0 +1,96 @@
...
+  - simple example:
+    - `9 mod 26 = 9`
+    - `9 * 3 mod 26 = 27 mod 26 = 1`
+    - `3` is the MMI of `9 mod 26`
+  - a more complicated example:
+    - `15 mod 26 = 15`
+    - `15 * 7 mod 26 = 105 mod 26 = 1`
+    - `7` is the MMI of `15 mod 26`
+
+## Exception messages
```
ErikSchierboom: Could you remove this section? At this point in the language track, the student will already have encountered exceptions, so this is not really needed.

generators/Exercises/Generators/AffineCipher.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Exercism.CSharp.Output;
+
+namespace Exercism.CSharp.Exercises.Generators
+{
+    public class AffineCipher : GeneratorExercise
```
ErikSchierboom: There is quite a bit of customization going on here, and I think we can simplify things quite a bit. Let me explain!

Looking at the [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/affine-cipher/canonical-data.json), the input data is straightforward, except for the `"key"` property, which is a nested property. You can work around this using the following code in the `UpdateTestMethod` method:

```csharp
testMethod.Input["a"] = testMethod.Input["key"]["a"];
testMethod.Input["b"] = testMethod.Input["key"]["b"];
testMethod.InputParameters = new[] { "phrase", "a", "b" };
```

The second bit is dealing with exception, which is also easy to fix:

```csharp
if (!(testMethod.Expected is string))
     testMethod.ExceptionThrown = typeof(ArgumentException);
```

With just these 5 lines of code, the generate becomes a lot simpler and still generates a nice test file. The only change you have to make in your example and stub files is make them use a static class with static methods.

config.json
```diff
@@ -1324,6 +1324,18 @@
       "difficulty": 0,
       "topics": null,
       "deprecated": true
+    },
+    {
+      "slug": "affine-cipher",
+      "uuid": "2462D9AE-6483-40C6-955A-79CB2AC25B34",
+      "core": false,
+      "unlocked_by": null,
```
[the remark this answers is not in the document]
davidelettieri: done

exercises/affine-cipher/AffineCipher.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+
+public class AffineCipher
```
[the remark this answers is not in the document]
davidelettieri: Done

generators/Exercises/Generators/AffineCipher.cs
davidelettieri: Generator is considerably better, we're missing check on exception message. Is there a simple way to do it?

ErikSchierboom:
This is looking great! I've solved the solution conflict. When the CI is green, I'll merge!

ErikSchierboom:
Merged, thanks a lot! 🎉

davidelettieri:
Thanks to you :)

PR #690, opened 2018-10-06
Rest API test bug fix

ChrisPritchard:
One of the tests incorrectly specifies -6 rather than -6.0 for a json number field. This inconsistent with the other tests

### DIFF
--- a/exercises/rest-api/Example.cs
+++ b/exercises/rest-api/Example.cs
@@ -117,8 +117,7 @@ public class RestApi
             lender.Lend(borrower, amount);
             borrower.Borrow(lender, amount);
 
-            // Replace to workaround inconsistency in specs
-            return JsonConvert.SerializeObject(new[] { lender, borrower }.OrderBy(x => x.name)).Replace("-6.0", "-6");
+            return JsonConvert.SerializeObject(new[] { lender, borrower }.OrderBy(x => x.name));
         }
 
         return String.Empty;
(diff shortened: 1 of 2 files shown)


ErikSchierboom:
Hi! The test suite for the REST API exercises is automatically generated based on the canonical data. There is a [PR[(https://github.com/exercism/problem-specifications/pull/1355) that will fix this.

As for this PR, it appears that the current implementation does not pass the tests. Could you look into that?

ChrisPritchard:
It fails because of the issue this PR is trying to resolve :) Ill close the PR since there is already a fix in the wings, cheers

ErikSchierboom:
Ah.  I understand now. I'll merge then.

ErikSchierboom:
But shouldnt the build use your fix?

ChrisPritchard:
Not too sure how the CI works, but it seems that the source solution it tests against is providing the incorrect value (-6) rather than what the corrected test now expects (-6.0)

jpreese:
@ChrisPritchard our CI will take the implementation found in the exercise's Example.cs file and run the suite of tests against the example.

As @ErikSchierboom pointed out, we don't want to be modifying the test suite directly, but rather wait until the canonical data is merged in and then run the test generator against the new data.

However, this gives us a glance into what is _going_ to happen when that data gets updated. It looks like our Example.cs does some smelly stuff.

https://github.com/exercism/csharp/blob/5f5968aeba4eec5e51dc9f987a94c765ef265640/exercises/rest-api/Example.cs#L121

We'll need to update our Example.cs for the RestAPI exercise in order to fix the build.

jpreese:
We'll need to do something. Your change is essentially whats going to happen when the canonical data is updated. Once that happens, we'll be back to our build failing. 

The simple solution is just updating Example.cs to not convert -6.0 to -6

ErikSchierboom:
@ChrisPritchard Maybe you could also fix the `Example.cs` file? That has to be done anyway, and that way we will be prepared for it.

ChrisPritchard:
Sure, done.

ErikSchierboom:
Thanks for doing this! 🎉

PR #695, opened 2018-10-10
Updates the README files for various exercises

Zureka:
This is the result of running the `./bin/configlet generate` command against the master branch of the problem-specifications repository. Please let me know if there is a better way for merging these types of changes.

### DIFF
--- a/exercises/accumulate/README.md
+++ b/exercises/accumulate/README.md
@@ -17,25 +17,20 @@ Your code should be able to produce the collection of squares:
 
 - 1, 4, 9, 16, 25
 
-Check out the test suite to see the expected method signature.
+Check out the test suite to see the expected function signature.
 
 ## Restrictions
 
-Keep your hands off that select/collect/map/fmap/whatchamacallit functionality
+Keep your hands off that collect/map/fmap/whatchamacallit functionality
 provided by your standard library!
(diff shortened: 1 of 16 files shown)


ErikSchierboom:
Hi! Actually, to update the docs we have a customs step (in which some readme files are patched). You can run this custom step by running the `update-docs.sh` command in the root directory.

Zureka:
@ErikSchierboom Ahh I see, I did get a little confused when I noticed all of the occurrences of the word "method" changed to "function" in all of these C# projects. 

In terms of updating these README files, would it be better to do them one exercise at a time with any potential generator updates that might be necessary? Or is a batch README update like this acceptable?

Zureka:
I mean... I would appreciate some Hacktoberfest PR's 😉. But I wasn't sure if the README updates would reflect changes for exercises that needed to be re-generated. 

I can do individual PR's if that's what you prefer.

ErikSchierboom:
I'm fine either way, whichever suits you best.

Zureka:
I'll close this PR and update the README files as part of any exercise updates that are necessary. That way I can familiarize myself with running/modifying the generators.

ErikSchierboom:
Remember that it is also fine to just send PR's with just the README updates. As I think most exercises will have been updated to the latest canonical data.

PR #697, opened 2018-10-12
Go Counting Fix - TerritoryFor to Territory

ChrisPritchard:
The basic template doesnt match the tests: the method TerritoryFor should be named Territory


jpreese:
Yeah, looks right to me. The example uses Terrirtory (or else the CI would fail) and the problem-specs use the Territory property as well. 👍

Issue #700, opened 2018-10-12
CI: fail fast when generators project does not build correctly


ErikSchierboom:
In our [Cake build script](https://github.com/exercism/csharp/blob/master/build.cake#L87), we verify that the generators project can be built successfully. At the moment, we do this _after_ the exercises have been tested. This order should be switched, to make the script fail as fast as possible if there is an error.

ErikSchierboom:
Closed by #701.

PR #701, opened 2018-10-12
Switch BuildGenerators and tests running order

avjgit:
https://github.com/exercism/csharp/issues/700

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -70,21 +70,21 @@ Task("ReplaceStubWithExample")
         }
     });
 
-Task("TestUsingExampleImplementation")
-    .IsDependentOn("ReplaceStubWithExample")
-    .Does(() => {
-        var allProjects = GetFiles(exercisesBuildDir + "/*/*.csproj");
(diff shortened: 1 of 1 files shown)


build.cake
```diff
@@ -70,21 +70,21 @@ Task("ReplaceStubWithExample")
...
-        var allProjects = GetFiles(exercisesBuildDir + "/*/*.csproj");
-        Parallel.ForEach(allProjects, parallelOptions, (project) => DotNetCoreTest(project.FullPath));
-    });
 
 Task("BuildGenerators")
     .Does(() => {
        DotNetCoreBuild("./generators/Generators.csproj");
     });
 
+Task("TestUsingExampleImplementation")
```
ErikSchierboom: I understand the idea of moving this task below the BuildGenerators task, but I'd rather the BuildGenerators task be placed higher in the build script as the TestUsingExampleImplementation task is dependent on the tasks above it, and I'd prefer keeping them listed sequentially.

avjgit: Agree, fixed

ErikSchierboom:
Thanks for this! 🎉

PR #703, opened 2018-10-15
Fixes exception in AffineCipher generator

Zureka:
When executing `dotnet run` in the **generators** directory, the **AffineCipher** generator throws this exception: 

> Microsoft.CSharp.RuntimeBinder.RuntimeBinderException: Operator '!' cannot be applied to operand of type 'string'

Adding the additional parentheses corrects this issue.


ErikSchierboom:
Thanks! 🎉

Issue #698, opened 2018-10-12
CI: optimize build by using solution file to test


ErikSchierboom:
At the moment, the [Cake build script](https://github.com/exercism/csharp/blob/master/build.cake) tests all exercises using parallel `dotnet test <project>` commands. We should examine if testing using the [exercises/Exercises.sln](https://github.com/exercism/csharp/blob/master/exercises/Exercises.sln) solution file gives us any performance benefits.

ErikSchierboom:
Closed by #704

PR #704, opened 2018-10-15
#698 test using exercise solution file

bmeverett:
Test using the solution file. Figured I kick off the build and we can measure how it does. Seemed to be faster when I was testing locally.


bmeverett:
`build.sh` and `build.ps1` seemed to work locally, but not working on CI. I wonder if the .sln file isn't in the build dir

ErikSchierboom:
This is great! You've managed to shave of 10 minutes of our build time with this change! Thanks a lot. 🎉

PR #705, opened 2018-10-15
Fixes 'expected error' based on canonical data

Zureka:
Based on the canonical data for this exercise, it appears that a `-1` is no longer the expected value for an error. The *expected* value is now a Dictionary based on the error message:

> Microsoft.CSharp.RuntimeBinder.RuntimeBinderException: Operator '==' cannot be applied to operands of type 'System.Collections.Generic.Dictionary<string,object>' and 'int'

Here's the error case from the canonical data:

```
{
    "description": "rejects negative span",
    "property": "largestProduct",
    "input": {
        "digits": "12345",
        "span": -1
    },
    "expected": {"error": "span must be greater than zero"}
}
```


Zureka:
@ErikSchierboom Should be all set now!

ErikSchierboom:
Thanks! 🎉

PR #707, opened 2018-10-17
Working word count with uncommented tests

najtmarr:
What about situation with quoted word with aposthophe? For example 'can't'? There is no unit test for that case.

### DIFF
--- a/exercises/word-count/WordCount.cs
+++ b/exercises/word-count/WordCount.cs
@@ -1,10 +1,46 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
 
 public static class WordCount
 {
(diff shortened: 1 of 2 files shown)


najtmarr:
This pull request contains working solution which passes all tests. In test file there is no changes, except for delete "skip" from attributes. Additionally in comment i just asked if there shouldn't be one more test case.

ErikSchierboom:
> This pull request contains working solution which passes all tests. 

The current solution already passes all the tests.

> In test file there is no changes, except for delete "skip" from attributes.

Those Skip attributes are key to how one is expected to solve an exercise: one solution at a time.

> Additionally in comment i just asked if there shouldn't be one more test case.

As I explained in my comment above, that is something that you should be discussing in the problem/specifications repo.

PR #711, opened 2018-10-26
use project file if only running single exercise

### LINKED ISSUE #710: Building with solution doesn't let you run individual exercises
In #698 we changed the build to use the solution file. If you build an individual exercise using `.\build.ps1 -Exercise darts` the `Exercises.sln` doesn't get copied over

bmeverett:
Fixes #710


ErikSchierboom:
Thanks!

PR #725, opened 2018-11-07
robot-simulator: cleanup generator and remove unused methods from stub

ErikSchierboom:
The concensus was to be more consistent :) So I've changed the name back to RobotSimulator, and have also cleaned up the generator a bit and remove some test methods in the stub file that the user appeared to have to implement, but actually doesn't.


robkeim:
Looks good, thanks @ErikSchierboom! 🎉

PR #726, opened 2018-11-09
Remind learners to remove Skip properties from tests (in Leap)

j2jensen:
Also combined the two "Notes" sections under one header.

### DIFF
--- a/config/exercise_readme.go.tmpl
+++ b/config/exercise_readme.go.tmpl
@@ -8,7 +8,7 @@
 
 To run the tests, run the command `dotnet test` from within the exercise directory.
 
-Most of the tests provided in the C# track's exercises have a `Skip` property set on them.
+Initially, only the first test will be enabled. This is to encourage you to solve the exercise one step at a time.
 Once you get the first test passing, remove the `Skip` from another test and get it passing.
 Once none of the tests are skipped and they are all passing, you can submit your solution 
 using `exercism submit {{ .Spec.Name }}.cs`
--- a/config/exercise_readme.go.tmpl
+++ b/config/exercise_readme.go.tmpl
@@ -9,7 +9,7 @@
 To run the tests, run the command `dotnet test` from within the exercise directory.
 
 Initially, only the first test will be enabled. This is to encourage you to solve the exercise one step at a time.
-Once you get the first test passing, remove the `Skip` from another test and get it passing.
+Once you get the first test passing, remove the `Skip` property from the next test and work on getting that test passing.
 Once none of the tests are skipped and they are all passing, you can submit your solution 
 using `exercism submit {{ .Spec.MixedCaseName }}.cs`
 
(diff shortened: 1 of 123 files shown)


exercises/leap/README.md
```diff
@@ -18,23 +18,27 @@ this look-up, pretend it doesn't exist and implement it yourself.
...
-
-The DateTime class in C# provides a built-in [IsLeapYear](https://msdn.microsoft.com/en-us/library/system.datetime.isleapyear(v=vs.110).aspx) method
+- The DateTime class in C# provides a built-in [IsLeapYear](https://msdn.microsoft.com/en-us/library/system.datetime.isleapyear(v=vs.110).aspx) method
 which you should pretend doesn't exist for the purposes of implementing this exercise.
 
 ## Running the tests
 
 To run the tests, run the command `dotnet test` from within the exercise directory.
 
+Most of the tests provided in the C# track's exercises have a `Skip`
```
ErikSchierboom: I would perhaps slightly rephrase this, to make it more clear what exactly the student has to do: solve the first test, unskip the next none, solve that test and so on. 

> Your work is not ready to submit until you have removed all the `Skip`
attributes and all the tests are passing.

This sounds a little bit harsh, maybe you could rephrase to something like: "Once all tests are passing, you can submit your solution using `exercism submit <exercise>.cs`? That way we also help the users understand how to submit the correct file (which sometimes they don't understand).

j2jensen:
@ErikSchierboom That makes sense, and I'll try to get to that soon.

exercises/leap/README.md
j2jensen: I like that phrasing. Thanks for the suggestion.

j2jensen:
@ErikSchierboom I've updated the template file, but unfortunately lack a linux machine to run the shell script from. I tried using git bash, and got the following error from the `fetch-configlet` command:

```
gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```

Not sure what to do with the pull request at this point.

config/exercise_readme.go.tmpl
```diff
@@ -8,6 +8,11 @@
 
 To run the tests, run the command `dotnet test` from within the exercise directory.
 
+Most of the tests provided in the C# track's exercises have a `Skip` property set on them.
+Once you get the first test passing, remove the `Skip` from another test and get it passing.
```
ErikSchierboom:
```suggestion
Once you get the first test passing, remove the `Skip` property from the next test and work on getting that test passing.
```

config/exercise_readme.go.tmpl
```diff
@@ -8,6 +8,11 @@
 
 To run the tests, run the command `dotnet test` from within the exercise directory.
 
+Most of the tests provided in the C# track's exercises have a `Skip` property set on them.
```
ErikSchierboom:
```suggestion
Initially, only the first test will be enabled. This is to encourage you to solve the exercise one step at a time.
```

I think it is important to remind the students of why we have this practice.

ErikSchierboom:
@j2jensen PR #730 has just been merged, which adds a `update-docs.ps1` script with which you can easily regenerate the docs on Windows.

j2jensen:
Yes, I'm working on it. Sorry it's been a while.

ErikSchierboom:
This is looking great! I think this will really help a lot of students. Thanks so much for doing this!

PR #728, opened 2018-11-16
Update gitignore

mrtank:
Add a fancier .gitignore file.


.gitignore
```diff
@@ -1,18 +1,259 @@
...
+# Paket dependency manager
+.paket/paket.exe
+
+# FAKE - F# Make
+.fake/
+
 *.swp
 .DS_Store
 tmp
-bin/configlet
```
ErikSchierboom: Could you re-add this line?

.gitignore
```diff
@@ -1,18 +1,259 @@
...
-bin/configlet.exe
```
ErikSchierboom: Could you re-add this line?

.gitignore
mrtank: Wouldn't `[Bb]in/` rule include this too?

ErikSchierboom: It would, but there is a `fetch-configlet` file in the [bin directory](https://github.com/exercism/csharp/tree/master/bin) that we wan't to retain.

mrtank:
```suggestion
[Bb]in/*
![Bb]in/configlet
![Bb]in/configlet.exe
```

mrtank: you mean like that?

ErikSchierboom:
Thanks!

Issue #720, opened 2018-11-05
Calculating saddle-points should be a static method


ErikSchierboom:
It currently is implemented as an instance method.

mrtank:
running `dotnet run -e SaddlePoints` in `git\csharp\generators\` made a change on the test file, which is not visible. Did it regenerate the same file, or I configured something wrong?

ErikSchierboom:
Closed by #729.

PR #729, opened 2018-11-16
SaddlePoints: Make the exercise testing a static method instead of instance member

mrtank:
#720 fixing that.


mrtank:
running `dotnet run -e SaddlePoints` in `git\csharp\generators\` made a change on the test file, which is not visible. Did it regenerate the same file, or I configured something wrong?

generators/Exercises/Generators/SaddlePoints.cs
```diff
@@ -10,11 +10,10 @@ public class SaddlePoints : GeneratorExercise
     {
         protected override void UpdateTestMethod(TestMethod testMethod)
         {
-            testMethod.TestedMethodType = TestedMethodType.InstanceMethod;
+            testMethod.TestedMethodType = TestedMethodType.StaticMethod;
             testMethod.TestedMethod = "Calculate";
-            testMethod.ConstructorInputParameters = new[] { "matrix" };
+            testMethod.InputParameters = new[] {"matrix"};
```
ErikSchierboom: I think you should be able to remove this line. Running the generator again should have the same outcome.

generators/Exercises/Generators/SaddlePoints.cs
```diff
@@ -10,11 +10,10 @@ public class SaddlePoints : GeneratorExercise
     {
         protected override void UpdateTestMethod(TestMethod testMethod)
         {
-            testMethod.TestedMethodType = TestedMethodType.InstanceMethod;
+            testMethod.TestedMethodType = TestedMethodType.StaticMethod;
```
ErikSchierboom: I think you should be able to remove this line. Running the generator again should have the same outcome.

exercises/saddle-points/SaddlePoints.cs
```diff
@@ -4,11 +4,7 @@
 
 public class SaddlePoints
```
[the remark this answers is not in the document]
mrtank: Done that. Something is still fishy about the generator. When I run `dotnet run -e saddle-points` it regenerates on the output the same as I would run `dotnet run -e SaddlePoints` (SaddlePointsTests.cs). VS asks if I want to reload the file, because change happened. Tho nothing changes. Git doesn't show change neither.

ErikSchierboom:
Thanks for this! 🎉

PR #733, opened 2018-11-25
Update RobotName name test to be more restrictive

haus:
Previously the tests would not catch invalid robot names such as AB1234, or ABC123 because the regex did not have start or end anchors specified. This commit updates the regex so that those names are no longer accepted by the tests.


ErikSchierboom:
Great catch! Thanks for the PR. 🎉

PR #737, opened 2018-12-17
734 use decimal for money

AndrewYHuang:
Updated BankAccount, BookStore and Ledger exercises to use decimal for money instead of float or double.
Updated RenderNumber and Render with decimal rendering.
Updated BookStore test generator to use decimal literals for BookStore tests.

### DIFF
--- a/exercises/bank-account/BankAccount.cs
+++ b/exercises/bank-account/BankAccount.cs
@@ -12,7 +12,7 @@ public class BankAccount
         throw new NotImplementedException("You need to implement this function.");
     }
 
-    public float Balance
+    public decimal Balance
     {
         get
         {
@@ -20,7 +20,7 @@ public class BankAccount
         }
     }
 
-    public void UpdateBalance(float change)
+    public void UpdateBalance(decimal change)
     {
         throw new NotImplementedException("You need to implement this function.");
     }
--- a/exercises/book-store/BookStore.cs
+++ b/exercises/book-store/BookStore.cs
@@ -3,7 +3,7 @@ using System.Collections.Generic;
 
 public static class BookStore
 {
-    public static double Total(IEnumerable<int> books)
+    public static decimal Total(IEnumerable<int> books)
     {
         throw new NotImplementedException("You need to implement this function.");
     }
--- a/exercises/ledger/Ledger.cs
+++ b/exercises/ledger/Ledger.cs
@@ -5,7 +5,7 @@ using System.Linq;
 
 public class LedgerEntry
 {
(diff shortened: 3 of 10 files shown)


generators/Output/Rendering/RenderNumber.cs
```diff
@@ -5,6 +5,7 @@ namespace Exercism.CSharp.Output.Rendering
 {
     public partial class Render
     {
+        public string Decimal(decimal dec) => $"{dec.ToString(CultureInfo.InvariantCulture)}m";
```
ErikSchierboom: Could you add a blank line after this method?

exercises/ledger/Ledger.cs
```diff
@@ -5,7 +5,7 @@
 
 public class LedgerEntry
 {
-   public LedgerEntry(DateTime date, string desc, float chg)
+   public LedgerEntry(DateTime date, string desc, decimal chg)
```
ErikSchierboom: Nice! Could you also update the `Example.cs` implementation (which is in the same directory)?

exercises/book-store/BookStore.cs
```diff
@@ -3,7 +3,7 @@
 
 public static class BookStore
 {
-    public static double Total(IEnumerable<int> books)
+    public static decimal Total(IEnumerable<int> books)
```
ErikSchierboom: Nice! Could you also update the `Example.cs` implementation (which is in the same directory)?

exercises/bank-account/BankAccount.cs
```diff
@@ -12,15 +12,15 @@ public void Close()
         throw new NotImplementedException("You need to implement this function.");
     }
 
-    public float Balance
+    public decimal Balance
```
ErikSchierboom: Nice! Could you also update the `Example.cs` implementation (which is in the same directory)?

AndrewYHuang:
Regarding lines 942-971 of the ci log, does the example implementation need updating? c# really doesn't like doing equals(decimal, float)

exercises/bank-account/BankAccount.cs
AndrewYHuang: Ok, sure! Didn't see that

ErikSchierboom:
> does the example implementation need updating

It does indeed.

AndrewYHuang:
Made all the requested changes: Updated examples and added that blank line to RenderNumber

ErikSchierboom:
Merged! This is great work. Thanks a lot for doing this. 🎉

PR #746, opened 2019-01-28
Test that names are unique; test that the name is correct after reset

vgrigoriu:
I added a couple of tests.

The first one verifies that the name is correct after reset. The existing tests can pass if the name is `""` after reset.

The second verifies that names do not repeat in 10,000 iterations (in my tests I always got collisions after no more than 2000).

### DIFF
--- a/exercises/robot-name/RobotNameTest.cs
+++ b/exercises/robot-name/RobotNameTest.cs
@@ -45,7 +45,7 @@ public class RobotNameTest
         var names = new HashSet<string>();
         for (int i = 0; i < 10_000; i++) {
             var robot = new Robot();
-            Assert.DoesNotContain(names, name => name == robot.Name);
+            Assert.DoesNotContain(robot.Name, names);
             names.Add(robot.Name);
         }
     }
--- a/exercises/robot-name/RobotNameTest.cs
+++ b/exercises/robot-name/RobotNameTest.cs
@@ -33,18 +33,20 @@ public class RobotNameTest
     }
 
     [Fact(Skip = "Remove to run test")]
-    public void After_reset_the_name_is_valid() {
+    public void After_reset_the_name_is_valid()
+    {
         robot.Reset();
         Assert.Matches(@"^[A-Z]{2}\d{3}$", robot.Name);
     }
 
     [Fact(Skip = "Remove to run test")]
-    public void Robot_names_are_unique() {
+    public void Robot_names_are_unique()
+    {
         var names = new HashSet<string>();
         for (int i = 0; i < 10_000; i++) {
-            Assert.False(names.Contains(robot.Name), $"Name {robot.Name} was already used ({i})");
+            var robot = new Robot();
+            Assert.DoesNotContain(names, name => name == robot.Name);
             names.Add(robot.Name);
-            robot.Reset();
         }
     }
 }
--- a/exercises/robot-name/RobotNameTest.cs
+++ b/exercises/robot-name/RobotNameTest.cs
@@ -1,3 +1,4 @@
+using System.Collections.Generic;
 using Xunit;
 
 public class RobotNameTest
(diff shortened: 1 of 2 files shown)


exercises/robot-name/RobotNameTest.cs
```diff
@@ -30,4 +31,20 @@ public void Can_reset_the_name()
         robot.Reset();
         Assert.NotEqual(originalName, robot.Name);
     }
+
+    [Fact(Skip = "Remove to run test")]
+    public void After_reset_the_name_is_valid() {
```
ErikSchierboom: Could you put the opening curly brace on the next line?

exercises/robot-name/RobotNameTest.cs
```diff
@@ -30,4 +31,20 @@ public void Can_reset_the_name()
...
+        robot.Reset();
+        Assert.Matches(@"^[A-Z]{2}\d{3}$", robot.Name);
+    }
+
+    [Fact(Skip = "Remove to run test")]
+    public void Robot_names_are_unique() {
```
ErikSchierboom: Could you put the opening curly brace on the next line?

exercises/robot-name/RobotNameTest.cs
```diff
@@ -30,4 +31,20 @@ public void Can_reset_the_name()
...
+        var names = new HashSet<string>();
+        for (int i = 0; i < 10_000; i++) {
+            Assert.False(names.Contains(robot.Name), $"Name {robot.Name} was already used ({i})");
+            names.Add(robot.Name);
+            robot.Reset();
```
ErikSchierboom: Instead of doing a reset, I think I would prefer to create a new robot in each iteration and then checking its name. This is also more in line with the test case's name.

exercises/robot-name/RobotNameTest.cs
ErikSchierboom: The existing [example implementation](https://github.com/exercism/csharp/blob/master/exercises/robot-name/Example.cs) fails this new test: https://travis-ci.org/exercism/csharp/builds/485576414?utm_source=github_status&utm_medium=notification Could you modify the example implementation to have it pass this test?

exercises/robot-name/RobotNameTest.cs
```diff
@@ -30,4 +31,20 @@ public void Can_reset_the_name()
...
+    public void After_reset_the_name_is_valid() {
+        robot.Reset();
+        Assert.Matches(@"^[A-Z]{2}\d{3}$", robot.Name);
+    }
+
+    [Fact(Skip = "Remove to run test")]
+    public void Robot_names_are_unique() {
+        var names = new HashSet<string>();
+        for (int i = 0; i < 10_000; i++) {
+            Assert.False(names.Contains(robot.Name), $"Name {robot.Name} was already used ({i})");
```
ErikSchierboom: I think there is an `Assert.DoesNotContain` method that you could use here. Note that we don't use an explicit assert message anywhere else in the test suites, so I wouldn't mind removing it here (especially as `Assert.DoesNotContain` will be quite descriptive).

vgrigoriu:
Thanks for the review!

exercises/robot-name/RobotNameTest.cs
vgrigoriu: I spend most of my time now in Java :-)

ErikSchierboom: :)

exercises/robot-name/RobotNameTest.cs
```diff
@@ -30,4 +31,22 @@ public void Can_reset_the_name()
...
+        Assert.Matches(@"^[A-Z]{2}\d{3}$", robot.Name);
+    }
+
+    [Fact(Skip = "Remove to run test")]
+    public void Robot_names_are_unique()
+    {
+        var names = new HashSet<string>();
+        for (int i = 0; i < 10_000; i++) {
+            var robot = new Robot();
+            Assert.DoesNotContain(names, name => name == robot.Name);
```
ErikSchierboom:
```suggestion
            Assert.DoesNotContain(name, names);
```

Reversing the argument allows you to get rid of the lambda.

ErikSchierboom:
Thanks for doing this!

PR #747, opened 2019-01-30
Fix #742

### LINKED ISSUE #742: clock: test suite should verify immutability
Some solutions have the `Add` and `Subtract`methods modify the current instance. This is not the goal of the exercise. We should add tests to verify this.

vgrigoriu:


ErikSchierboom:
Merged. Thanks for doing this! 🎉

PR #760, opened 2019-02-14
generators: fix wordy generator

### LINKED ISSUE #748: Fix wordy example to pass with the latest test data
e.g. the "just a number" test fails

To repro:
- go to the generators folder
- run `dotnet run -e wordy`
- go to the root folder
- run `./build.sh --exercise=wordy` or `.\build.ps1 -Exercise wordy`

ErikSchierboom:
Closes #748

### DIFF
--- a/exercises/wordy/Example.cs
+++ b/exercises/wordy/Example.cs
@@ -1,51 +1,41 @@
 using System;
-using System.Collections.Generic;
 using System.Text.RegularExpressions;
 
 public static class Wordy
 {
-    private static readonly Regex VariableAndOperatorGroupsRegex =
-        new Regex(string.Format(@"{0} {1} {0}\s*{1}*\s*{0}*", @"(-?\d+)", "(plus|minus|multiplied by|divided by)"));
-    private static readonly Dictionary<string, Func<int, int, int>> Operations = new Dictionary<string, Func<int, int, int>>
+    private static readonly Regex EquationRegex = new Regex(
+        @"^What is (?<left>-?\d+)(?<operations> (?<operand>plus|minus|multiplied by|divided by) (?<right>-?\d+))*\?$", RegexOptions.Compiled);
+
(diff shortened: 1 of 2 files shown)


exercises/wordy/Example.cs
```diff
@@ -1,51 +1,49 @@
...
 
     public static int Answer(string question)
     {
-        var parsedProblem = ParseProblem(question);
-        var firstPass = Operations[parsedProblem.Operation1](parsedProblem.X, parsedProblem.Y);
-        if (parsedProblem.Operation2.Length == 0)
-            return firstPass;
-        return Operations[parsedProblem.Operation2](firstPass, parsedProblem.Z);
+        var match = Parse(question);
+        if (match is null)
```
robkeim: I believe `Match` always returns a value but you should check `if (!match.Success)` to determine if there was nothing to match:

> You can determine whether the regular expression pattern has been found in the input string by checking the value of the returned Match object's Success property.

https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regex.match?view=netframework-4.7.2

exercises/wordy/Example.cs
```diff
@@ -1,51 +1,49 @@
...
-        public string Operation2;
-        public int Z;
+        switch (operand)
+        {
+            case "plus": return left + right;
+            case "minus": return left - right;
+            case "multiplied by": return left * right;
+            case "divided by": return left / right;
+            default: throw new ArgumentException();
+
```
robkeim: Super minor: unneeded newline here

exercises/wordy/Example.cs
ErikSchierboom: You are right. I've removed the superfluous check.

robkeim:
Changes look good to me, this can be merged as soon as the CI completes.

ErikSchierboom:
Merged. Thanks for the review!

PR #769, opened 2019-02-22
rational-numbers: remove untested methods

### LINKED ISSUE #764: rational-numbers: remove non-operator methods for Add/Sub/Div/Mul
The rational-numbers [test suite](https://github.com/exercism/csharp/blob/master/exercises/rational-numbers/RationalNumbersTest.cs) tests the addition/subtraction/division/multiplication of rational numbers through the use of operators. However, the [stub file](https://github.com/exercism/csharp/blob/master/exercises/rational-numbers/RationalNumbers.cs) includes explicit `Add/Sub/Div/Mul` methods, which makes it confusing for the user. We should remove these methods, and rely on the operators.

ErikSchierboom:
Closes #764

### DIFF
--- a/exercises/rational-numbers/Example.cs
+++ b/exercises/rational-numbers/Example.cs
@@ -22,45 +22,25 @@ public struct RationalNumber
     public int Numerator { get; }
     public int Denominator { get; }
 
-    public RationalNumber Add(RationalNumber r)
-    {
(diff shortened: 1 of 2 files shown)


robkeim:
Thanks @ErikSchierboom 🎉

PR #775, opened 2019-02-27
Auto-generate dependabot's config.yml

robkeim:
I'm not a Powershell wizard, so if you have any suggestions on how to improve the code I'm definitely open to them!

### DIFF
--- a/update-dependabot.ps1
+++ b/update-dependabot.ps1
@@ -1,10 +1,8 @@
-$outputPath = "./.dependabot/config.yml"
+$output = [System.Text.StringBuilder]::new()
 
-Remove-Item $outputPath
-
-Add-Content $outputPath "version: 1"
-Add-Content $outputPath ""
-Add-Content $outputPath "update_configs:"
+[void]$output.AppendLine("version: 1")
+[void]$output.AppendLine("")
+[void]$output.AppendLine("update_configs:")
 
 $paths = Get-ChildItem -Recurse -Path *.csproj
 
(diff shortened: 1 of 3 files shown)


update-dependabot.ps1
```diff
@@ -0,0 +1,24 @@
+$outputPath = "./.dependabot/config.yml"
+
+Remove-Item $outputPath
+
+Add-Content $outputPath "version: 1"
```
ErikSchierboom: An alternative would be to build up a string, and then replacing the contents at once. This will open and close the file for each call to `Add-Content`. Note that I don't really mind this a lot, as the script is not run often.

robkeim: You're totally right, I went ahead and fixed that. The script runs in only a few seconds so it's pretty quick, but avoiding needless work is always a good idea.

robkeim:
@ErikSchierboom I updated the exercise template and script, but unfortunately I don't have a linux/unix machine that I could write/test the script on :(

ErikSchierboom:
Great work! I've merged and created an [issue](https://github.com/exercism/csharp/issues/776) for the unix alternative.

ErikSchierboom:
@robkeim I think I've seen a small issue with the current script. It also includes the projects in the `/build/` directory :) Could you perhaps make sure that those projects are excluded?

robkeim:
@ErikSchierboom good catch. Fix is here: https://github.com/exercism/csharp/pull/805

robkeim:
Sorry for all of the noise in terms of the issues created :(

ErikSchierboom:
> Sorry for all of the noise in terms of the issues created :(

No worries. We're just figuring this out as we go :)

PR #805, opened 2019-02-27
Ignore build directory for dependabot

robkeim:


update-dependabot.ps1
```diff
@@ -13,10 +13,15 @@ foreach ($path in $paths)
 	$relativePath = $relativePath.Replace(".\", "/")
 	$relativePath = $relativePath.Replace("\", "/")
 	
+	if ($relativePath.StartsWith("/build/"))
```
ErikSchierboom: I think a slightly more elegant way would be to use the `-Exclude` parameter of the `Get-ChildItem` method.

robkeim: Ok, so it looks like the `-Exclude` only includes the leaf and not the directory. I could send it through an additional filter though like this:
`$paths = Get-ChildItem -Recurse -Path *.csproj | ? { $_.FullName -inotmatch '\\build\\' }`

Here's a StackOverflow discussing this:
https://stackoverflow.com/a/35993562

Do you think the piping solution is cleaner?

ErikSchierboom: Ah dammit, I sometimes get so disappointed with PowerShell :( I think the current solution is more readable then. Merged!

PR #984, opened 2019-02-27
Manual test for automerge of dependencies

robkeim:
If this works then I'll update the script and add this to all of the projects. I just didn't want to create another wave of spam if something isn't right with this configuration.

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -4,6 +4,10 @@ update_configs:
     - package_manager: "dotnet:nuget"
       directory: "/exercises/accumulate"
       update_schedule: "live"
+      automerged_updates:
+        - match:
+            dependency_type: "all"
+            update_type: "semver:minor"
 
     - package_manager: "dotnet:nuget"
       directory: "/exercises/acronym"


.dependabot/config.yml
```diff
@@ -4,6 +4,10 @@ update_configs:
     - package_manager: "dotnet:nuget"
       directory: "/exercises/accumulate"
       update_schedule: "live"
+      automerged_updates:
+        - match:
+            dependency_type: "all"
+            update_type: "semver:minor"
```
ErikSchierboom: Does this also wait for the CI to become green?

robkeim: Yes

https://dependabot.com/docs/config-file/#automerged_updates

> For all of the options below Dependabot will wait until all your status checks pass before merging.

ErikSchierboom:
@robkeim I'm really hoping work will be done on [this issue](https://github.com/dependabot/feedback/issues/5).

robkeim:
Yeah that would definitely be nice, because each time xunit gets updated we're going to get ~100 PRs generated which can create a lot of noise. I have to admit I've never used this on a project with so many csproj's before :)

ErikSchierboom:
Right, merged! Let's see what happens :)

ErikSchierboom:
This is a rather unusual project with some many individual project files. So it would be a most welcome feature.

greysteil:
Hey, if you update the config file to have a single update config, pointing at the `/exercises` directory, then Dependabot will pick up the `Exercises.sln` file there and create PRs that group together updating all of the individual projects, rather than individual PRs for each project.

(We need to make it clearer that that's the case, particularly for folks using config files.)

ErikSchierboom:
@greysteil That's a great hint! @robkeim, did you see this?

ErikSchierboom:
@robkeim When we have this work properly, we can then try and do the same for F#, which should then result in far less PR's :)

ErikSchierboom:
@greysteil Would the config look something like this then:

```yaml
version: 1

update_configs:
    - package_manager: "dotnet:nuget"
      directory: "/exercises"
      update_schedule: "live"
      automerged_updates:
        - match:
            dependency_type: "all"
            update_type: "semver:minor"

    - package_manager: "dotnet:nuget"
      directory: "/generators"
      update_schedule: "live"
      automerged_updates:
        - match:
            dependency_type: "all"
            update_type: "semver:minor"
```

greysteil:
Yes, that should work!

robkeim:
Thanks @greysteil for the tip!

PR #1243, opened 2019-03-01
structure: move circular-buffer up

### LINKED ISSUE #767: track structure: move circular-buffer up in core exercise order
The circular-buffer exercise is a bit too simple for an exercise that relatively late in the track. We should be moving it earlier in the track.

ErikSchierboom:
The canonical solution to the circular-buffer exercise, is to use a queue. This makes it quite simple. I do like to keep it a core exercise though, as it can teach students about different data structure (e.g. `Queue<T>`) and introduces generic classes. As such, I've moved it up a bit in the core ordering.

Closes #767

### DIFF
--- a/config.json
+++ b/config.json
@@ -152,6 +152,17 @@
       ]
     },
     {
+      "slug": "circular-buffer",
+      "uuid": "657ac368-9b17-4f87-b815-decfe2bc0b5d",
+      "core": true,
+      "unlocked_by": null,
+      "difficulty": 4,
+      "topics": [
+        "classes",
+        "queues"
+      ]
+    },
+    {
       "slug": "clock",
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -151,6 +151,17 @@
         "transforming"
       ]
     },
+    {
+      "slug": "circular-buffer",
+      "uuid": "657ac368-9b17-4f87-b815-decfe2bc0b5d",
+      "core": true,
+      "unlocked_by": null,
+      "difficulty": 4,
```
robkeim: @ErikSchierboom was the difficulty change intentional here? (It was 5 previously)

ErikSchierboom: Yes it was intentional, based on the students submissions I see.

robkeim:
Looks good to me, thanks! 🎉

PR #1245, opened 2019-03-08
docs: remove submitting incomplete solutions from README

ErikSchierboom:
This PR removes the unhelpful "You can submit incomplete solutions" part of the README. While doing this, I found that the Git patch solution did not work anymore. I also found it to interfere with updating a single exercise, which is why I've removed said functionality in this PR. We lose a little bit in the process (function is no longer renamed to method), but I'm okay with that.


robkeim:
It's too bad the patch solution doesn't work anymore, but thanks for doing this! 👍

PR #1254, opened 2019-03-21
general: add .editorconfig file to exercises

### LINKED ISSUE #1252: Consider adding .editorconfig file to exercises
Many of our students will be new to the .NET platform. As such, they won't be familiar with the usual conventions applied to C# code. It is possible to define the conventions in an `.editorconfig` file (see e.g. https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-naming-conventions?view=vs-2017#example). With the next version of Roslyn (which will be released somewhere in the near future), it will even be possible for the compiler to throw an error if one of those rules doesn't match. Maybe we should consider adding an `.editorconfig` file that defines the default .NET conventions?

Note: if we add `.editorconfig` files, we could also consider supporting the [dotnet format](https://github.com/dotnet/roslyn/tree/master/src/Tools/dotnet-format) tool, to allow users to automatically format the code.

ErikSchierboom:
Closes #1252


.editorconfig
```diff
@@ -1,19 +1,143 @@
...
+dotnet_naming_rule.constant_fields_should_be_pascal_case.style = pascal_case_style
+dotnet_naming_symbols.constant_fields.applicable_kinds = field
+dotnet_naming_symbols.constant_fields.applicable_accessibilities = *
+dotnet_naming_symbols.constant_fields.required_modifiers = const
+
+###############################
+# C# Code Style Rules         #
+###############################
+
+[*.cs]
```
robkeim: This is unneeded right as this is a continuation of the above section?

robkeim:
My one concern with this PR is maintenance. If we want to update the editorconfig we have to remember to update in 125 places. Also for new exercises being added we need to remember to add this file as well.

Does it make sense to have a script that takes the root editorconfig and copies it to all of the exercise directories?

exercises/accumulate/.editorconfig
```diff
@@ -0,0 +1,143 @@
+###############################
+# Core EditorConfig Options   #
+###############################
+
+; This file is for unifying the coding style for different editors and IDEs.
+; More information at:
+; https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference?view=vs-2017
+; https://docs.microsoft.com/en-us/visualstudio/ide/create-portable-custom-editor-options?view=vs-2017
+
+root = true
```
robkeim: Should each of the exercises mark themselves as a root config?

.editorconfig
ErikSchierboom: That is correct! Will remove.

exercises/accumulate/.editorconfig
ErikSchierboom: I think they do, as otherwise the editorconfig format specifies that it will keep on looking for .editorconfig files in the directory's parents, until one is found with `root = true`. In practice, omitting the `root = true` in the exercise `.editorconfig` means that if two/three directories up one would have an `.editorconfig` that adds an unwanted or new formatting feature. My preference would be to include the `root = true`, to prevent situations where the formatting is somehow incorrectly applied due to some parent `.editorconfig` file.

ErikSchierboom:
> My one concern with this PR is maintenance. If we want to update the editorconfig we have to remember to update in 125 places. Also for new exercises being added we need to remember to add this file as well.
> 
> Does it make sense to have a script that takes the root editorconfig and copies it to all of the exercise directories?

I was a bit worried that we might end up with too many shell scripts, but that might not be necessary. I'll add such scripts.

exercises/accumulate/.editorconfig
Mehni: Naive question: Does that mean you could put one `.editorconfig` with `root = true` in the `Exercism/csharp` directory, instead of one per `Exercism/csharp/exercise`?

If doing that works without side-effects, it would also solve the maintenance aspect.

robkeim: @Mehni yes that's correct for people that check out the entire solution. The problem is that when students download a specific exercise they're not going to receive the everything in the `Exercism/csharp` directory on the `Exercism/csharp/EXERCISE` directory.

ErikSchierboom: Unfortunately, there is no functionality to have track specific, cross-exercise files.

ErikSchierboom:
@robkeim I've gone ahead and merged, as it was open for four days and I applied your review comments. If that is not okay, let me know, and I'll revert.

robkeim:
@ErikSchierboom sorry I missed the update you made to this branch. Thanks for making the script to ensure these are easy to update in the future.

PR #1261, opened 2019-03-28
styling: only apply .editorconfig settings to implementation file

### LINKED ISSUE #1259: Exclude test file from .editorconfig file
See https://github.com/exercism/csharp/issues/1252#issuecomment-477145822

ErikSchierboom:
This is to prevent suggestions/warnings for the test file.

Closes #1259.


ErikSchierboom:
@SleeplessByte I've gone the reverse route, where instead of excluding the test file, I only apply the rules to the implementation file. This should work, right?

SleeplessByte:
@ErikSchierboom your solution is actually _better_.

SleeplessByte:
I also suggest you open an issue (if not already) which updates the tests to match these suggestions, and gradually you can add the test files to those configs. That way contributing adheres to the same rules.

ErikSchierboom:
@SleeplessByte Good idea. I've [created an issue](https://github.com/exercism/csharp/issues/1262).

ErikSchierboom:
Merged. Thanks for the reviews!

PR #1264, opened 2019-03-29
Improve performance on Robot_names_are_unique

### LINKED ISSUE #1263: Improve performance on Robot_names_are_unique
Due to a [weird implementation](https://github.com/xunit/xunit/issues/1543) of Assert.DoesNotContain in XUnit, `Robot_names_are_unique` (in RobotNameTest.cs) takes much longer than it needs to. 

`Assert.False(names.Contains(robot.Name));` causes an xunit analyzer warning, but if we combine that with the next line we can get even faster behavior and avoid the analyzer warning:
```
Assert.True(names.Add(robot.Name));
```
This runs in 1.7 seconds instead of 7.3 seconds.

j2jensen:
Assert.DoesNotContain() has performance issues.
Resolves #1263

### DIFF
--- a/exercises/robot-name/RobotNameTest.cs
+++ b/exercises/robot-name/RobotNameTest.cs
@@ -45,8 +45,7 @@ public class RobotNameTest
         var names = new HashSet<string>();
         for (int i = 0; i < 10_000; i++) {
             var robot = new Robot();
-            Assert.DoesNotContain(robot.Name, names);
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
Looking good! There was a flaky test, so I've restarted the build.

ErikSchierboom:
Thanks @j2jensen!

PR #1269, opened 2019-04-03
two-fer: don't specify default null parameter

### LINKED ISSUE #1268: Remove default `null` value in TwoFer stub
The [two-fer stub implementation](https://github.com/exercism/csharp/blob/master/exercises/two-fer/Example.cs) defines the `input` parameter as having a default value of `null`. This is problematic, as we really don't want the second exercise in the track teach about nullability (which is a hard concept). It also means that we are already steering people towards a specific solution, namely one where the default value is `null`. As such, many students never even consider changing the default value, which is actually the most elegant solution (`input = "you"`). We thus lose the option to teach students about default values, which really is a missed opportunity.

Side note: we could also rename `input` to `name` while were at it (in a separate commit).

ErikSchierboom:
Closes #1268

### DIFF
--- a/exercises/two-fer/Example.cs
+++ b/exercises/two-fer/Example.cs
@@ -1,16 +1,5 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Text;
-
-  public static class TwoFer
-  {
-      public static string Name(string input = null)
-      {
-          if(string.IsNullOrWhiteSpace(input))
-          {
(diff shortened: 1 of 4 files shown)


robkeim:
Thanks @ErikSchierboom 🎉

ErikSchierboom:
Thanks for reviewing @robkeim!

PR #1272, opened 2019-04-10
build: remove shell scripts

### LINKED ISSUE #1255: Consider replacing shell scripts with PowerShell
Currently, we have two versions of our scripts:

- A Unix shell script
- A Windows command script 

Now that PowerShell is fully cross-platform (it works nicely on my Mac), we could consider writing all our scripts in PowerShell. This would reduce some duplication and would make it easier to add new scripts, as you only have to write one version. What do you think?

ErikSchierboom:
Closes #1255


ErikSchierboom:
We'll see if I broke the build :)

ErikSchierboom:
This PR does a couple of things:

1. It removes the `.sh` scripts. We now only have to maintain the PowerShell scripts.
1. The Travis integration has been simplified.
1. The build script now also has a CI build target, that does the linting.

robkeim:
Great work, thanks @ErikSchierboom!

PR #1275, opened 2019-04-13
scripts: allow passing argument to configlet to regenerate a single README

### LINKED ISSUE #757: Allow generating single README
The `update-docs.sh` and `update-docs.ps1` scripts allows one to re-generate the README files. They don't, however, allow for generating a single exercise. The scripts should be modified to take an `exercise` command-line parameter (see `build.ps1` and `build.sh` for an example), and if specified, should only update a single exercise.

### LINKED ISSUE #1246: Make update-docs work without having manually to checkout the problem-specifications repo
Currently, the update-docs scripts rely on the [problem-specifications repo](https://github.com/exercism/problem-specifications) to be checked out in `../problem-specifications`. This means that there is a prerequisite to getting the scripts to run, which is not ideal. What we should do is:

1. Checkout the problem-specifications repository from within the script itself (skip if already done).
1. Update the problem-specifications repository to the latest version.

ErikSchierboom:
Closes #757
Closes #1246

### DIFF
--- a/update-docs.ps1
+++ b/update-docs.ps1
@@ -1,5 +1,8 @@
 # Usage: .\update-docs.ps1. This will regenerate the docs for all exercises.
 # Usage: .\update-docs.ps1 -o <exercise>. This will regenerate the docs for the specified exercise.
 
+git submodule init
+git submodule update --remote
+
 .\bin\fetch-configlet
-.\bin\configlet generate . $args
+.\bin\configlet generate . -p problem-specifications $args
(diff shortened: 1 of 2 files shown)


update-docs.ps1
```diff
@@ -1,2 +1,8 @@
+# Usage: .\update-docs.ps1. This will regenerate the docs for all exercises.
+# Usage: .\update-docs.ps1 -o <exercise>. This will regenerate the docs for the specified exercise.
```
robkeim: Why `-o` out of curiosity? I would have expected `-e` for exercise.

ErikSchierboom: Well, that is how the underlying `configlet` binary expects its parameter.

In a future update, I plan to bring back the strongly-typed parameters, so you'll do `.\update-docs.ps1 -Exercise <exercise>` then.

PR #1276, opened 2019-04-15
bracket-push: rename to matching-brackets

SleeplessByte:
This rename was proposed in:
exercism/problem-specifications#1501

The rationale for the name is:

* to name the exercise by its story, not by what it potentially teaches
* to avoid unnecessarily biasing the solution space


ErikSchierboom:
Merged. Thanks @SleeplessByte !

PR #1283, opened 2019-05-02
Check the random number distibution

jdehaan:
In many solutions people don't implement exactly the described algorithm. They just generate a random number - uniformly - between 3 and 18 to make the checks pass. Some implementations I saw even missed to generate some of the numbers in the range due to an off by one. This test checks that the distribution fits with a +/- 20% tolerance vs the ideal values.

### DIFF
--- a/exercises/dnd-character/DndCharacterTest.cs
+++ b/exercises/dnd-character/DndCharacterTest.cs
@@ -140,4 +140,36 @@ public class DndCharacterTest
             Assert.Equal(sut.Charisma, sut.Charisma);
         }
     }
+    
+    [Fact(Skip = "Remove to run test")]
+    public void Random_ability_is_distributed_correctly()
+    {
+        var idealDistribution = new Dictionary<int, int>() {
+            [3] = 1,            [4] = 4,
+            [5] = 10,           [6] = 21,
+            [7] = 38,           [8] = 62,
+            [9] = 91,           [10] = 122,
+            [11] = 148,         [12] = 167,
+            [13] = 172,         [14] = 160,
+            [15] = 131,         [16] = 94,
+            [17] = 54,          [18] = 21
+        };
+        var distribution = new Dictionary<int, int>();
+        int times = 100;
+        for (var i = 3; i <= 18; i++)
+            distribution[i] = 0;
+        for (var i = 0; i < times * 6*6*6*6; i++)
+        {
+            var a = DndCharacter.Ability();
+            distribution[a] += 1;
+        }
+        int min(int ideal) {
+            return (int)(ideal * (times * 0.8));
+        }
+        int max(int ideal) {
+            return (int)(ideal * (times * 1.2));
+        }
+        foreach (var k in idealDistribution.Keys)
+            Assert.InRange(distribution[k], min(idealDistribution[k]), max(idealDistribution[k]));
+    }
 }
--- a/exercises/dnd-character/DndCharacterTest.cs
+++ b/exercises/dnd-character/DndCharacterTest.cs
@@ -161,7 +161,7 @@ public class DndCharacterTest
         for (var i = 0; i < times * 6*6*6*6; i++)
         {
             var a = DndCharacter.Ability();
-            distribution[a] += 1;
+            distribution[a]++;
         }
         int min(int ideal) {
             return (int)(ideal * (times * 0.8));
--- a/exercises/dnd-character/DndCharacterTest.cs
+++ b/exercises/dnd-character/DndCharacterTest.cs
@@ -140,7 +140,7 @@ public class DndCharacterTest
             Assert.Equal(sut.Charisma, sut.Charisma);
         }
     }
-    
+
     [Fact(Skip = "Remove to run test")]
     public void Random_ability_is_distributed_correctly()
     {
--- a/exercises/dnd-character/DndCharacterTest.cs
+++ b/exercises/dnd-character/DndCharacterTest.cs
@@ -154,7 +154,7 @@ public class DndCharacterTest
             [15] = 131,         [16] = 94,
             [17] = 54,          [18] = 21
         };
-        var distribution = new Dictionary<int, int>();
+        var actualDistribution = new Dictionary<int, int>();
         int times = 100;
         for (var i = 3; i <= 18; i++)
             distribution[i] = 0;
--- a/exercises/dnd-character/DndCharacterTest.cs
+++ b/exercises/dnd-character/DndCharacterTest.cs
@@ -156,16 +156,16 @@ public class DndCharacterTest
         };
         var actualDistribution = new Dictionary<int, int>();
         var times = 100;
-        const int POSSIBLE_COMBINATIONS_COUNT = 6*6*6*6; // 4d6
+        const int PossibleCombinationsCount = 6*6*6*6; // 4d6
         for (var i = 3; i <= 18; i++)
             actualDistribution[i] = 0;
-        for (var i = 0; i < times * POSSIBLE_COMBINATIONS_COUNT; i++)
+        for (var i = 0; i < times * PossibleCombinationsCount; i++)
         {
             var ability = DndCharacter.Ability();
             actualDistribution[ability]++;
         }
-        int min(int ideal) => (int)(ideal * (times * 0.8));
-        int max(int ideal) => (int)(ideal * (times * 1.2));
+        int min(int expected) => (int)(expected * (times * 0.8));
+        int max(int expected) => (int)(expected * (times * 1.2));
         foreach (var k in idealDistribution.Keys)
             Assert.InRange(actualDistribution[k], min(expectedDistribution[k]), max(expectedDistribution[k]));
     }
--- a/exercises/dnd-character/DndCharacterTest.cs
+++ b/exercises/dnd-character/DndCharacterTest.cs
@@ -144,18 +144,18 @@ public class DndCharacterTest
     [Fact(Skip = "Remove to run test")]
     public void Random_ability_is_distributed_correctly()
     {
-        var idealDistribution = new Dictionary<int, int>() {
-            [3] = 1,            [4] = 4,
-            [5] = 10,           [6] = 21,
-            [7] = 38,           [8] = 62,
-            [9] = 91,           [10] = 122,
-            [11] = 148,         [12] = 167,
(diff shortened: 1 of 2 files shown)


jdehaan:
If you want to know how I figured out the ideal numbers, as I am lame at probability I just generated all possible combinations and counted the hits with this code:

```csharp
        var combinations = new List<int[]>(6*6*6*6);
        for (int d1 = 1 ; d1 <= 6; d1++)
            for (int d2 = 1 ; d2 <= 6; d2++)
                for (int d3 = 1 ; d3 <= 6; d3++)
                    for (int d4 = 1 ; d4 <= 6; d4++)
                        combinations.Add(new int[] {d1,d2,d3,d4});
        System.Console.WriteLine($"# combinations = {combinations.Count}");
        var stats = combinations
            .GroupBy(x => x.Sum() - x.Min())
            .ToDictionary(g => g.Key, g => g.Count());
        
        foreach(var k in stats.Keys.OrderBy(x => x)) {
            System.Console.WriteLine($"{k,2}: {stats[k],4} -> {stats[k],4}/{6*6*6*6} = {stats[k]*100.0/(6*6*6*6)}");
        }
```

jdehaan:
There is an inherent problem with the test due to the randomness. There could be false negatives... A proper choice of the tolerance and number of runs should be safe enough but there is no guarantee. This is challenging but how else test the correct effect of the implementation? Not testing the distribution leaves too much room for taking some shortcuts and having this test leads people to think about important differences in the implementation, in dnd games this matters...

exercises/dnd-character/DndCharacterTest.cs
```diff
@@ -140,4 +140,36 @@ public void Each_ability_is_only_calculated_once()
...
         }
     }
-}
\ No newline at end of file
+    
+    [Fact(Skip = "Remove to run test")]
+    public void Random_ability_is_distributed_correctly()
+    {
+        var idealDistribution = new Dictionary<int, int>() {
+            [3] = 1,            [4] = 4,
```
ErikSchierboom: Any reason for the large number of spaces between the elements? Personally, I think having a little less spaces might be a bit easier to read:

```suggestion
            [3] = 1,        [4] = 4,
```

exercises/dnd-character/DndCharacterTest.cs
```diff
@@ -140,4 +140,36 @@ public void Each_ability_is_only_calculated_once()
             Assert.Equal(sut.Charisma, sut.Charisma);
         }
     }
-}
\ No newline at end of file
+    
+    [Fact(Skip = "Remove to run test")]
+    public void Random_ability_is_distributed_correctly()
+    {
+        var idealDistribution = new Dictionary<int, int>() {
```
ErikSchierboom: Should it be `idealDistribution` or `expectedDistribution`?

exercises/dnd-character/DndCharacterTest.cs
```diff
@@ -140,4 +140,36 @@ public void Each_ability_is_only_calculated_once()
...
+        var distribution = new Dictionary<int, int>();
+        int times = 100;
+        for (var i = 3; i <= 18; i++)
+            distribution[i] = 0;
+        for (var i = 0; i < times * 6*6*6*6; i++)
+        {
+            var a = DndCharacter.Ability();
+            distribution[a] += 1;
+        }
+        int min(int ideal) {
```
ErikSchierboom: I think you could make this an expression bodied member : `int min(int ideal) => (int)(ideal * (times * 0.8));`

exercises/dnd-character/DndCharacterTest.cs
```diff
@@ -140,4 +140,36 @@ public void Each_ability_is_only_calculated_once()
...
+            return (int)(ideal * (times * 0.8));
+        }
+        int max(int ideal) {
```
ErikSchierboom: I think you could make this an expression bodied member : `int max(int ideal) => (int)(ideal * (times * 1.2));`

exercises/dnd-character/DndCharacterTest.cs
```diff
@@ -140,4 +140,36 @@ public void Each_ability_is_only_calculated_once()
...
+            [5] = 10,           [6] = 21,
+            [7] = 38,           [8] = 62,
+            [9] = 91,           [10] = 122,
+            [11] = 148,         [12] = 167,
+            [13] = 172,         [14] = 160,
+            [15] = 131,         [16] = 94,
+            [17] = 54,          [18] = 21
+        };
+        var distribution = new Dictionary<int, int>();
+        int times = 100;
```
ErikSchierboom: All the other tests files use `var` everywhere:

```suggestion
        var times = 100;
```

exercises/dnd-character/DndCharacterTest.cs
```diff
@@ -140,4 +140,36 @@ public void Each_ability_is_only_calculated_once()
...
+        for (var i = 3; i <= 18; i++)
+            distribution[i] = 0;
+        for (var i = 0; i < times * 6*6*6*6; i++)
+        {
+            var a = DndCharacter.Ability();
+            distribution[a] += 1;
```
ErikSchierboom:
```suggestion
            distribution[a]++;
```

exercises/dnd-character/DndCharacterTest.cs
```diff
@@ -140,4 +140,36 @@ public void Each_ability_is_only_calculated_once()
...
+            [15] = 131,         [16] = 94,
+            [17] = 54,          [18] = 21
+        };
+        var distribution = new Dictionary<int, int>();
+        int times = 100;
+        for (var i = 3; i <= 18; i++)
+            distribution[i] = 0;
+        for (var i = 0; i < times * 6*6*6*6; i++)
+        {
+            var a = DndCharacter.Ability();
```
ErikSchierboom:
```suggestion
            var score = DndCharacter.Ability();
```

or 

```suggestion
            var ability = DndCharacter.Ability();
```

exercises/dnd-character/DndCharacterTest.cs
```diff
@@ -140,4 +140,36 @@ public void Each_ability_is_only_calculated_once()
...
+            [3] = 1,            [4] = 4,
+            [5] = 10,           [6] = 21,
+            [7] = 38,           [8] = 62,
+            [9] = 91,           [10] = 122,
+            [11] = 148,         [12] = 167,
+            [13] = 172,         [14] = 160,
+            [15] = 131,         [16] = 94,
+            [17] = 54,          [18] = 21
+        };
+        var distribution = new Dictionary<int, int>();
```
ErikSchierboom: Maybe to contract this with `idealDistribution`, name it:

```suggestion
        var actualDistribution = new Dictionary<int, int>();
```

exercises/dnd-character/DndCharacterTest.cs
```diff
@@ -140,4 +140,36 @@ public void Each_ability_is_only_calculated_once()
...
+        int times = 100;
+        for (var i = 3; i <= 18; i++)
+            distribution[i] = 0;
+        for (var i = 0; i < times * 6*6*6*6; i++)
```
ErikSchierboom: To me, it isn't immediately clear what the significance is of `6*6*6*6`. Perhaps introduce a well-named `const` for it?

exercises/dnd-character/DndCharacterTest.cs
jdehaan: Ok

jdehaan:
> You are currently directly editing the test file, however most of the tests (including this one), are automatically generated by our [test generators](https://github.com/exercism/csharp/blob/master/docs/GENERATORS.md). The source for this generator can be found [here](https://github.com/exercism/csharp/blob/master/generators/Exercises/Generators/DndCharacter.cs). Could you incorporate your change in the test generator?

Sure, that makes sense. But I need some time to get into it.

ErikSchierboom:
> Sure, that makes sense. But I need some time to get into it.

Sure! Let us know if you need any help.

jdehaan:
Thx for the review, I quickly cleaned up the current contents. I'll dig into the generator and transfer the code later into it and extend the PR.

exercises/dnd-character/DndCharacterTest.cs
```diff
@@ -140,4 +140,33 @@ public void Each_ability_is_only_calculated_once()
...
+            [7] = 38,       [8] = 62,
+            [9] = 91,       [10] = 122,
+            [11] = 148,     [12] = 167,
+            [13] = 172,     [14] = 160,
+            [15] = 131,     [16] = 94,
+            [17] = 54,      [18] = 21
+        };
+        var actualDistribution = new Dictionary<int, int>();
+        var times = 100;
+        const int POSSIBLE_COMBINATIONS_COUNT = 6*6*6*6; // 4d6
```
ErikSchierboom: Tiny nit:

```suggestion
        const int PossibleCombinationsCount = 6*6*6*6; // 4d6
```

jdehaan:
@ErikSchierboom I had a short look at the generators (I need to dig more into that) and saw that it might be easier to generate a set of assertions with the values calculated within the genrator. What is your opinion, should the resulting test be readable (like it is now) or might it be a bit cryptic? I would opt for the first option (but will make the generator "ugly" in the sense it'll contain much hard coded text)

I think this might be common trade-off between maintainability of the generator vs readability of the result... and therefore I am looking forward to getting some stronger opinion.

ErikSchierboom:
Definitely go for the readable test. Don't worry about the generator being less pretty or having hardcoded strings, this is an exceptional case after all.

jdehaan:
*I consider the contribution as ready*

@ErikSchierboom that was not too difficult :-) if the way I did it with a verbatim data addition to the generator.

I had some trouble running all the generators under Linux (Ubuntu 18.04), installed libcurl3 and could only run generators partly. I re-generated the content using `dotnet run -e DndCharacter` that fortunately worked well and produced the result I committed.

---

BTW: I tried to figure out how to contribute something to the canonical data but that kind of test does not yet seem to be part of the "DSL", the semantics used are constrained and I thing this would end up in testing magic ranges somehow... That is a challenging task to level up the canonical description, but would be a good thing for all language tracks.

ErikSchierboom:
Great work! 

> That is a challenging task to level up the canonical description, but would be a good thing for all language tracks.

Yeah, I agree. This is hard to do unfortunately.

PR #1290, opened 2019-06-22
two-fer: fix compilation issues

dmahnkopf:


exercises/two-fer/TwoFer.cs
```diff
@@ -1,6 +1,8 @@
-﻿public static class TwoFer
+﻿using System;
```
ErikSchierboom: Good catch!

exercises/two-fer/TwoFer.cs
```diff
@@ -1,6 +1,8 @@
...
+
+public static class TwoFer
 {
-    public static string Speak()
+    public static string Speak(string input = "")
```
ErikSchierboom: Could you revert this change please? It is a deliberate decision to not have the example code compile, as we want people to figure out how to solve the exercise themselves, without already leading them towards a solution. Previously, we use `string input = null` as the example implementation, but that made people not be aware that they could also just use a string. Furthermore, it also led them to automatically discard method overloading. So now having a compiling example file here is deliberate.

I would, however, be fine with adding a comment somewhere in this file stating that we know the code doesn't compile, and that it is up to the user to figure out how to fix it.

dmahnkopf: O, I see.
What would You think of `// In order to get the tests running, first you need to make sure the Speak method can be run either without any or by passing one string argument.` just above the method?

ErikSchierboom: I like that a lot!

exercises/two-fer/TwoFer.cs
```diff
@@ -1,5 +1,8 @@
-﻿public static class TwoFer
+﻿using System;
+
+public static class TwoFer
 {
+    // In order to get the tests running, first you need to make sure the Speak method can be run both without any and also by passing one string argument.
```
ErikSchierboom:
```suggestion
    // In order to get the tests running, first you need to make sure the Speak method 
    // can be called both without any arguments and also by passing one string argument.
```

Issue #1299, opened 2019-07-06
AllergiesTest.cs giving error when run.


Mascasc:
I'm trying to do the C# exercism exercise Allergies.
When I run the usual `dotnet test`, I get the following error:
```
error CS0111: Type 'AllergiesTest' already defines a member called 'Not_allergic_to_anything' with the same parameter types 
```
There is one of these errors for every test in the AllergiesTest.cs file.
So far this is the only exercise that has given this error.

Edit: I figured out the problem.  There are several tests that all share the name 'Not_allergic_to_anything', as well as 'Allergic_to_everything'.  The test suite also expects a method called `AllergicTo`, while the provided template names this method `IsAllergicTo`.  When I changed the method names the test suite ran without issues.

ErikSchierboom:
Yep, this is an error on our part. We'll fix it!

PR #1300, opened 2019-07-07
Ensure PowerShell errors are propagated

ErikSchierboom:
When issue #1299 was created, I found out that there was something wrong with our build pipeline. The [CI run](https://travis-ci.org/exercism/csharp/builds/554366818?utm_source=github_status&utm_medium=notification) had an error, but Travis still thought that it was a succesful build. Apparently, one has to be explicit about returning the error code of called scripts, which is what this PR adds.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -1,2 +1,4 @@
 dotnet tool install -g Cake.Tool --version 0.33.0
 dotnet cake build.cake $args
+
+exit $LastExitCode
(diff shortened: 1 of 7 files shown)


ErikSchierboom:
Well, apparently, the build has been not detecting any errors for a while. I've also fixed the dnd-character, and will fix the other issues later.

PR #1301, opened 2019-07-07
Fix allergies test suite

### LINKED ISSUE #1299: AllergiesTest.cs giving error when run.

ErikSchierboom:
Closes #1299

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -255,6 +255,7 @@ packages/
 
 .vscode/
 .idea/
+.ionide/
 
 *.un~
 
--- a/exercises/allergies/AllergiesTest.cs
+++ b/exercises/allergies/AllergiesTest.cs
@@ -5,283 +5,283 @@ using Xunit;
 public class AllergiesTest
 {
     [Fact]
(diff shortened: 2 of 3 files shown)


ErikSchierboom:
@robkeim I've gone ahead and merged this, as this is broken code that gets downloaded for students. If you have any questions or objections, let me know!

robkeim:
Looks good @ErikSchierboom! Thanks for jumping on this and fixing it so quickly.

PR #1310, opened 2019-08-17
WIP Update prime factors to use longs

### LINKED ISSUE #1309: Inherent flaw in PrimeFactor exercise
I understand the motivation behind implementing a test which exceeds the 32-bit range:

```
    [Fact]
    public void Factors_include_a_large_prime()
    {
        Assert.Equal(new[] { 11, 9539, 894119 }, PrimeFactors.Factors(93819012551));
    }
```

but I think it is misguided at best if not encouraging questionable practice to then have the solution template mixing 32- and 64-bit (`public static int[] Factors(long number)`) and the tests all testing against an `int` instead of `long`. If the student is meant to implement support for 64-bit numbers it should also test against 64-bit values.

robkeim:
This change causes all of the ints to be treated as longs throughout the rest of the exercises. I left the other exercises out of this WIP PR so it makes it easier to see what changes I made.

@ErikSchierboom do you have an idea what I'm doing wrong?

Fixes #1309

### DIFF
--- a/exercises/prime-factors/Example.cs
+++ b/exercises/prime-factors/Example.cs
@@ -2,10 +2,10 @@ using System.Collections.Generic;
 
 public class PrimeFactors
 {
-    public static int[] Factors(long number)
+    public static long[] Factors(long number)
     {
-        var primes = new List<int>();
-        int divisor = 2;
+        var primes = new List<long>();
+        long divisor = 2;
         while (number > 1)
         {
             while (number % divisor == 0)
--- a/exercises/prime-factors/PrimeFactors.cs
+++ b/exercises/prime-factors/PrimeFactors.cs
@@ -2,7 +2,7 @@ using System;
 
 public static class PrimeFactors
 {
-    public static int[] Factors(long number)
+    public static long[] Factors(long number)
     {
         throw new NotImplementedException();
     }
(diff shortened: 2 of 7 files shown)


ErikSchierboom:
@robkeim You're not really doing anything wrong, it's just that JSON.net by default will deserialize numbers to a `long`, as some numbers might not fit into a `int`. 

Now what to do about this. There are several options I feel:

1. At the places where a long is needed, we render it as a long string instead of relying on the automatic rendering (which has an effect on all exercises).
1. We could plug a custom JsonConverter into our deserialization, which correctly identifies if a value should actually be converted to an integer instead of a long. I have some experience with this, so I can help here.

I think the second one is a little bit nicer, but also a bit more work. I'm happy to help here though. What do you think?

robkeim:
I'm also in favor of the second solution. I poked around a bit and tried to understand how the deserialization works, but I don't see where/how we can add a custom JsonConverter to parse integer values correctly.

I'm not sure what's the best way to proceed here. Did you want to try to make the change or did you want to try to explain to me where the change needs to be made and then have me do it? Either way works for me.

ErikSchierboom:
> Did you want to try to make the change [...]

I think this is probably the easiest option. I'll let you know when I have anything for you to work with!

robkeim:
Sounds good, if there's anything I can do let me know!

ErikSchierboom:
@robkeim See PR #1311 for a fix for the long issue!

robkeim:
Great thanks @ErikSchierboom! I'll go ahead and close PR and review the other one.

PR #1311, opened 2019-08-19
Improved long data type support

ErikSchierboom:
This PR adds improved support for the `long` data type.

### DIFF
--- a/exercises/prime-factors/PrimeFactorsTest.cs
+++ b/exercises/prime-factors/PrimeFactorsTest.cs
@@ -7,42 +7,42 @@ public class PrimeFactorsTest
     [Fact]
     public void No_factors()
     {
-        Assert.Empty(PrimeFactors.Factors(1));
+        Assert.Empty(PrimeFactors.Factors(1L));
     }
 
     [Fact(Skip = "Remove to run test")]
     public void Prime_number()
     {
-        Assert.Equal(new[] { 2 }, PrimeFactors.Factors(2));
+        Assert.Equal(new[] { 2 }, PrimeFactors.Factors(2L));
     }
 
     [Fact(Skip = "Remove to run test")]
     public void Square_of_a_prime()
     {
-        Assert.Equal(new[] { 3, 3 }, PrimeFactors.Factors(9));
+        Assert.Equal(new[] { 3, 3 }, PrimeFactors.Factors(9L));
     }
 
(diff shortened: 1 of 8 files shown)


exercises/prime-factors/PrimeFactorsTest.cs
```diff
@@ -7,42 +7,42 @@ public class PrimeFactorsTest
     [Fact]
     public void No_factors()
     {
-        Assert.Empty(PrimeFactors.Factors(1));
+        Assert.Empty(PrimeFactors.Factors(1L));
```
ErikSchierboom: I've now made it clear that each input value is actually a long value. Previously, an implicit conversion was taking place, but this is more clear I feel.

exercises/say/SayTest.cs
```diff
@@ -8,90 +8,90 @@ public class SayTest
     [Fact]
     public void Zero()
     {
-        Assert.Equal("zero", Say.InEnglish(0));
```
ErikSchierboom: I've now made it clear that each input value is actually a long value. Previously, an implicit conversion was taking place, but this is more clear I feel.

generators/Exercises/Generators/Say.cs
```diff
@@ -8,8 +8,9 @@ public class Say : GeneratorExercise
         protected override void UpdateTestMethod(TestMethod testMethod)
         {
             testMethod.TestedMethod = "InEnglish";
+            testMethod.Input["number"] = (long)testMethod.Input["number"]; 
 
-            if (!(testMethod.Expected is string))
+            if (testMethod.ExpectedIsError)
```
ErikSchierboom: I couldn't resist a tiny cleanup :)

generators/Input/JTokenHelper.cs
```diff
@@ -45,16 +45,12 @@ public static dynamic ConvertJToken(JToken jToken)
 
         private static dynamic ConvertJObject(JObject jObject)
         {
-            var properties = jObject.ToObject<IDictionary<string, dynamic>>();
```
ErikSchierboom: This line was the main issue, as the conversion here would result in the dictionary not containing `JToken` values, which meant that the `ConvertJToken` functionality (which already had code to deal with long/int value conversions) was not called.

Interestingly, the new code is way easier!

PR #1312, opened 2019-08-21
Update prime factors to return longs

### LINKED ISSUE #1309: Inherent flaw in PrimeFactor exercise

robkeim:
Fixes #1309

### DIFF
--- a/generators/Exercises/Generators/PrimeFactors.cs
+++ b/generators/Exercises/Generators/PrimeFactors.cs
@@ -1,10 +1,18 @@
-﻿using Exercism.CSharp.Output;
+﻿using System.Linq;
+using Exercism.CSharp.Output;
 
 namespace Exercism.CSharp.Exercises.Generators
 {
     public class PrimeFactors : GeneratorExercise
     {
-        protected override void UpdateTestMethod(TestMethod testMethod) =>
+        protected override void UpdateTestMethod(TestMethod testMethod)
+        {
             testMethod.Input["value"] = (long)testMethod.Input["value"];
+
+            if (testMethod.Expected is int[] expected)
+            {
+                testMethod.Expected = expected.Select(l => (long)l).ToArray();
+            }
+        }
     }
 }
--- a/exercises/prime-factors/Example.cs
+++ b/exercises/prime-factors/Example.cs
@@ -2,9 +2,9 @@ using System.Collections.Generic;
 
 public class PrimeFactors
 {
-    public static int[] Factors(long number)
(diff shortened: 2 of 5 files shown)


generators/Exercises/Generators/PrimeFactors.cs
```diff
@@ -1,10 +1,18 @@
...
 namespace Exercism.CSharp.Exercises.Generators
 {
     public class PrimeFactors : GeneratorExercise
     {
-        protected override void UpdateTestMethod(TestMethod testMethod) =>
+        protected override void UpdateTestMethod(TestMethod testMethod)
+        {
             testMethod.Input["value"] = (long)testMethod.Input["value"];
+
+            if (testMethod.Expected is int[])
```
ErikSchierboom: I'm on mobile, so that makes reviewing a bit harder. I think you can do "Expected is int[] expected", which allows you to remove the cast below.

generators/Exercises/Generators/PrimeFactors.cs
```diff
@@ -1,10 +1,18 @@
...
+            {
+                testMethod.Expected = ((int[])testMethod.Expected).Select(l => (long)l).ToArray();
```
ErikSchierboom: I'm not entirely sure if it works, but you could check to see if instead of the select and the lambda, you could use "Cast<long>".

generators/Exercises/Generators/PrimeFactors.cs
robkeim: Nice I didn't know you could do that!

generators/Exercises/Generators/PrimeFactors.cs
robkeim: I tried the following:
`testMethod.Expected = expected.Cast<long>();`

But that gives the following error:
`Unable to cast object of type 'System.Int32' to type 'System.Int64'.'`

I'd expect that error in the other sense, but not here... 🤔

ErikSchierboom:
Thanks!

PR #1324, opened 2019-10-07
Matching brackets update

ErikSchierboom:


generators/Exercises/Generators/MatchingBrackets.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace Exercism.CSharp.Exercises.Generators
 {
-    public class BracketPush : GeneratorExercise
+    public class MatchingBrackets : GeneratorExercise
```
robkeim: :+1:

robkeim:
Thanks for helping to keep these up-to-date @ErikSchierboom! :smile:

PR #1328, opened 2019-10-23
Update Clock.cs

ErikSchierboom:
We should not suggest that the student will have an `Hours` or `Minutes, not that the `ToString` method should be overridden.


robkeim:
Thanks @ErikSchierboom! :+1:

PR #1338, opened 2019-12-02
Move from Travis to GitHub actions

ErikSchierboom:
This should hopefully help reduce the time needed to run all tests.


.github/workflows/test.yml
```diff
@@ -0,0 +1,13 @@
+name: Test
+
+on: [push, pull_request]
+
+jobs:
+  test:
+    runs-on: ubuntu-16.04
```
robkeim: Is there any reason we're using 16.04 instead of 18.04?

ErikSchierboom: No good reason. I've updated!

robkeim:
Thanks @ErikSchierboom 🎉 Fingers crossed this fixes our build issues

ErikSchierboom:
Indeed!

PR #1352, opened 2020-01-30
Add test for side effect of sorting function

brunotag:
I noticed quite a high amount of people fall for an implementation of PersonalBest like
```
public int PersonalBest()
{
        scores.Sort();
        return scores.Last();
}
```

this introduces a side effect that breaks Latest().

I think it's worthwhile adding a test for it.

### DIFF
--- a/exercises/high-scores/HighScoresTest.cs
+++ b/exercises/high-scores/HighScoresTest.cs
@@ -60,4 +60,14 @@ public class HighScoresTest
         var sut = new HighScores(new List<int> { 40 });
         Assert.Equal(new List<int> { 40 }, sut.PersonalTopThree());
     }
+    
+    [Fact(Skip = "Remove to run test")]
+    public void Lastest_score_messed_up_after_calling_personal_best()
+    {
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
@brunotag Thanks for the PR! I think this would be a nice addition to the test suite. The test suite is auto-generated though, so this change should be built into [this exercise's generator](https://github.com/exercism/csharp/blob/master/generators/Exercises/Generators/HighScores.cs). For documentation on the test generators, see [this page](https://github.com/exercism/csharp/blob/master/docs/GENERATORS.md). 

Let me know if that makes sense.

brunotag:
oh cool I'll have a read. I hope the generation system can support this kind of test.

ErikSchierboom:
@brunotag I think you'll just have to add it manually, as in adding it as a string. There are some other generators already doing this.

robkeim:
Hey @brunotag, are you still planning on looking into the generators?

valentin-p:
Done as part of PR #1422

ErikSchierboom:
Superseded by #1422.

PR #1378, opened 2020-03-12
Fix tests suffix

ErikSchierboom:


ErikSchierboom:
No problem! I just noticed that some tests were skipped :)

PR #1379, opened 2020-03-12
Use hashes when checking out GH actions

ErikSchierboom:
This is a security feature


robkeim:
Thanks @ErikSchierboom!

PR #1382, opened 2020-03-28
added partially paired internal brackets test

shm613:
I added a test for internal partially paired brackets.

### DIFF
--- a/exercises/matching-brackets/MatchingBracketsTests.cs
+++ b/exercises/matching-brackets/MatchingBracketsTests.cs
@@ -59,6 +59,13 @@ public class MatchingBracketsTests
         var value = "{[]}";
         Assert.True(MatchingBrackets.IsPaired(value));
     }
+    
+    [Fact(Skip = "Remove to run test")]
+    public void Partially_paired_internal_brackets()
+    {
+        var value = "{[)}";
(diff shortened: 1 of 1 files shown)


shm613:
I am not sure if I am reading the test correctly but I believe the failure is within the DndCharacter exercise and I modified the MatchingBracket exercise. From my understanding of testing I shouldn't have broken anything within another exercise???

robkeim:
Thanks for the contribution @shm613!

The file you modified is actually a file that is automatically generated based on our test generators (that are based on shared canonical data between the different language tracks). In order to incorporate this change you would need to update the test generator to add this custom test. You can find an example of this being done for the DnD test that failed the first time your build ran:
https://github.com/exercism/csharp/blob/dcb8d6dcee83589e1b0eb9277edc4cf2570197a1/generators/Exercises/Generators/DndCharacter.cs#L83

Regarding the failed test, it's a test that's verifying probabilities so it can fail for statistical anomalies, like what happened in your case. When I re-ran the build the test passed. That's the first time I saw that test fail randomly like this, so I'll keep my eye on it and if it becomes problematic in the future we can address it.

ErikSchierboom:
This PR has been open for a long time with no activity. I'm gonna close this PR for now. @shm613 thanks for working on this.

PR #1394, opened 2020-05-08
Include step for running tests automatically

rabuf:
`dotnet watch test` allows the tests to be executed each time one of the source files is changed on disk. This is a useful quality-of-life feature for developing solutions.

### DIFF
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -6,6 +6,12 @@ To run the tests, execute the following command:
 dotnet test
 ```
 
+To run the tests automatically as each source file is changed, execute the following command:
+
+```bash
+dotnet watch test
+```
+
 ## Solving the exercise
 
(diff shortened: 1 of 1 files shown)


wolf99:
Is there a similar command to _stop_ watch tests? It may not be obvious to people new to dotnet, perhaps that could be added also.

rabuf:
Good point. Control-c will stop the test runner on Linux and MacOS. Should also work on Windows but I don’t have a machine to verify that with.

wolf99:
Sorry, I meant "disable", not stop. i.e. undo the `dotnet watch test` command so that tests can still be run but will no longer run on every file change

rabuf:
Ah, that’s the same thing in this case. It’s either running or not running. When it’s running it watches all the project files and runs the tests when any changes. When it’s not, you’re just in the shell and can either restart with watch or just run tests normally or whatever you want.

ErikSchierboom:
Thanks @rabuf!

PR #1420, opened 2020-10-07
added new exercise resistor trio

### LINKED ISSUE #1419: Implement resistor-color-trio exercise
Based on existing the problem specification

valentin-p:
resolves #1419 

I implemented the exercise resistor color trio, let me know if there is anything that should be improved.
I added some automating to get the right version of the nuget packages in the csproj for any new exercice

### DIFF
--- a/add-new-exercise.ps1
+++ b/add-new-exercise.ps1
@@ -45,7 +45,7 @@ function Add-Project {
 
     $csProj = "$exerciseDir/$exerciseName.csproj"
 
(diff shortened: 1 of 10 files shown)


ErikSchierboom (approved):
Great work!

I also really appreciate the update to the `add-new-exercise.ps1` script. For future reference, PRs should normally only do one thing only. In this case, I would have suggested to do a separate PR to fix up the `add-new-exercise.ps1` script as that technically does not have anything to do with the resistor-color-trio exercise. But it's fine for now!

Thanks a ton for doing this 🎉

PR #1434, opened 2020-11-19
Added test case for idempotency of `PersonalTopThree` method in `HighScoresTests`

pjangam:


ErikSchierboom (changes requested):
Hi! Thanks for wanting to add some functionality! The usual process is to open an issue first to check if everyone is on board with the change. That said, the change does look good. The thing is though that the test suite is automatically generated by a test generator. This means that manually adding a test to just the test file will be overwritten when the generator is run again. Could you add the test to the generator for this exercise: https://github.com/exercism/csharp/blob/master/generators/Exercises/Generators/HighScores.cs#L25 and then run the generator for that exercise to verify it works? There is documentation on the test generators here: https://github.com/exercism/csharp/blob/master/docs/GENERATORS.md

pjangam:
@ErikSchierboom Thanks. Will make sure to raise the issue first before raising PR from next time onwards.

Will have a look at the generators and update this PR.

ErikSchierboom:
@pjangam Great! Let me know if you need any help :)

ErikSchierboom:
@pjangam Let me know when to take a new look at this PR!

pjangam:
@ErikSchierboom You can check the PR. I've already updated generators

generators/Exercises/Generators/HighScores.cs
```diff
@@ -30,6 +30,14 @@ public void Latest_score_should_not_change_after_calling_personal_best()
     var sut = new HighScores(new List<int> { 20, 10, 30, 3, 2, 1 });
     Assert.Equal(30, sut.PersonalBest());
     Assert.Equal(1, sut.Latest());
+}
+
+[Fact(Skip = ""Remove this Skip property to run this test"")]
```
ErikSchierboom, line 35: If you run the test generator for this exercise using `pwsh .\generate-tests.ps1 -Exercise high-scores`, you'll find that there is no newline between the new method and the old additional method. The fix for this is relatively easy. Instead of using one big string for both additional methods, add them separately:

```suggestion
}");

testClass.AdditionalMethods.Add(@"[Fact(Skip = ""Remove this Skip property to run this test"")]
```

ErikSchierboom (commented):
One tiny nit!

pjangam:
@ErikSchierboom PR updated

ErikSchierboom (approved):
Perfect! Thanks!

Issue #1492, opened 2021-02-08
Lint powershell scripts


ErikSchierboom:
Use https://github.com/PowerShell/PSScriptAnalyzer to verify that the powershell scripts adhere to the standards.

PR #1497, opened 2021-02-11
Rework all the PowerShell files to avoid errors

valentin-p:
- Add parameters to inline functions
- Change function names to singular:
> PSUseSingularNouns - The cmdlet 'Copy-Exercises' uses a plural noun. A singular noun should be used instead
- Use ShouldProcess when required.
> PSUseShouldProcessForStateChangingFunctions - Function 'Update-Canonical-Data' has verb that could change system state. Therefore, the function has to support 'ShouldProcess'.
- Avoid use of `args`
- PSAvoidUsingPositionalParameters
- PSUseApprovedVerbs


valentin-p:
I could not test all the scripts yet because some script are broken due to v3 changes.

valentin-p:
The only error left is:
```
Warning [19, 5] shared.ps1 PSAvoidUsingInvokeExpression Invoke-Expression is used. Please remove Invoke-Expression from script and find other options instead.
```

add-new-exercise.ps1
```diff
@@ -81,35 +85,67 @@ function Add-Project {
...
+    $generator = Join-Path $generatorsExerciseGeneratorsDir "$ExerciseName.cs"
+    $generatorClass = "namespace Exercism.CSharp.Exercises.Generators`n{`n    public class $ExerciseName : GeneratorExercise`n    {`n    }`n}"
     
     Set-Content -Path $generator -Value $generatorClass
 }
 
-function Copy-Track-Files {
-    Write-Output "Copying track files"
-    ./copy-track-files.ps1 $Exercise
+function Copy-ExerciseTrackFile {
```
ErikSchierboom, line 104: Why did you add "Exercise"? Is the following also possible?

```suggestion
function Copy-TrackFiles {
```

add-new-exercise.ps1
```diff
@@ -81,35 +85,67 @@ function Add-Project {
...
+    if ($PSCmdlet.ShouldProcess($true)) {
+        Write-Output "Updating README"
+        ./update-docs.ps1 $Exercise
+    }
 }
 
-function Update-Tests { 
-    Write-Output "Updating test suite"
-    ./generate-tests.ps1 $Exercise
+function Update-ExerciseTest {
```
ErikSchierboom, line 128: Similar question as above: why the "Exercise" prefix? Can this be:

```suggestion
function Update-Tests {
```

bin/fetch-configlet.ps1
```diff
@@ -2,25 +2,28 @@ $ErrorActionPreference = "Stop"
...
 
-$arch = If ([Environment]::Is64BitOperatingSystem) { "64bit" } Else { "32bit" }
-$fileName = "configlet-windows-$arch.zip"
-
-Function Get-DownloadUrl {
+function Get-DownloadUrl($FileName, $RequestOpts) {
+    if ($PSVersionTable.PSVersion.Major -le 5) {
+        # PreserveAuthorizationOnRedirect requires PowerShell 7+
+        throw "Please upgrade to PowerShell Core. For example: 'dotnet tool install --global PowerShell'"
+    }
```
ErikSchierboom, line 14: I'd prefer this check to be at the top of the file.

bin/fetch-configlet.ps1
```diff
@@ -2,25 +2,28 @@ $ErrorActionPreference = "Stop"
...
 }
 
-$arch = If ([Environment]::Is64BitOperatingSystem) { "64bit" } Else { "32bit" }
-$fileName = "configlet-windows-$arch.zip"
-
-Function Get-DownloadUrl {
+function Get-DownloadUrl($FileName, $RequestOpts) {
+    if ($PSVersionTable.PSVersion.Major -le 5) {
+        # PreserveAuthorizationOnRedirect requires PowerShell 7+
+        throw "Please upgrade to PowerShell Core. For example: 'dotnet tool install --global PowerShell'"
```
ErikSchierboom, line 13: Did you look into the dotnet tool manifest? We can add that later too if you'd prefer that.

bin/fetch-configlet.ps1
```diff
@@ -2,25 +2,28 @@ $ErrorActionPreference = "Stop"
...
 
-$downloadUrl = Get-DownloadUrl
 $outputDirectory = "bin"
+$arch = If ([Environment]::Is64BitOperatingSystem) { "64bit" } Else { "32bit" }
+$fileName = "configlet-windows-$arch.zip"
+
+$downloadUrl = Get-DownloadUrl $fileName $requestOpts
 $outputFile = Join-Path -Path $outputDirectory -ChildPath $fileName
-Invoke-WebRequest -Uri $downloadUrl -OutFile $outputFile @requestOpts
+Invoke-WebRequest -Uri $downloadUrl -OutFile $outputFile -PreserveAuthorizationOnRedirect @requestOpts
```
ErikSchierboom, line 27: I think I purposefully not added that as it did not work. Did you check to see if it worked?

copy-track-files.ps1
```diff
@@ -21,17 +21,23 @@ param (
...
+
+    $defaultEditorConfigSettings = Get-Content -Path ".editorconfig"
     $editorConfigSettings = $defaultEditorConfigSettings.Replace( "[*.cs]", "[${exerciseName}.cs]")
     $exerciseEditorConfigPath = Join-Path $ExerciseDirectory.FullName ".editorconfig"
 
     Set-Content -Path $exerciseEditorConfigPath $editorConfigSettings
 }
 
-function Copy-Track-Files {
+function Copy-TrackFile {
```
ErikSchierboom, line 35:
```suggestion
function Copy-TrackFiles {
```

update-docs.ps1
```diff
@@ -22,18 +22,31 @@ param (
 . ./shared.ps1
 
 function Update-Canonical-Data {
```
ErikSchierboom, line 24: Should this be:

```suggestion
function Update-CanonicalData {
```

add-new-exercise.ps1
```diff
@@ -36,30 +36,34 @@ param (
...
+    param (
+        [Parameter(Position = 0, Mandatory = $true)][string]$Exercise,
+        [Parameter(Position = 1, Mandatory = $true)][string]$ExerciseName
+    )
+
     Write-Output "Adding project"
 
-    $csProj = "$exerciseDir/$exerciseName.csproj"
+    $exercisesDir = Resolve-Path "exercises"
+    $exerciseDir = Join-Path $exercisesDir $Exercise
```
ErikSchierboom, line 48: This won't be the correct path, as we have practice and concept exercises. Could you update to:

```suggestion
    $exercisesDir = Resolve-Path "exercises"
    $exerciseDir = Join-Path $exercisesDir "practice" $Exercise
```

Could you add a TODO indicating that we should support concept exercises too?

add-new-exercise.ps1
```diff
@@ -36,30 +36,34 @@ param (
...
 
-    Run-Command "dotnet new xunit -lang ""C#"" --target-framework-override netcoreapp3.0 -o $exerciseDir -n $exerciseName"
-    Run-Command "dotnet sln ""$exercisesDir/Exercises.sln"" add $csProj"
+    $csProj = "$exerciseDir/$ExerciseName.csproj"
+
+    Invoke-CommandExecution "dotnet new xunit -lang ""C#"" --target-framework-override net5.0 -o $exerciseDir -n $ExerciseName"
```
ErikSchierboom, line 52: `Invoke-CommandExecution` sounds a bit odd to me. Is `Invoke-Command` also possible?

add-new-exercise.ps1
```diff
@@ -81,35 +85,67 @@ function Add-Project {
...
-function Copy-Track-Files {
-    Write-Output "Copying track files"
-    ./copy-track-files.ps1 $Exercise
+function Copy-ExerciseTrackFile {
+    [CmdletBinding(SupportsShouldProcess)]
+    param (
+        [Parameter(Position = 0, Mandatory = $true)][string]$Exercise
+    )
+
+    if ($PSCmdlet.ShouldProcess($true)) {
```
ErikSchierboom, line 110: What does this do?

add-new-exercise.ps1
```diff
@@ -81,35 +85,67 @@ function Add-Project {
...
+        Write-Output "Copying track files"
+        ./copy-track-files.ps1 $Exercise
+    }
 }
 
-function Update-Readme {
-    Write-Output "Updating README"
-    ./update-docs.ps1 $Exercise
+function Update-ExerciseReadme {
```
ErikSchierboom, line 116: Similar question as above: why the "Exercise" prefix?

bin/fetch-configlet.ps1
```diff
@@ -2,25 +2,28 @@ $ErrorActionPreference = "Stop"
...
 Expand-Archive $outputFile -DestinationPath $outputDirectory -Force
-Remove-Item -Path $outputFile
+Remove-Item -Path $outputFile
```
ErikSchierboom, line 29: Could you add the newline at the end of the file? It makes git diffs easier to spot.

generate-tests.ps1
```diff
@@ -23,15 +23,31 @@ param (
 . ./shared.ps1
 
 function Update-Canonical-Data {
```
ErikSchierboom, line 25: Should this be:

```suggestion
function Update-CanonicalData {
```

generate-tests.ps1
```diff
@@ -23,15 +23,31 @@ param (
...
-    Run-Command "dotnet run --project ./generators $args"
+function Update-TestFile {
+    [CmdletBinding(SupportsShouldProcess)]
+    param (
+        [Parameter(Position = 0, Mandatory = $false)]
+        [string]$Exercise
+    )
+
+    $generators_exec = "./generators"
+    $generators_args = if ($Exercise) { @("--exercise", $Exercise) } else { @() }
```
ErikSchierboom, line 43:
```suggestion
    $generatorsProject = "./generators"
    $generatorsArgs = if ($Exercise) { @("--exercise", $Exercise) } else { @() }
```

test.ps1
```diff
@@ -25,89 +25,100 @@ param (
...
+function Invoke-Configlet-Lint {
     Write-Output "Linting config.json"
-    Run-Command "./bin/fetch-configlet"
-    Run-Command "./bin/configlet lint"
+    Invoke-CommandExecution "./bin/fetch-configlet"
+    Invoke-CommandExecution "./bin/configlet lint"
 }
 
-function Build-Generators { 
+function Invoke-DotNetBuild-Generators {
```
ErikSchierboom, line 35:
```suggestion
function Invoke-Build-Generators { 
```

test.ps1
```diff
@@ -25,89 +25,100 @@ param (
...
 
-function Clean {
+function Clean($BuildDir) {
     Write-Output "Cleaning previous build"
-    Remove-Item -Recurse -Force $buildDir -ErrorAction Ignore
+    Remove-Item -Recurse -Force $BuildDir -ErrorAction Ignore
 }
-
-function Copy-Exercises {
+function Copy-Exercise($SourceDir, $BuildDir) {
```
ErikSchierboom, line 44:
```suggestion
function Copy-Exercises($SourceDir, $BuildDir) {
```

test.ps1
```diff
@@ -25,89 +25,100 @@ param (
...
     Write-Output "Copying exercises"
-    Copy-Item $sourceDir -Destination $buildDir -Recurse
+    Copy-Item $SourceDir -Destination $BuildDir -Recurse
 }
 
-function Enable-All-Tests {
+function ConvertTo-UnSkippedTest($BuildDir) {
```
ErikSchierboom, line 49: I don't really like this new name. Is Enable forbidden?

test.ps1
```diff
@@ -25,89 +25,100 @@ param (
...
-    Get-ChildItem -Path $buildDir -Include "*Tests.cs" -Recurse | ForEach-Object {
+    Get-ChildItem -Path $BuildDir -Include "*Tests.cs" -Recurse | ForEach-Object {
         (Get-Content $_.FullName) -replace "Skip = ""Remove this Skip property to run this test""", "" | Set-Content $_.FullName
     }
 }
 
-function Test-Refactoring-Projects {
-    Write-Output "Testing refactoring projects"
-    @("tree-building", "ledger", "markdown") | ForEach-Object { Run-Command "dotnet test $practiceExercisesDir/$_" }
+function Test-ExeciseWithRefactoring($PracticeExercisesDir) {
```
ErikSchierboom, line 56: Why this naming change?

test.ps1
```diff
@@ -25,89 +25,100 @@ param (
...
+    @("tree-building", "ledger", "markdown") | ForEach-Object { Invoke-CommandExecution "dotnet test $practiceExercisesDir/$_" }
 }
 
-function Replace-Stubs-With-File ($ExercisesDir, $ReplaceFileName) {
-    Get-ChildItem -Path $ExercisesDir -Include "*.csproj" -Recurse | ForEach-Object {
-        $stub = Join-Path $_.Directory ($_.BaseName + ".cs")
-        $example = Join-Path $_.Directory ".meta" $ReplaceFileName
-    
-        Move-Item -Path $example -Destination $stub -Force
+function Set-SolutionFileForExercise {
```
ErikSchierboom, line 61: Why this naming change?

test.ps1
```diff
@@ -25,89 +25,100 @@ param (
...
+        
+            Move-Item -Path $example -Destination $stub -Force
+        }
     }
 }
 
-function Replace-Stubs {
-    Write-Output "Replacing concept exercise stubs with exemplar"
-    Replace-Stubs-With-File $conceptExercisesDir "Exemplar.cs"
+function Set-AllSolutionsForExercise {
```
ErikSchierboom, line 75: Why this naming change?

test.ps1
```diff
@@ -25,89 +25,100 @@ param (
...
 
-    Write-Output "Replacing practice exercise stubs with example"
-    Replace-Stubs-With-File $practiceExercisesDir "Example.cs"
+        Write-Output "Replacing practice exercise stubs with example"
+        Set-SolutionFileForExercise $PracticeExercisesDir "Example.cs"
+    }
 }
 
-function Test-Using-Example-Implementation {
+function Test-EveryExerciseImplementation($Exercise, $BuildDir, $ConceptExercisesDir, $PracticeExercisesDir) {
```
ErikSchierboom, line 88: Why this naming change?

update-docs.ps1
```diff
@@ -22,18 +22,31 @@ param (
...
+        Invoke-CommandExecution "./update-canonical-data.ps1" 
+    }
 }
 
-function Update-Docs {
-    Write-Output "Updating docs"
-    $args = if ($Exercise) { @("-o", $Exercise) } else { @() }
-    Run-Command "./bin/fetch-configlet"
-    Run-Command "./bin/configlet generate . -p problem-specifications $args"
+function Update-DocFile {
```
ErikSchierboom, line 34: Why this naming change?

valentin-p: PSAnalyzer rejects plurals:
PSUseSingularNouns

update-docs.ps1
```diff
@@ -22,18 +22,31 @@ param (
...
+    [CmdletBinding(SupportsShouldProcess)]
+    param (
+        [Parameter(Position = 0, Mandatory = $false)]
+        [string]$Exercise
+    )
+
+    if ($PSCmdlet.ShouldProcess($true)) {
```
valentin-p, line 41: What if: Performing the operation "fetch configlet and pull specifications" on target "All Exercises".

`Update-DocsFolder"example" -WhatIf`
`What if: Performing the operation "fetch configlet and pull specifications" on target "example"`

bin/fetch-configlet.ps1
valentin-p: We could do later

bin/fetch-configlet.ps1
valentin-p: I was wondering if it was removed on purpose, so I did add any.

bin/fetch-configlet.ps1
valentin-p: Both with and without works fine for me, so wanted to use only named parameters.

add-new-exercise.ps1
valentin-p: `Invoke-Command` was my first try, but it's a reserved name.
Maybe `Invoke-ExpressionExitOnError` ?

add-new-exercise.ps1
```diff
@@ -36,30 +36,34 @@ param (
...
     New-Item -ItemType File -Path "$exerciseDir/Example.cs"
     
     [xml]$proj = Get-Content $csProj
     $propertyGroup = $proj.Project.PropertyGroup
 
     $nugetItemGroup = $proj.Project.ItemGroup;
     $nugetItemGroup.RemoveAll();
-    $nugetList = @(@{nuget="Microsoft.NET.Test.Sdk";Version="16.7.1"}, @{nuget="xunit";Version="2.4.1"}, @{nuget="xunit.runner.visualstudio";version="2.4.3"})
-    $nugetList | % { 
+    $nugetList = @(@{nuget = "Microsoft.NET.Test.Sdk"; Version = "16.8.3" }, @{nuget = "xunit"; Version = "2.4.1" }, @{nuget = "xunit.runner.visualstudio"; version = "2.4.3" })
```
valentin-p, line 65: update the version to match the current exercises

ErikSchierboom: Excellent!

test.ps1
valentin-p: Replace-Stubs had two warning:
- Replace: PSUseApprovedVerbs - The cmdlet 'Replace-Stubs' uses an unapproved verb.
- Stubs: PSUseSingularNouns  - The cmdlet 'Replace-Stubs' uses a plural noun. A singular noun should be used instead.

update-docs.ps1
ErikSchierboom: That's really nice. I think you forgot to add `[CmdletBinding(SupportsShouldProcess)]` to the top-level `param`, as otherwise it won't flow to this sub-task.

update-docs.ps1
ErikSchierboom: Ah, I find that a weird rule, but let's see if we can work around that. I don't like `Update-DocFile` because it suggests only one doc file is changed and that is just incorrect. In general, if the singular version is not actually correct, I'd prefer disabling the analyzer rule over a valid name that is incorrect.

That said, would `Update-Documentation` work here?

test.ps1
ErikSchierboom: > The cmdlet 'Replace-Stubs' uses an unapproved verb.

Oh my, I think that is a very harsh rule. How could they possible have a whitelist of all valid verbs?

How about we rename to `Use-ExampleImplementation`? Is that valid?

bin/fetch-configlet.ps1
ErikSchierboom: Ah, it does indeed work on Windows. I tested it using powershell on Linux (WSL), where it fails with:

```
erik@VELPIE:~/csharp$ pwsh ./bin/fetch-configlet.ps1 
Invoke-WebRequest: /home/erik/csharp/bin/fetch-configlet.ps1:27:1
Line |
  27 |  Invoke-WebRequest -Uri $downloadUrl -OutFile $outputFile -PreserveAut …
     |  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Cannot access a disposed object. Object name: 'System.Net.Http.HttpClient'.
```

This is apparently some bug that only occurs on linux. I'd really like the powershell to also work on linux, so could you revert this change?

test.ps1
valentin-p: Test-Refactoring-Projects. Project is not singular.
new names could be:
Test-Refactoring-Exercise
Test-Refactoring-Project

add-new-exercise.ps1
valentin-p: done

add-new-exercise.ps1
valentin-p: see other comments.

test.ps1
ErikSchierboom: I can't come up with a better name, so I'd prefer to ignore the naming rule entirely. I think it leads to less easy to understand code, so I think we should just completely ignore the naming rule and revert the naming changes.

I see that the analyzer supports a settings file to ignore rules: https://github.com/PowerShell/PSScriptAnalyzer#explicit

test.ps1
valentin-p: https://docs.microsoft.com/en-us/powershell/scripting/developer/cmdlet/approved-verbs-for-windows-powershell-commands?view=powershell-7.1

yes :)

test.ps1
valentin-p: you found a better name

test.ps1
valentin-p: it would be plural

update-docs.ps1
valentin-p: Yes, or Update-DocsFolder ?

I think this is what actually happens, some files in the folder docs are updated ?

add-new-exercise.ps1
```diff
@@ -81,35 +87,67 @@ function Add-Project {
...
+    [CmdletBinding(SupportsShouldProcess)]
+    param (
+        [Parameter(Mandatory = $true, Position = 0)][string]$Exercise,
+        [Parameter()][string[]]$Topics = @(),
+        [Parameter()][switch]$Core,
+        [Parameter()][int]$Difficulty = 1,
+        [Parameter()]$UnlockedBy
+    )
 
     $configJson = Resolve-Path "config.json"
```
valentin-p, line 153: the config file changes, so currently this does not work

generate-tests.ps1
valentin-p: I can use the file now.

add-new-exercise.ps1
ErikSchierboom: Yeah, we need to do some tweaking later. Or possibly two separate scripts.

update-docs.ps1
ErikSchierboom: Yeah that could work

bin/fetch-configlet.ps1
valentin-p: I tested on WSL ubuntu 20.04.1 LTS and WSL 18.04LTS with pwsh 7.1 on Win10-20H2, and both with and without the named parameter `-PreserveAuthorizationOnRedirect`  work fine.

What was you tested on?

copy-track-files.ps1
valentin-p: Copy-TrackFilesForExercise works for the analyzer.

update-docs.ps1
```diff
@@ -13,27 +13,33 @@
...
 
 # Import shared functionality
 . ./shared.ps1
+. ./update-canonical-data.ps1
 
-function Update-Canonical-Data {
-    Write-Output "Updating canonical data"
-    Run-Command "./update-canonical-data.ps1" 
-}
+function Update-DocsFolder {
```
ErikSchierboom, line 26:
```suggestion
function Update-Documentation {
```

update-docs.ps1
```diff
@@ -13,27 +13,33 @@
...
-    Run-Command "./bin/configlet generate . -p problem-specifications $args"
+        $configletArgs = if ($Exercise) { @("-e", $Exercise) } else { @() }
+        Invoke-ExpressionExitOnError "./bin/configlet sync -p problem-specifications -o $configletArgs"
+    }
 }
 
-Update-Canonical-Data
-Update-Docs
+Update-CanonicalData
+Update-DocsFolder $Exercise
```
ErikSchierboom, line 43: I see some changes where parameters are passed by name. Would it perhaps make sense to do that everywhere, as I think it can help with readability and also is beneficial when refactoring later on.

So something like:

```suggestion
Update-DocsFolder -Exercise $Exercise
```

And then require the parameter to be specified by name and not by position anymore.

add-new-exercise.ps1
```diff
@@ -73,67 +80,108 @@ function Add-Project {
...
+    if ($PSCmdlet.ShouldProcess("exercise $Exercise", "update Readme files")) {
+        Write-Output "Updating README"
+        ./update-docs.ps1 $Exercise
+    }
 }
 
-function Update-Tests { 
-    Write-Output "Updating test suite"
-    ./generate-tests.ps1 $Exercise
+function Update-Test {
```
ErikSchierboom, line 131: What about:

```suggestion
function Update-TestSuite {
```

bin/fetch-configlet.ps1
```diff
@@ -1,26 +1,34 @@
...
-
-Function Get-DownloadUrl {
+function Get-DownloadUrl($FileName, $RequestOpts) {
     $latestUrl = "https://api.github.com/repos/exercism/configlet/releases/latest"
-    Invoke-RestMethod -Uri $latestUrl -PreserveAuthorizationOnRedirect @requestOpts
-    | Select-Object -ExpandProperty assets
-    | Where-Object { $_.browser_download_url -match $FileName }
-    | Select-Object -ExpandProperty browser_download_url
+    $assets = Invoke-RestMethod -Uri $latestUrl -PreserveAuthorizationOnRedirect @RequestOpts | Select-Object -ExpandProperty assets
+    return $assets | Where-Object { $_.browser_download_url -match $FileName } | Select-Object -ExpandProperty browser_download_url -First 1
```
ErikSchierboom, line 13: Was the return a suggestion by the analyzer?

ErikSchierboom: Same question for the -First 1

generate-tests.ps1
```diff
@@ -13,25 +13,31 @@
...
-
 # Import shared functionality
 . ./shared.ps1
+. ./update-canonical-data.ps1
 
-function Update-Canonical-Data {
-    Write-Output "Updating canonical data"
-    Run-Command "./update-canonical-data.ps1" 
-}
+function Update-TestFile {
```
ErikSchierboom, line 26:
```suggestion
function Update-TestFilesForTrack {
```

ErikSchierboom (approved):
I'm happy to merge this and do improvements later.

update-docs.ps1
valentin-p: yes I did that.

bin/fetch-configlet.ps1
valentin-p: So the analyzer was unhappy with the pipe with no argument before.

```
ParseError [16 , 5] fetch-configlet.ps1 EmptyPipeElement - An empty pipe element is not allowed.
```

If you prefer to rollback or have another idea, I am good with both.

valentin-p:
Relates to #1492

Issue #1499, opened 2021-02-11
Docuent Concepts Required for Practice Exercises


mikedamay:
This is a working document.  Please add concepts as required and show them prefixed with a "(documented)" when they have been documented.

* lambdas
* enumerables

Issue #1459, opened 2021-01-28
[C#] add explanation of concepts to various design.md documents


mikedamay:
There is a mismatch (see tables, below) between the concepts attributed to an exercise in [csharp/reference/README.md](https://github.com/exercism/v3/blob/master/languages/csharp/reference/README.md) which reflects the original concept-scraping task performed by the track's maintainers and what is shown for each exercise in [csharp/config.json](https://github.com/exercism/v3/blob/master/languages/csharp/config.json) which is the official catalog of completed exercises.

The task, here, is to align the concepts associated in an exercise in _csharp/config.json_ with those in _csharp/reference/exercises.json_ (from which _csharp/reference/README.md_ is generated).

For each concept found in _config.json_, not in _exercises.json_ 
- check if the concept is fully illustrated or described in the exercise to which it is attributed
- if the concept is fully illustrated or described then:
  - add the concept's name and a brief description to the exercise's _design.md_ document
  - add the concept to the exercise in _csharp/reference/exercises.json_.
- if the concept is not fully illustrated or described in the exercise but it is an appropriate concept for the exercise then:
  - amend the exercise's _introduction.md_, _after.md_ and/or the code files and instructions to reflect the concept
  - add the concept's name and a brief description to the exercise's _design.md_ document
  - add the concept to the exercise in _csharp/reference/exercises.json_.
- if the concept is not fully illustrated or described in the exercise and it is not an appropriate concept for the exercise then:
  - add the concept to the _unallocated-concepts_ exercise at the end of _exercises.json_ where it can be picked up for further review

For each concept found in _exercises.json_, not in _config.json_ 
- check if the concept is fully illustrated or described in the exercise to which it is attributed
- if the concept if fully illustrated or described then:
  - add the concept to the exercise in _csharp/config.json_.
- if the concept is not fully illustrated or described in the exercise but it is an appropriate concept for the exercise then:
  - amend the exercise's _introduction.md_, _after.md_ and/or the code files and instructions to reflect the concept
  - add the concept to the exercise in _csharp/config.json_.
- if the concept is not fully illustrated or described in the exercise and it is not an appropriate concept for the exercise then:
  - take the concept from the exercise in _exercises.json_ and move it to the _unallocated-concepts_ exercise in _exercises.json_ where it can be picked up for further review

### Concepts in csharp/config.json not in csharp/reference/exercises.json

| Exercise | Concept |
|----------|---------|
| flag-enums | compound-assignment |
| integral-numbers | casting |
[...]

PR #1500, opened 2021-02-11
add practice and prerequisites - 1

mikedamay:
I'll update a few practice exercises in _config.json_ merge that and then add document the missing concepts separately.


ErikSchierboom:
Thanks for working on this! I'll do a proper review once you're done.

mikedamay:
Please review this as is.  I'm not going to do all 120 in a single PR.  It will lock up the _config.json_ / create merge conflicts.  I'm also documenting the concepts in separate PRs.  It's a question of keeping things straight.  I'm even finding anomalies with the concept exercises that I had not expected.

I will keep PRs as draft until a review is appropriate.

config.json
```diff
@@ -1059,8 +1059,8 @@
         "slug": "accumulate",
         "name": "Accumulate",
         "uuid": "3c0563dc-665a-45b4-9b29-f133e235efd0",
-        "practices": [],
-        "prerequisites": [],
+        "practices": ["lambdas", "foreach"],
```
ErikSchierboom, line 1062: Maybe also add `yield`? I think that's allowed.

```suggestion
        "practices": ["lambdas", "foreach", "yield"],
```

config.json
```diff
@@ -653,8 +653,8 @@
         "slug": "leap",
         "name": "Leap",
         "uuid": "8ba15933-29a2-49b1-a9ce-70474bad3007",
-        "practices": [],
-        "prerequisites": [],
+        "practices": ["math-operators", "if-statements"],
+        "prerequisites": ["math-operators", "if-statements"],
```
ErikSchierboom, line 657: We don't yet have a Concept Exercise that unlocks `math-operators` . This means that for the website, this prerequisite will be completely ignored. I was wondering if it would make sense to guard against these types of situations by being explicit and also adding `numbers` as the prerequisite. What do you think?

config.json
```diff
@@ -641,8 +641,8 @@
         "slug": "two-fer",
         "name": "Two Fer",
         "uuid": "57f02d0e-7b75-473b-892d-26a7d980c4ce",
-        "practices": [],
-        "prerequisites": [],
+        "practices": ["optional-parameters", "interpolation"],
+        "prerequisites": ["optional-parameters", "interpolation"],
```
ErikSchierboom, line 645: According to the config.json file, the concept is `string-interpolation`:

```suggestion
        "practices": ["optional-parameters", "string-interpolation"],
        "prerequisites": ["optional-parameters", "string-interpolation"],
```

ErikSchierboom: Like with my below comment, maybe guard against the `string-interpolation` not being unlocked by any Concept Exercise and add `strings` to the prerequisites?

config.json
```diff
@@ -1059,8 +1059,8 @@
         "slug": "accumulate",
         "name": "Accumulate",
         "uuid": "3c0563dc-665a-45b4-9b29-f133e235efd0",
-        "practices": [],
-        "prerequisites": [],
+        "practices": ["lambdas", "foreach"],
+        "prerequisites": ["lambdas", "generic-types", "foreach", "enumerables"],
```
ErikSchierboom, line 1063: The method is defined as an extension method:

```suggestion
        "prerequisites": ["lambdas", "generic-types", "foreach", "enumerables", "extension-methods"],
```

ErikSchierboom (approved):

config.json
mikedamay: I thought the mandate was to create concepts for practice exercises even where there was no concept exercise illustrating the concept.  Did I misunderstand?

Issue #1499 addresses the question of concepts without illustrative exercises.  So far we need concepts for `lambdas`, `enumerables`, `math-operators` and `yield` - after looking at 3 exercises!

Issue #1459 (mismatch between _exercises.json_ and _config.json_) also relates to this.  I think it will indeed cause `math-operators` to be folded into `numbers`.  I hesitated to do it in this PR as the links in the `numbers`'s _about.md_ do not include `%` so text will need to be altered as part of issue #1459.

This is what I mean about keeping things straight.

ErikSchierboom: > I thought the mandate was to create concepts for practice exercises even where there was no concept exercise illustrating the concept. Did I misunderstand?

No you understood correctly. This is all just trial and error. I'm just thinking of the effect it would have one students for the time being while we don't have a `math-operators` exercise. That is why I suggested adding the `numbers` prerequisite, althought that would be a slightly duplicate prerequisite once we have `math-operators`.

mikedamay: Fair enuf.  `numbers` in addition.  n.p.

Issue #1538, opened 2021-06-15
Add blurb to Concepts


ErikSchierboom:
Add blurbs to all Concepts.

- [x] arrays
- [x] attributes
- [ ] basics
- [ ] bit-manipulation
- [x] booleans
- [x] casting
- [x] chars
- [x] classes
- [x] compound-assignment
- [ ] const-readonly
- [ ] constants
- [x] constructors
- [x] datetimes
- [x] defensive-copying
- [x] dictionaries
- [x] do-while-loops
- [ ] enums
- [ ] equality
- [ ] exception-filtering
- [x] exceptions
- [ ] explicit-casts
- [x] expression-bodied-members
- [ ] flag-enums
- [x] floating-point-numbers
- [x] for-loops
- [x] foreach-loops
- [x] generic-types
- [x] if-statements
- [ ] indexers
- [x] inheritance
- [ ] integral-numbers
- [x] interfaces
- [x] lists
- [ ] memory-allocation
- [ ] method-overloading
- [ ] named-arguments
- [ ] namespaces
- [ ] nested-types
- [ ] nullability
- [ ] numbers
- [ ] object-initializers
- [ ] operator-overloading
- [ ] optional-parameters
- [ ] ordering
- [ ] overflow
- [ ] parameters
- [ ] properties
- [ ] randomness
- [ ] readonly-collections
- [ ] regular-expressions
- [ ] resource-cleanup
- [ ] resource-lifetime
- [ ] sets
- [ ] string-builder
- [ ] string-formatting
- [ ] string-interpolation
- [ ] strings
- [ ] structs
- [ ] switch-expressions
- [ ] switch-statements
- [ ] ternary-operators
- [ ] throw-expressions
- [ ] time
- [ ] timezone
- [ ] tuples
- [ ] user-defined-exceptions
- [ ] varargs
- [ ] verbatim-strings
- [ ] while-loops

yzAlvin:
updated list

* [x]  arrays
* [x]  attributes
* [ ]  basics
* [ ]  bit-manipulation
* [x]  booleans
* [x]  casting
* [x]  chars
* [x]  classes
* [x]  compound-assignment
* [x]  const-readonly
* [x]  constants
* [x]  constructors
* [x]  datetimes
* [x]  defensive-copying
* [x]  dictionaries
* [x]  do-while-loops
* [x]  enums
* [x]  equality
* [ ]  exception-filtering
* [x]  exceptions
* [ ]  explicit-casts
[...]

PR #1544, opened 2021-06-26
Add blurb to Concepts

yzAlvin:
#1538 

Should I be squashing all the commits into one? not sure about git best practices


ErikSchierboom:
> Should I be squashing all the commits into one? not sure about git best practices

At Exercism, we usually squash when merging, so it is totally fine to have individual commits in a PR. If there is an express desire by the PR author for a merging option, said author should indicate that when creating the PR.

So in this case, I'll do a "Squash and merge" when merging.

ErikSchierboom (approved):
These are all excellent!

PR #1551, opened 2021-07-05
Fix forked_from after exercise renames

angelikatyborska:
I want that extra rep for somebody forking my exercise 😏


ErikSchierboom (approved):
Thanks!

PR #1565, opened 2021-08-11
Update hints.md

FaisalAfroz:
Changed link to point correctly to C# Operator overload reference page instead of c++ Operator overload reference page.


ErikSchierboom (approved):
What a silly mistake! Thanks for fixing.

Issue #1564, opened 2021-08-07
Sieve, test "No_primes_under_two" is incorrect


HugoRoss:
There should not be an ArgumentOutOfRangeException expected in unit test "No_primes_under_two" of exercise "Sieve". In my opinion the question "Give me all primes up to the number of 1" is absolutely okay and should return an empty array instead of throwing an Exception.

An ArgumentOutOfRangeException may be thrown for negative numbers if they are not supported.

PR #1566, opened 2021-08-13
change sieve unit tests

yzAlvin:
no primes under two returns an empty array

throws exception on negative numbers

#1564

### DIFF
--- a/exercises/practice/sieve/.meta/Example.cs
+++ b/exercises/practice/sieve/.meta/Example.cs
@@ -11,13 +11,15 @@ public class Sieve
 
     private static int[] InitializePrimes(int limit)
     {
-        if (limit < 2)
+        if (limit < 0)
         {
             throw new ArgumentOutOfRangeException(nameof(limit));
         }
 
-        var candidates = new Queue<int>(Enumerable.Range(2, limit - 1));
         var primes = new List<int>();
+        if (limit == 0 || limit == 1) return primes.ToArray();
+
+        var candidates = new Queue<int>(Enumerable.Range(2, limit - 1));
         do
         {
             var prime = candidates.Dequeue();
--- a/exercises/practice/sieve/.meta/Example.cs
+++ b/exercises/practice/sieve/.meta/Example.cs
@@ -16,8 +16,8 @@ public class Sieve
             throw new ArgumentOutOfRangeException(nameof(limit));
         }
 
+        if (limit == 0 || limit == 1) return Array.Empty<int>();
         var primes = new List<int>();
-        if (limit == 0 || limit == 1) return primes.ToArray();
 
         var candidates = new Queue<int>(Enumerable.Range(2, limit - 1));
         do
--- a/exercises/practice/sieve/SieveTests.cs
+++ b/exercises/practice/sieve/SieveTests.cs
@@ -1,4 +1,3 @@
-using System;
 using Xunit;
 
 public class SieveTests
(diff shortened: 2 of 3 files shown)


ErikSchierboom (commented):
I like the improvement. The test file for the sieve exercise is automatically generated using a [test generator](https://github.com/exercism/csharp/blob/main/generators/Exercises/Generators/Sieve.cs). This allows us to easily keep tests up to date. Could you update the linked generator to have the desired result? You can read more about test generators here: https://github.com/exercism/csharp/blob/main/docs/GENERATORS.md (note that the docs might be slightly outdated as we had a big refactoring lately).

yzAlvin:
I removed the existing code in the Test Generator because the test case for Sieve(1) already expected an empty array in the canonical-data json,

Just added a TestCase at the end to check negative numbers hope I've done it in a readable way 

The SieveTest changes are just whatever changes happened after I ran `dotnet run -e Sieve`

ErikSchierboom:
@yzAlvin Nice! There are two final things to do:

1. The current generated test code fails with the following error:

> Passed!  - Failed:     0, Passed:    10, Skipped:     0, Total:    10, Duration: 14 ms - /home/runner/work/csharp/csharp/build/practice/series/bin/Debug/net5.0/Series.dll (net5.0)
Error: /home/runner/work/csharp/csharp/build/practice/sieve/SieveTests.cs(38,23): error CS0246: The type or namespace name 'ArgumentOutOfRangeException' could not be found (are you missing a using directive or an assembly reference?) [/home/runner/work/csharp/csharp/build/practice/sieve/Sieve.csproj]
  SimpleCipher -> /home/runner/work/csharp/csharp/build/practice/simple-cipher/bin/Debug/net5.0/SimpleCipher.dll
Test run for /home/runner/work/csharp/csharp/build/practice/simple-cipher/bin/Debug/net5.0/SimpleCipher.dll (.NETCoreApp,Version=v5.0)
Microsoft (R) Test Execution Command Line Tool Version 16.8.3
Copyright (c) Microsoft Corporation.  All rights reserved.

This is because now that we've moved to a stringly based test method, the test generator doesn't know it needs to use the `System` namespace. There is code in the generators that allows you to do that, although I don't know how out of the top of my head (probably something with namespaces).

2. The example implementation needs to be updated to pass the latest version: https://github.com/exercism/csharp/blob/main/exercises/practice/sieve/.meta/Example.cs

yzAlvin:
Thanks!

exercises/practice/sieve/.meta/Example.cs
```diff
@@ -11,13 +11,15 @@ public static int[] Primes(int limit)
...
     {
-        if (limit < 2)
+        if (limit < 0)
         {
             throw new ArgumentOutOfRangeException(nameof(limit));
         }
 
-        var candidates = new Queue<int>(Enumerable.Range(2, limit - 1));
         var primes = new List<int>();
+        if (limit == 0 || limit == 1) return primes.ToArray();
```
ErikSchierboom, line 20:
```suggestion
        if (limit == 0 || limit == 1) return Array.Empty<int>();
        var primes = new List<int>();
```

ErikSchierboom (approved):

ErikSchierboom:
> Could I ask when do we decide to edit the test generator and when do we decide to edit the canonical data?

We have to decide whether a change is beneficial to _all_ (or most) tracks or just applies to C#. For the former we should edit the canonical data, for the latter the test generator.

> My assumption is that you can't throw exceptions in canonical-data so any test case that 'works' should be added there, is that right?

No we actually have an expected `error` value for error cases.

exercises/practice/sieve/.meta/Example.cs
ErikSchierboom: Maybe change to the above? If you disagree, I'm fine with merging _as is).

yzAlvin: omg I was trying to do that but didn't realise why it wasn't letting me, I was missing the <int> ! :( thanks

ErikSchierboom (approved):
Excellent!

Issue #1567, opened 2021-08-20
Check docs are up to date


ErikSchierboom:
Please check if your documentation files are still up-to-date.

The key documentation files to check are:

- `docs/ABOUT.md`
- `docs/INSTALLATION.md`
- `docs/LEARNING.md`
- `docs/RESOURCES.md`
- `docs/TESTS.md`
- `exercises/shared/.docs/help.md`
- `exercises/shared/.docs/tests.md`

There might be more.

## Link check report

To help identify invalid links, we've automatically checked the links of _all_ `*.md` files in this repo.
This is the report of that check:

<pre style="color:#bbb;white-space:pre-wrap;word-wrap:break-word;overflow-wrap:break-word">📝 Summary
---------------------
🔍 Total..........792
✅ Successful.....765
⏳ Timeouts.........1
🔀 Redirected.......0
👻 Excluded.........0
🚫 Errors..........26

Errors in ./exercises/practice/sublist/.docs/instructions.append.md
<span style="color:#ff5555">✗ https://msdn.microsoft.com/en-us/library/system.icomparable(v=vs.110).aspx (HTTP status client error (403 Forbidden) for url (https://msdn.microsoft.com/en-us/library/system.icomparable(v=vs.110).aspx))</span>

Errors in ./exercises/practice/simple-linked-list/.docs/instructions.append.md
<span style="color:#ff5555">✗ https://msdn.microsoft.com/en-us/library/9eekhta0(v=vs.110).aspx (HTTP status client error (403 Forbidden) for url (https://msdn.microsoft.com/en-us/library/9eekhta0(v=vs.110).aspx))</span>

Errors in ./exercises/practice/diffie-hellman/.docs/instructions.append.md
<span style="color:#ff5555">✗ https://msdn.microsoft.com/en-us/library/system.numerics.biginteger(v=vs.110).aspx (HTTP status client error (403 Forbidden) for url (https://msdn.microsoft.com/en-us/library/system.numerics.biginteger(v=vs.110).aspx))</span>

Errors in ./exercises/practice/allergies/.docs/instructions.append.md
<span style="color:#ff5555">✗ https://msdn.microsoft.com/en-us/library/6a71f45d.aspx (HTTP status client error (403 Forbidden) for url (https://msdn.microsoft.com/en-us/library/6a71f45d.aspx))</span>

Errors in ./exercises/concept/faceid-2/.docs/instructions.md
[...]

PR #1568, opened 2021-08-21
update links

yzAlvin:
#1567 

Replaced links in the error report.

None of the links found were 'broken' (except 1), they just redirected to a different page so I updated the link to the place it redirected me. 

Also where I could I replaced link to more recent documentation


ErikSchierboom (approved):
This is great!

PR #1578, opened 2021-09-03
Fixed requirements for Task 4

henry-js:
Task 4 asks the user to format the string differently to what the unit test requires

image
image


henry-js:
Not sure what the desired string should look like

Issue #1445, opened 2020-04-02
Implement new Concept Exercise: extension-methods


ErikSchierboom:
This issue describes how to implement the `extension-methods` concept exercise for the C# track.

## Getting started

**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before diving into the implementation, please read up on the following documents:

- [The features of v3](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md).
- [Rationale for v3](https://github.com/exercism/v3/blob/master/docs/rationale-for-v3.md).
- [What are concept exercise and how they are structured?](https://github.com/exercism/v3/blob/master/docs/features-of-v3.md)

Please also watch the following video:

- [The Anatomy of a Concept Exercise](https://www.youtube.com/watch?v=gkbBqd7hPrA).

## Goal

The goal of this exercise is to teach the student the Concept of Extension Methods in C#.

## Learning objectives

- Know what extension methods are.
- Know how to define extension methods on reference and value types..

## Out of scope

- Assemblies.

## Concepts

This Concepts Exercise's Concepts are:

- `extension-methods`: know what extension methods are; know how to define extension methods on reference and value types.

## Prequisites

This Concept Exercise's prerequisites Concepts are:

- `classes`: know what instance methods are.
- `namespaces`: know what namespaces are; know how to import namespaces.

Any data types used in this exercise (e.g. `strings`) should also be added as prerequisites.

### Hints

- [Extension methods][extension-methods]: how to define extension methods.
- [Namespaces][namespaces]: how to define an import namespaces.

### After

- [Extension methods][extension-methods]: how to define extension methods.
- [Namespaces][namespaces]: how to define an import namespaces.

## Representer

This exercise does not require any specific representation logic to be added to the [representer][representer].

## Analyzer

This exercise does not require any specific analyzer logic to be added to the [analyzer][analyzer].

## Implementing

To implement this exercise, please [follow these instructions](https://github.com/exercism/v3/blob/master/languages/csharp/reference/implementing-a-concept-exercise.md).

## Help

If you have any questions while implementing the exercise, please post the questions as comments in this issue.

[how-to-implement-a-concept-exercise]: https://github.com/exercism/v3/blob/master/docs/maintainers/generic-how-to-implement-a-concept-exercise.md
[...]

PR #1585, opened 2021-09-05
Implement new concept exercise for extension methods

yzAlvin:
I have followed the instructions [here](https://github.com/exercism/v3/blob/main/languages/csharp/reference/implementing-a-concept-exercise.md) and read the docs listed in the issue #1445

This adds a new concept exercise focused around teaching extension-methods

The problem is to extend the IEnumerable<int> interface with a Sum, Product, Mean, Median and Mode method, as they could be commonly used in a statistical context

Please let me know if there is anything I should revise


yzAlvin:
Actually I don't think this is a very good exercise for learning extension methods

This only goes into how to write one but doesn't actually use them except for the tests

But I am curious if this could be revised to be practice for some other thing?

ErikSchierboom:
I don't have much time at the moment, but I'll try and look at it this week!

ErikSchierboom (commented):
First of all, thank you so much for working on this! It's greatly appreciated.

> Actually I don't think this is a very good exercise for learning extension methods
> 
> This only goes into how to write one but doesn't actually use them except for the tests

This is totally valid point. We should try to design the exercise such that we encourage/force people to use extension methods too (writing them is the more important part here I feel). One way we do this in Concept Exercises is to have later tasks depend on functionality that is implemented in a previous task.

The main thing about the current exercise that I'm unsure of is the task: the statistical analysis part of it. In general, we try to shy away from maths as much as possible (which is not always possible), especially for Concept Exercises as it is often hard to come up with a nice story/theme around math problems. 

Maybe we could brainstorm a bit on what could make a good story for this exercise?

Personally I think that maybe extending the `string` class could be useful. We could even save us a lot of trouble by re-using the existing `log-levels` exercise story/theme, which is something that we do a lot actually. This exercise already deals with strings and has a student extract parts of information from the string, which could work well for extension methods.

What do you think?

yzAlvin:
Yeah I agree :D, I'll try and design an exercise around extending log levels for extension methods.

ErikSchierboom:
Awesome! Thanks.

yzAlvin:
😬 😬 
That was harder than I thought it would be, I am not creative at all..

I couldn't think of anything else other than re-using the idea of a log level aha, and tbh I'm still not sure this makes for a good concept exercise, but happy to hear what others think

exercises/concept/log-analysis/.docs/instructions.md
```diff
@@ -0,0 +1,35 @@
...
+log.LogLine(); // => returns "20"
+
+log.Truncate(10); // => returns "[Error]: {Line 20} - 'Critical error found'"
+
+log.Truncate(5); // => returns "Error - 20"
+```
+
+## 1. Implement the extension method WordCount
+
+Implement the `.Word()` method to return the total of all words in the string. A 'word' in this case is just any string separated by a space.
```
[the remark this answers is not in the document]
ErikSchierboom: Each task should contain an example of a invocation and its return value.

ErikSchierboom (commented):
> That was harder than I thought it would be, I am not creative at all..
> 
> I couldn't think of anything else other than re-using the idea of a log level aha, and tbh I'm still not sure this makes for a good concept exercise, but happy to hear what others think

Yeah, I struggled with this too at first. But once you've done a couple, things _do_ become easier.

I've been giving this some more thought, and I _think_ I've found a nice way to have students define and use the extension method. What if we would replace the `WordCount` extension method with a `SubstringAfter` method, which can be defined like:

```csharp
public static string SubstringAfter(this string str, string delimiter)
{
    return str.Substring(str.IndexOf(delimiter) + 1);
}
```

and also have the students define a `SubstringBetween` method:

```csharp
public static string SubstringBetween(this string str, string start, string stop)
{
    return str.Substring(str.IndexOf(start) + 1, str.IndexOf(stop) -1);
}
```

If we make this the very first two tasks of the exercise (like WordCount is the first task now), we could then reasonably expect the student to use these methods for the `Message` and `LogLevel` methods. We can then drop the `Truncate` method I think, which would have become unnecessary. What do you think?

p.s. it is totally normal for a Concept Exercise to go through several iterations before arriving at a satisfactory implementation

ErikSchierboom:
@yzAlvin What do you think about https://github.com/exercism/csharp/pull/1585#pullrequestreview-755128883?

yzAlvin:
I like it :) I wanted an extension method that had some arguments and that achieves it

ErikSchierboom:
@yzAlvin Great!

config.json
```diff
@@ -420,6 +420,17 @@
...
+      },
+      {
+        "slug": "log-analysis",
+        "name": "Log Analysis",
+        "uuid": "4a940e54-eda8-4500-bcb2-25d83af01fb1",
+        "concepts": ["extension-methods"],
+        "prerequisites": [
+          "collections",
+          "strings",
+          "booleans"
```
ErikSchierboom, line 432: I think we just need `strings` as a prerequisite with the current setup.

```suggestion
          "strings"
```

ErikSchierboom (commented):
Love it! I added some minor comments, nothing big. Just some tweaks and then we're ready to merge!

yzAlvin:
Will add the things about this and namespaces soon, is it alright that the test case for SubstringAfter is generic and unrelated to the story?

ErikSchierboom:
> is it alright that the test case for SubstringAfter is generic and unrelated to the story?

Yeah sure.

yzAlvin:
@ErikSchierboom alright, thanks for all your input, feel free to point out anything else :D

exercises/concept/log-analysis/.docs/instructions.md
```diff
@@ -0,0 +1,53 @@
...
+Implement the (_static_) `LogAnalysis.SubstringAfter()` extension method, that takes in some string delimeter and returns the substring after the delimiter.
+
+```csharp
+var log = "[INFO]: File Deleted.";
+log.SubstringAfter(": "); // => returns "File Deleted."
+```
+
+## 2. Allow retrieving the string in between two substrings
+
+On further inspection, you see that the log level is always located between two substrings. As you're also anticipating having to extract the log level sometime in the near future, you decide to create a another helper method to help you with that.
```
ErikSchierboom, line 28:
```suggestion
On further inspection, you see that the log level is always located between square brackets (`[` and `]`). As you're also anticipating having to extract the log level sometime in the near future, you decide to create a another helper method to help you with that.
```

ErikSchierboom (approved):

concepts/extension-methods/about.md
```diff
@@ -0,0 +1,22 @@
+# About
+
+[Extension methods][extension-methods] allow adding methods to existing types without creating a new derived type, recompiling, or otherwise modifying the original type.
+
+Extension methods are brought into scope at the namespace level.
+
+Extension methods are static methods, but they're called as if they were instance methods on the extended type. It achieves this by using `this` before the type, indicating the instance we put the `.` on is passed as the first parameter. For client code, there's no apparent difference between calling an extension method and the methods defined in a type.
+
+```csharp
+public static int WordCount(this string str)
```
ErikSchierboom, line 10: It might be nice to be a bit more explicit here and give the class a namespace, and then have a second code sample indicating another file that has a `using` statement for importing the extension method. That might make it a bit more explicit.

ErikSchierboom (approved):
Excellent! I've left two small comments, but I'd be totally fine with merging this as is and addressing them in a follow-up PR. Whatever you prefer.

ErikSchierboom (approved):

ErikSchierboom:
Merged, thanks! This should appear on the website shortly. Great work @yzAlvin!

Issue #1593, opened 2021-09-08
Wizards and warriors not fully tested


joce:
Nothing is currently testing the damage done by a warrior; only for a wizard.

I wrote my solution in the online editor and left the `DamagePoints()` method of the Warrior with the default implementation (`throw new NotImplementedException("Please implement the Warrior.DamagePoints() method");`) and all tests pass and I can submit the code as is.

ErikSchierboom:
Closed by #1595

PR #1595, opened 2021-09-08
Add tests for warrior damage

joce:
New tests have been added to test the warrior's damage, according to the specs for the exercise.

This was previously reported as issue [#1593](https://github.com/exercism/csharp/issues/1593)


ErikSchierboom (approved):
Well spotted! Thanks for fixing this.

PR #1610, opened 2021-09-12
Change exception type for weighing machine example

### LINKED ISSUE #1609: Mismatch between stated expectations and tests for Weighing Machine example.
The text of the exercise states:

**2. Ensure that a negative input weight is rejected**
Add validation to the `WeighingMachine.InputWeight` property to throw an `ArgumentOutOfRangeException` when trying to set it to a negative weight
```csharp
var wm = new WeighingMachine();
wm.InputWeight = -10m; // Throws an ArgumentException
```

So the statement and the example code don't match (the statement says `ArgumentOutOfRangeException` whereas the code snippet says `ArgumentException`), and the actual tests checks for an `ArgumentException`.

We should just use `ArgumentOutOfRangeException` everywhere.

joce:
Fixes #1609
Replace the exception checked in the tests and the exception stated in the code example in the instruction to be an `ArgumentOutOfRangeException`

### DIFF
--- a/exercises/concept/weighing-machine/.docs/instructions.md
+++ b/exercises/concept/weighing-machine/.docs/instructions.md
@@ -48,7 +48,7 @@ Add validation to the `WeighingMachine.InputWeight` property to throw an `Argume
 
 ```csharp
 var wm = new WeighingMachine();
-wm.InputWeight = -10m; // Throws an ArgumentException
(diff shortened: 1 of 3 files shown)


ErikSchierboom (approved):
Yep, this is more idiomatic!

PR #1616, opened 2021-09-15
Removed unused property on the Manager class

CodeVelo:
The Activity property on the Manager class is never used in the tests. Also, the instructions don't make mention of it as one of the tasks that needs to be implemented. I suggest removing the property to add clarity to the Manager class and associated tests.


ErikSchierboom (commented):
Well spotted! Could you also update the exemplar file: https://github.com/exercism/csharp/blob/main/exercises/concept/football-match-reports/.meta/Exemplar.cs#L76?

CodeVelo:
Done.

ErikSchierboom (approved):
Brilliant!

PR #1620, opened 2021-09-17
Change the return type for TestTrack.Race() to void

CodeVelo:
The instructions are not clear on what value should be returned. Most users are returning `DistanceTravelled`. However, the unit tests do not check the return value.

I suggest changing the return type to void.


ErikSchierboom (approved):

ErikSchierboom:
Thanks!

PR #1623, opened 2021-09-19
The count should be 4 as there are already 3 values in Dictionary.

18-F-cali:
see test #2 
```cs

    [Fact]
    public void Existing_dictionary_count_is_3()
    {
        var prePopulated = DialingCodes.GetExistingDictionary();
        Assert.Equal(3, prePopulated.Count);
    }
```
and problem description
image

### DIFF
--- a/exercises/concept/international-calling-connoisseur/InternationalCallingConnoisseurTests.cs
+++ b/exercises/concept/international-calling-connoisseur/InternationalCallingConnoisseurTests.cs
@@ -157,11 +157,11 @@ public class DialingCodesTest
     }
 
     [Fact(Skip = "Remove this Skip property to run this test")]
-    public void Update_country_name_in_dictionary_for_non_existent_country_count_is_3()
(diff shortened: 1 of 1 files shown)


exercises/concept/international-calling-connoisseur/InternationalCallingConnoisseurTests.cs
```diff
@@ -157,11 +157,11 @@ public void Update_country_name_in_dictionary_91_is_India()
     }
 
     [Fact(Skip = "Remove this Skip property to run this test")]
-    public void Update_country_name_in_dictionary_for_non_existent_country_count_is_3()
+    public void Update_country_name_in_dictionary_for_non_existent_country_count_is_4()
```
ErikSchierboom, line 160: I think you missed the bit in the instructions that said:

> If the dialing code does not exist in the dictionary then the dictionary remains unchanged.

This is why the count should remain 3.

PR #1625, opened 2021-09-21
Fix little markdown typo

joce:
The `*` doesn't need to be escaped.


ErikSchierboom:
I'm using Prettier to format all Markdown files, and Prettier _does_ escape the `*` character, so I'd like to keep this behavior. At some point I'll add this to CI too (likely).

PR #1627, opened 2021-09-21
Fix incorrect code sample in instructions

joce:
The text references `TestMultiplication` whereas the code that follows was showing the use of `Multiply` (even though the shown expected result is the one that would be provided by `TestMultiplication`


ErikSchierboom (approved):
👍

Issue #1592, opened 2021-09-08
Nullibility concept exercise - Tim From Marketing - format in instructions doesn't match format in tests


Willz42:
https://exercism.org/tracks/csharp/exercises/tim-from-marketing / https://github.com/exercism/csharp/tree/782086650efff991f61184f3c55bc173ad28955b/exercises/concept/tim-from-marketing

Instructions ask for the output to be formatted like: **"[id] - [name] - [DEPARTMENT]"** but in the test's it's looking for **"[17] Ryder Herbert - MARKETING"** (no hyphen between Id and name)

I'm happy to have a go at a PR to fix update either the tests or the instructions if appropriate.

ErikSchierboom:
Good catch! I've just looked at the instructions and I think we should update them to something like:

```
Employee badge labels are formatted as follows: `"[<id>] - <name> - <DEPARTMENT>"`
```

I think the brackets help make it clear that they are placeholders, separate from the square brackets. What do you think?

Willz42:
I think that's good - and would we change the tests to match, or make it:
`Employee badge labels are formatted as follows: `"[<id>] <name> - <DEPARTMENT>"`
To match the current test format?

ErikSchierboom:
No the tests are fine. With ``"[<id>] - <name> - <DEPARTMENT>"`` I mean: replace `<id>` with the ID, `<name>` with the name and `<DEPARTMENT>` with the uppercase department.

PR #1660, opened 2021-09-29
Fix Tim from Marketing instructions to reflect the correct task output

giorgospetkakis:
In Tim from Marketing, this model pattern is given:

> "[id] - [name] - [DEPARTMENT]"

with this result:

```csharp
// => "[734] - Ernest Johnny Payne - STRATEGIC COMMUNICATION"
```

I suggest that the model pattern is changed to:

> "[id] - name - DEPARTMENT"

to reflect the absence of square brackets around the latter two terms.

Closes https://github.com/exercism/csharp/issues/1592


ErikSchierboom (approved):

ErikSchierboom:
Thanks!

Issue #1663, opened 2021-09-30
robot-name: Add edge case for name


FaisalAfroz:
Currently all test pass even if the Robot does not generate a name with edge cases.

PR #1664, opened 2021-09-30
[Robot-Name]: Added edge cases in Name

FaisalAfroz:
Earlier all test case passed even if the Robot did not generated the edge case 'A' or 'Z' in the name.
Since to generate random letters most use 'Next()' method of 'Random' class and it's overload method  has inclusive 'minvalue' and exclusive 'maxvalue' , so we need to check those cases.

Fixes: #1663


ErikSchierboom (approved):

ErikSchierboom:
Thanks!

PR #1665, opened 2021-09-30
Improve calculator-conundrum concept exercise

### LINKED ISSUE #1462: calculator-conundrum exercise (exceptions) hints.md text is required
Add text to _hints.md_ for calculator-conundrum.

Also remove the _Hints_ and _After_ sections from the _design.md_ document.

### LINKED ISSUE #1641: calculator-conundrum: move class that shouldn't be edited into a separate file
There is a class in the `calculator-conundrum` exercise's stub that we indicate the student shouldn't edit. We now have functionality to support editor-only (readonly) classes, which means that we can:

- Extract the class into a separate file
- Add that new file's name to the exercise's `.meta/config.json` file's `files.editor` key (which might not exist yet)

sanderploegsma:
- Moves read-only code from the stub implementation to a separate file that is marked as editor file in `.meta/config.json`, fixes #1641 
- Moves the hints from `.meta/design.md` to general hints in `.docs/hints.md`, fixes #1462 
- Updates the instructions to better match the exercise specification and adds missing usage examples


exercises/concept/calculator-conundrum/.docs/hints.md
```diff
@@ -1,2 +1,13 @@
 # Hints
 
+## General
+
+- [Exceptions][exceptions]: basic introduction to exceptions.
+- [Using exceptions][using-exceptions]: how to use exceptions.
+- [Exception handling][exception-handling]: how to catch exceptions.
```
ErikSchierboom, line 7: It would be great if the hints could be linked to individual tasks. Take for example [task 2](https://github.com/exercism/csharp/blob/2eaeb0a8c8b38e0cff3f762bd85d68d6c727cacb/exercises/concept/calculator-conundrum/.docs/instructions.md#2-handle-illegal-operations), which requires one to throw exceptions. Instead of having the throwing of exceptions as a general hint, you could move it to a task-specific hint:

```markdown
## 2. Handle illegal operations

- You need to [throw an exception][creating-and-throwing-exceptions] here.

The same goes for catching exceptions, which belongs to task 3.
```

exercises/concept/calculator-conundrum/.meta/config.json
```diff
@@ -1,21 +1,11 @@
 {
   "blurb": "Learn about error handling by working on a simple calculator.",
-  "contributors": [
-    "valentin-p",
-    "yzAlvin"
-  ],
-  "authors": [
-    "archrisV"
-  ],
+  "contributors": ["valentin-p", "yzAlvin"],
```
ErikSchierboom, line 3: You can add yourself as a contributor.

ErikSchierboom (commented):
Great work! While reading the instructions, I found two more things that should be fixed:

1. Task 2 mentions that "This functionality and handling of operations should be contained in a try block." I don't think this is needed yet at this point and that this bit of text instead belongs to task 3.
2. Task 2 does not have an example.

Would you be interested in fixing those issues too?

sanderploegsma:
> Task 2 mentions that "This functionality and handling of operations should be contained in a try block." I don't think this is needed yet at this point and that this bit of text instead belongs to task 3.

@ErikSchierboom I actually think that the instructions should not mention using a try block at all, and instead mention that in the hints for task 3. Otherwise I think the instructions are giving away too much, don't you think?

ErikSchierboom:
> I actually think that the instructions should not mention using a try block at all, and instead mention that in the hints for task 3. Otherwise I think the instructions are giving away too much, don't you think?

@sanderploegsma Ah yes, you're totally right. That's _much_ better.

exercises/concept/calculator-conundrum/.docs/instructions.md
```diff
@@ -22,7 +22,15 @@ The main method for implementation in this task will be the (_static_) `SimpleCa
...
+
+```csharp
+SimpleCalculator.Calculate(100, 10, "-"); // => throws ArgumentOutOfRangeException
+
+SimpleCalculator.Calculate(8, 2, ""); // => throws ArgumentException
+
+SimpleCalculator.Calculate(58, 6, null); // => throws ArgumentNullException
+```
 
 ## 3. Handle the thrown DivideByZero exceptions
```
ErikSchierboom, line 35: Maybe rename this to:

```suggestion
## 3. Handle division by zero
```

Note: the hints also need to be updated accordingly

sanderploegsma: Good suggestion. I also think we should not mention `DivideByZeroException` in the task instruction and let the student figure that out themselves. Maybe we should help them out by pointing to it from the hints?

ErikSchierboom: Yeah I like that a lot. Great suggestions!

exercises/concept/calculator-conundrum/.docs/hints.md
```diff
@@ -1,2 +1,20 @@
...
+- [Exceptions][exceptions]: basic introduction to exceptions.
+- [Using exceptions][using-exceptions]: how to use exceptions.
+
+## 2. Handle illegal operations
+
+- You need to [throw an exception][creating-and-throwing-exceptions] here.
+
+## 3. Handle errors when dividing by zero
+
+- You need to [catch an exception][exception-handling] here to handle the [`DivideByZeroException`][divide-by-zero-exception].
```
ErikSchierboom, line 14: Tiny suggestion:

```suggestion
- You need to [catch an exception][exception-handling] here to handle the [`DivideByZeroException`][divide-by-zero-exception] that is thrown when dividing by zero.
```

ErikSchierboom (approved):
One tiny nit. This is so much better!

ErikSchierboom:
Sweet!

PR #1666, opened 2021-09-30
hyperia-forex: Add missing test cases and fix Exemplar.cs

### LINKED ISSUE #1652: hyperia-forex: add missing tests
The three existing tests work with the equality operators but do not test any of the others implemented in sub-problems 2-6.

sanderploegsma:
This adds test cases to cover all tasks of the exercise, instead of just the `==` operator. To get all of the tests to pass two changes were required to the exemplar implementation:

1. The `!=` operator was not correctly overridden: it actually calls itself causing a stack overflow.
2. The `/` operator was overloaded to support both `new CurrencyAmount(100, "USD") / 2m` and `2m / new CurrencyAmount(100, "USD")`. The first version is part of the specification and makes total sense. However, the second version is not mentioned in the instructions anywhere, and in my opinion also doesn't make any sense: what would this even return?

This fixes #1652.

### DIFF
--- a/exercises/concept/hyperia-forex/HyperiaForexTests.cs
+++ b/exercises/concept/hyperia-forex/HyperiaForexTests.cs
@@ -2,7 +2,7 @@ using System;
 using Xunit;
 using Exercism.Tests;
 
-public class OperatorOverloadingTests
+public class HyperiaForexTests
 {
     [Fact]
     public void Equality_true()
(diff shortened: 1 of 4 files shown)


exercises/concept/hyperia-forex/HyperiaForexTests.cs
```diff
@@ -17,8 +17,117 @@ public void Equality_false()
...
+    [Fact(Skip = "Remove this Skip property to run this test")]
+    public void GreaterThan_different_currencies()
+    {
+        Assert.Throws<ArgumentException>(() => new CurrencyAmount(60, "HD") > new CurrencyAmount(55, "USD"));
+    }
+
+    [Fact(Skip = "Remove this Skip property to run this test")]
+    public void Addition_same_currencies()
+    {
+        Assert.Equal(new CurrencyAmount(155, "HD"), new CurrencyAmount(55, "HD") + new CurrencyAmount(100, "HD"));
```
sanderploegsma, line 82: Note that these test cases are very basic and only cover about as much as is shown in the instructions. I feel like this exercise is a perfect candidate to use property-based testing instead, so to test the `+` operator overload happy cases for instance, we could test:

- Commutativity: `x + y = y + x`
- Associativity: `(x + y) + z = x + (y + z)`

IMO these tests do a better job of verifying the exercise implementation and may even trigger students to think about edge-cases they did not consider yet. 

I'm not sure if property-based testing is used at all in Exercism test suites, but it could be worth considering to do in the future. What do you think?

ErikSchierboom:
> However, the second version is not mentioned in the instructions anywhere, and in my opinion also doesn't make any sense: what would this even return?

Yeah, let's remove this.

exercises/concept/hyperia-forex/HyperiaForexTests.cs
sanderploegsma: I gave it a shot, have a look to see if you think it's still readable and understandable for students.

sanderploegsma:
I just realized that adding these test cases will mean that you are no longer able to run the tests half-way through the exercises, because the test suite won't compile until all the operators on `CurrencyAmount` are defined. I haven't worked much with the V3 online editor yet so I'm not sure whether running the tests before finishing all the tasks is possible/recommended, but I wanted to point it out nonetheless.

ErikSchierboom:
@sanderploegsma Yeah this is a problem that we have with several exercises, and not something that is easy to fix. It's fine for now.

Merging this is blocked by https://github.com/exercism/csharp-test-runner/pull/67 BTW. I just found out that I don't unskip non-`Fact` tests. Whoops! Once that is merged (should be a couple of hours), I'll merge this.

sanderploegsma:
Ah that's a good catch! I did check the `test.ps1` script to make sure that it was able to unskip the non-`Fact` tests (which it is) but I didn't think to check the test runner.

ErikSchierboom (approved):
Test runner update has been deployed. Thanks!

Issue #1659, opened 2021-09-29
extension-methods: improve about.md


ErikSchierboom:
The `about.md` file of the `extension-methods` concept should be updated to make it clear that in order to use an extension method, its namespace must in a `using` statement.

jodihamann:
Looks like the fix for this may have been merged a few days ago, so I didn't want to comment on that already completed PR.

It's a small change, but what the about.md repeatedly refers to as `using statements` are actually `using directives`. It may be easier for learners to find additional details if that's changed.

I'm happy to put in a PR if it's welcome.

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-directive

ErikSchierboom:
@jodihamann Yes, a PR would be much appreciated!

henry-js:
Can we close this issue if changes have been merged?

PR #1667, opened 2021-09-30
add more detail about namespaces in extension-methods

yzAlvin:
#1659

### DIFF
--- a/concepts/extension-methods/about.md
+++ b/concepts/extension-methods/about.md
@@ -17,7 +17,7 @@ namespace MyExtensions
 // => 2
 ```
 
-Extension methods are brought into scope at the namespace level. This means that if you are in different namespace to the one the extension method is defined in, it's namespace must in a using statement first; For example, if we wanted to use the above example in our code, we would first need a `using MyExtensions` statement. If you are in the same namespace as the one the extension method is defined in, you can use the extension methods without a using statement.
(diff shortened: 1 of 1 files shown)


concepts/extension-methods/about.md
```diff
@@ -2,21 +2,24 @@
...
+        return str.Split().Length;
+    }
 }
 
 "Hello World".WordCount();
 // => 2
 ```
 
-A well-known example of extension methods are the [LINQ][linq] standard query operators that add query functionality to the existing IEnumerable types.
+Extension methods are brought into scope at the namespace level. This means that if you are in different namespace to the one the extension method is defined in, it's namespace must in a using statement first; For example, if we wanted to use the above example in our code, we would first need a `using MyExtensions` statement. If you are in the same namespace as the one the extension method is defined in, you can use the extension methods without a using statement.
```
ErikSchierboom, line 20:
```suggestion
Extension methods are brought into scope at the namespace level. This means that if you are in different namespace to the one the extension method is defined in, it's namespace must be in a `using` statement first; For example, if we wanted to use the above example in our code, we would first need a `using MyExtensions` statement. If you are in the same namespace as the one the extension method is defined in, you can use the extension methods without a `using` statement.
```

ErikSchierboom (approved):
One tiny nit.

ErikSchierboom:
Thanks a lot!

PR #1668, opened 2021-09-30
Interest is interesting add Interest task

### LINKED ISSUE #1634: interest-is-interesting: add extra task for `Interest` method
To reduce the number of representations, we should add an `Interest` method that returns the actual interest based on the balance and its interest rate:

```csharp
public static decimal Interest(decimal balance)
{
    return Math.Abs(balance) * ((decimal)InterestRate(balance) / 100);
}
```

This should become the second task in the instructions.

yzAlvin:
Closes #1634


ErikSchierboom (approved):
This is perfect! Thanks

PR #1669, opened 2021-10-01
football-match-reports: Move read-only code from implementation stub to editor-only file

### LINKED ISSUE #1640: football-match-reports: move classes that shouldn't be edited into a separate file
There are some classes in the `football-match-reports` exercise's stub that we indicate the student shouldn't edit. We now have functionality to support editor-only (readonly) classes, which means that we can:

- Extract those classes into a separate file (I recommend just one file)
- Add that new file's name to the exercise's `.meta/config.json` file's `files.editor` key (which might not exist yet)

sanderploegsma:
This fixes #1640.

### DIFF
--- a/exercises/concept/football-match-reports/.meta/Exemplar.cs
+++ b/exercises/concept/football-match-reports/.meta/Exemplar.cs
@@ -71,32 +71,3 @@ public static class PlayAnalyzer
         return description;
     }
 }
-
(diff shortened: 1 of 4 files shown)


ErikSchierboom (approved):

ErikSchierboom:
Thanks!

PR #1670, opened 2021-10-01
football-match-reports: revise exercise tasks

### LINKED ISSUE #1636: football-match-report: add test for shirt number in AnalyzeOffField method
The `football-match-report`'s instructions mention that the `AnalyzeOffField()` method should treat any `int` as a shirt number and return the `AnalyzeOnField()` text. There are no tests to verify this behavior though.

### LINKED ISSUE #1637: football-match-reports: add more tasks
The third task describes the various off-field cases. It would be nice if these were split up into individual tasks: one for integers, one for strings, one for incidents, one for injuries and one for managers.

We could then merge the last task with the new task for managers.

### LINKED ISSUE #1638: football-match-reports: explain that the Injury class requires a different treatment
The instructions mention that incidents should be treated a certain way, but then the `Injury` class is treated differently (it is specified in the comments). It would be much better if this is explained in a specific task. See #1637

### LINKED ISSUE #1639: football-match-reports: have manager task make more sense
Currently, there is a special case for a manager without a name. This makes relatively little sense (I expect _all_ managers to have a name), but maybe there can be a second property on the `Manager` class (e.g. the club?) that can be empty, or maybe a boolean or integer (easier default values).

sanderploegsma:
This revises the tasks of the exercise so that each requirement of the `PlayAnalyzer.AnalyzeOffField()` method is introduced in a separate task.

### Task 3: type guard on built-in types

This task requires the student to switch on the built-in types `int` and `string`, and provide a default case to match on any other type. I deliberately did not split up these two types into their own tasks to avoid people going in the wrong direction using something like this:

``` csharp
if (report is int count)
{
    return $"There are {count} supporters at the match.";
}
```

To make the instructions a little bit less confusing I changed the meaning of `int` to be the number of supporters at the match instead of the shirt number (since shirt numbers are not "off the field"). Also, a `string` now represents an announcement through the stadium's PA system instead of "free form text", which speaks a little bit more to the imagination.

Throwing the `ArgumentException` in the default case was not mentioned anywhere in the instructions so I added it here. This should not be a problem since the same pattern is also used in Task 2. Doing so here also gives the added bonus that the tests can be run before finishing the exercise because the implementation code otherwise probably doesn't compile.

### Task 4: type guard on inherited classes

This task requires the student to switch on the `Incident` type and subclasses thereof. The instructions did not explicitly mention that injuries should be treated differently from other incidents, so I tried to make this more explicit.

### Task 5: type guard with `when` keyword

This final task requires the student to switch on the `Manager` type with two cases: one where the manager's club is known and one where it isn't. Because the `Manager.Club` property is a `string?` I added the `nullability` concept as a prerequisite.

## Linked issues

Fixes #1636.  
Fixes #1637.  
Fixes #1638.  
Fixes #1639.

### DIFF
--- a/config.json
+++ b/config.json
@@ -373,7 +373,7 @@
         "name": "Football Match Reports",
         "uuid": "eb9faaec-9754-4df1-8eb1-09ca0ad94665",
         "concepts": ["switch-statements"],
-        "prerequisites": ["classes", "inheritance"]
+        "prerequisites": ["classes", "inheritance", "nullability"]
       },
       {
         "slug": "beauty-salon-goes-global",
@@ -413,9 +413,7 @@
         "name": "Log Analysis",
         "uuid": "4a940e54-eda8-4500-bcb2-25d83af01fb1",
         "concepts": ["extension-methods"],
-        "prerequisites": [
-          "strings"
-        ]
+        "prerequisites": ["strings"]
       }
     ],
     "practice": [
--- a/exercises/concept/football-match-reports/.docs/hints.md
+++ b/exercises/concept/football-match-reports/.docs/hints.md
@@ -13,7 +13,7 @@
 ## 3. Extend the coverage to include off field activity
 
 - [Pattern matching on types][switch-pattern-matching] is the key to this task.
-- Use the [`default`][default] statement to match any other type
+- Use the [`default`][default] statement to match any other type.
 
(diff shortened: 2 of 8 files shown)


ErikSchierboom (approved):
This is great!

Issue #1645, opened 2021-09-24
weighing-machine: consider tweaking design


ErikSchierboom:
The `USWeight` struct is used to have students work with initializing a getter only property from the constructor. The design of the exercise with a separate `USWeight` struct feels a bit convoluted though. The issues are:

1. It feels weird to return a new struct each time a property is accessed
2. The rounding of pounds feels weird.

I also don't particularly like the fact that the input weight is a decimal, whereas the display weight is an int.

I wonder if we could tweak the design a bit to get rid of the `USWeight` struct and instead pass an argument to the `WeighingMachine` constructor itself. I was thinking of maybe passing in a precision argument to the constructor, which would then be set to a getter-only property. We could then change the display weight to return a string an have that string use the passed in precision.

PR #1672, opened 2021-10-01
Updated  'Weighing Machine ' exercise.

### LINKED ISSUE #1629: Wrong Expected Values in WeighingMachine
```cs
[Fact]
    public void Get_us_display_weight_pounds()
    {
        var wm = new WeighingMachine();
        wm.InputWeight = 60m;// no default unit supplied hence assumed kg
        Assert.Equal(132, wm.USDisplayWeight.Pounds);// 60 kg == 132.277 pounds  // **precision lost** 
    }

    [Fact]
    public void Get_us_display_weight_ounces()
    {
        var wm = new WeighingMachine();
        wm.InputWeight = 60m;// no default unit supplied hence assumed kg
        Assert.Equal(4, wm.USDisplayWeight.Ounces); // **60 kg  = 2116.44 ounces** 
    }

    [Fact]
    public void Input_pounds_and_get_us_display_weight_pounds()
    {
        var wm = new WeighingMachine();
        wm.Units = Units.Pounds;
        wm.InputWeight = 175.5m;
        Assert.Equal(175, wm.USDisplayWeight.Pounds); // **precision lost**
    }

    [Fact]
    public void Input_pounds_and_get_is_display_weight_ounces()
    {
        var wm = new WeighingMachine();
        wm.Units = Units.Pounds;
        wm.InputWeight = 175.5m;
        Assert.Equal(8, wm.USDisplayWeight.Ounces); // **175.5 pounds == 2808 ounces**
    }
```

18-F-cali:
Here I did the following.
Updated the hints and instructions.
I also updated the design file, `## Analyser`, which might need special attention. 
I replaced the Exemplar with @ErikSchierboom 's solution. 
I corrected the language stub file by removing unnecessary instructions.
I updated the test according to the design, which might also need attention if current test are not sufficient.
I also updated the config file with my exercism handle :yum:
closes #1629 and #1645

### DIFF
--- a/exercises/concept/weighing-machine/.docs/hints.md
+++ b/exercises/concept/weighing-machine/.docs/hints.md
@@ -5,25 +5,29 @@
 - [Properties][docs.microsoft.com-properties]
 - [Using Properties][docs.microsoft.com-using-properties]
 
-## 1. Allow the weight to be set on the weighing machine
+## 1. Allow the weighing machine to have a precision.
 
-A property with a private [backing field][docs.microsoft.com-properties-with-backing-fields] is appropriate here.
+A get only property that can be intialised only from the constructor.
 
-## 2. Ensure that a negative input weight is rejected.
+## 2. Allow the weight to be set on the weighing machine
 
-Add [validation][stackoverflow.com-validating-properties] to the `InputWeight`'s `set` accessor to throw an exception.
+A property with a private [backing field][docs.microsoft.com-properties-with-backing-fields] is appropriate here
 
-## 3. Allow the US weight to be retrieved
+## 3. Ensure that a negative input weight is rejected
 
-A property can return a reference to an object.
+Add [validation][stackoverflow.com-validating-properties] to the `Weight`'s `set` accessor to throw an exception.
 
-## 4. Allow the machine's units to be set to pounds
+## 4. Allow a tare adjustment to be applied to the weighing machine
 
-`Units` is a good candidate for an [auto-implemented property][docs.microsoft.com-auto-implemented-properties].
+`TareAdjustment ` is a good candidate for an [auto-implemented property][docs.microsoft.com-auto-implemented-properties].
 
-## 5. Allow a tare adjustment to be applied to the weighing machine
+## 5. Ensure that the weighing machine has a default tare Adjustment.
 
-Accessors can have [different access levels][docs.microsoft.com-properties-and-restricted-access] to each other.
+You can initialize [auto-implemented property][docs.microsoft.com-auto-implemented-properties] similarly to fields.
+
+## 6. Allow the weight to be retrieved
+
+Perform the right calculation in the `DisplayWeight`'s `get` acessor.
 
 [docs.microsoft.com-properties]: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/properties
 [docs.microsoft.com-using-properties]: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/using-properties
--- a/exercises/concept/weighing-machine/.docs/instructions.md
+++ b/exercises/concept/weighing-machine/.docs/instructions.md
@@ -42,7 +42,18 @@ wm.InputWeight = 60m;
 //  => wm.InputWeight == 60m
 ```
 
-## 2. Ensure that a negative input weight is rejected
+## 2. Ensure that a Default unit is kilograms
+
+Implement the `WeigingMachine.Unit` property to allow the unit to be Kilograms by default:
+
+
+```csharp
+var wm = new WeighingMachine();
+
+//  => wm.Unit == Kilograms
+```
+
+## 3. Ensure that a negative input weight is rejected
 
 Add validation to the `WeighingMachine.InputWeight` property to throw an `ArgumentOutOfRangeException` when trying to set it to a negative weight:
 
@@ -51,7 +62,7 @@ var wm = new WeighingMachine();
 wm.InputWeight = -10m; // Throws an ArgumentOutOfRangeException
 ```
 
-## 3. Allow the US weight to be retrieved
+## 4. Allow the US weight to be retrieved
 
 Implement the `WeighingMachine.USDisplayWeight` property and the `USWeight` class:
 
@@ -63,7 +74,7 @@ var usw = wm.USDisplayWeight;
 // => usw.Pounds == 132 && usw.Ounces == 4
 ```
 
-## 4. Allow the machine's unit to be set to pounds
+## 5. Allow the machine's unit to be set to pounds
 
 Implement the `WeighingMachine.Unit` property:
 
@@ -76,7 +87,7 @@ var usw = wm.USDisplayWeight;
 // => usw.Pounds == 175 && usw.Ounces == 8
 ```
 
-## 5. Allow a tare adjustment to be applied to the weighing machine
+## 6. Allow a tare adjustment to be applied to the weighing machine
 
 Implement the `WeighingMachine.TareAdjustment` and `WeighingMachine.DisplayWeight` properties:
 
--- a/exercises/concept/weighing-machine/.docs/instructions.md
+++ b/exercises/concept/weighing-machine/.docs/instructions.md
@@ -63,14 +63,14 @@ var usw = wm.USDisplayWeight;
 // => usw.Pounds == 132 && usw.Ounces == 4
 ```
 
-## 4. Allow the machine's units to be set to pounds
+## 4. Allow the machine's unit to be set to pounds
 
-Implement the `WeighingMachine.Units` property:
+Implement the `WeighingMachine.Unit` property:
 
 ```csharp
 var wm = new WeighingMachine();
 wm.InputWeight = 175.5m;
-wm.Units = Units.Pounds;
+wm.Unit = Unit.Pounds;
 
 var usw = wm.USDisplayWeight;
 // => usw.Pounds == 175 && usw.Ounces == 8
--- a/exercises/concept/weighing-machine/.docs/instructions.md
+++ b/exercises/concept/weighing-machine/.docs/instructions.md
@@ -2,99 +2,72 @@
 
 In this exercise you'll be modelling a weighing machine.
 
-The weight can be set and retrieved in pounds or kilograms and cannot be negative.
+You have 6 tasks each of which requires you to implement one or more properties:
 
-The weight can be displayed in SI units or US units
-, pounds and ounces.
+## 1. Allow the weighing machine to have a precision.
 
-A tare adjustment can be applied to the weight (for instance to deduct the
-weight of a container). This can be any value (even negative or a value that makes the display weight negative)
-as there are doubts about the accuracy
-of the weighing machine. For security reasons this value cannot be retrieved.
-
-Note that:
-
-```
-display-weight = input-weight - tare-adjustment
-```
-
-Conversion ratios are as follows:
-
-- 16 ounces to a pound
-- 2.20462 kg to a pound
-
-For Example:
-
-- 60 kilograms == 132.2772 ponds
-- 132.2772 pounds == 132 pounds 4 ounces
-
-You have 5 tasks each of which requires you to implement one or
-more properties:
-
-## 1. Allow the weight to be set on the weighing machine
-
-Implement the `WeigingMachine.InputWeight` property to allow the weight to be get and set:
+It important to ensure that our weighing machine is precise.
+Implement the `WeigingMachine` class to have a precision property set only from the constructor:
 
 ```csharp
-var wm = new WeighingMachine();
-wm.InputWeight = 60m;
+var wm = new WeighingMachine(precision:3);
 
-//  => wm.InputWeight == 60m
+//  => wm.Precision == 3
 ```
 
-## 2. Ensure that a Default unit is kilograms
+## 2. Allow the weight to be set on the weighing machine
 
-Implement the `WeigingMachine.Unit` property to allow the unit to be Kilograms by default:
-
+Implement the `WeigingMachine.Weight` property to allow the weight to be get and set:
 
 ```csharp
-var wm = new WeighingMachine();
+var wm = new WeighingMachine(precision:3);
+wm.Weight = 60m;
 
-//  => wm.Unit == Kilograms
+//  => wm.Weight == 60m
 ```
 
 ## 3. Ensure that a negative input weight is rejected
 
-Add validation to the `WeighingMachine.InputWeight` property to throw an `ArgumentOutOfRangeException` when trying to set it to a negative weight:
+The weight can be set and retrieved, but it can't be negative.
+Add validation to the `WeighingMachine.Weight` property to throw an `ArgumentOutOfRangeException` when trying to set it to a negative weight:
 
 ```csharp
-var wm = new WeighingMachine();
-wm.InputWeight = -10m; // Throws an ArgumentOutOfRangeException
+var wm = new WeighingMachine(precision:3);
+wm.Weight = -10m; // Throws an ArgumentOutOfRangeException
 ```
 
-## 4. Allow the US weight to be retrieved
+## 4. Ensure that the weighing machine has a default tare Adjustment.
 
-Implement the `WeighingMachine.USDisplayWeight` property and the `USWeight` class:
+A tare adjustment can be applied to the weight they usually have a bias towards overestimating the weight.
+Implement the `WeighingMachine.TareAdjustment` property to have the default value as 5.
 
 ```csharp
-var wm = new WeighingMachine();
-wm.InputWeight = 60m;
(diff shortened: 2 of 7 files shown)


exercises/concept/weighing-machine/.docs/hints.md
```diff
@@ -5,25 +5,29 @@
 - [Properties][docs.microsoft.com-properties]
 - [Using Properties][docs.microsoft.com-using-properties]
 
-## 1. Allow the weight to be set on the weighing machine
+## 1. Allow the weighing machine to have a precision.
```
ErikSchierboom, line 8:
```suggestion
## 1. Allow the weighing machine to have a precision
```

exercises/concept/weighing-machine/.docs/hints.md
```diff
@@ -5,25 +5,29 @@
...
+Add [validation][stackoverflow.com-validating-properties] to the `Weight`'s `set` accessor to throw an exception.
 
-## 4. Allow the machine's units to be set to pounds
+## 4. Allow a tare adjustment to be applied to the weighing machine
 
-`Units` is a good candidate for an [auto-implemented property][docs.microsoft.com-auto-implemented-properties].
+`TareAdjustment ` is a good candidate for an [auto-implemented property][docs.microsoft.com-auto-implemented-properties].
 
-## 5. Allow a tare adjustment to be applied to the weighing machine
+## 5. Ensure that the weighing machine has a default tare Adjustment.
```
ErikSchierboom, line 24:
```suggestion
## 5. Ensure that the weighing machine has a default tare adjustment
```

exercises/concept/weighing-machine/.docs/hints.md
```diff
@@ -5,25 +5,29 @@
 - [Properties][docs.microsoft.com-properties]
 - [Using Properties][docs.microsoft.com-using-properties]
 
-## 1. Allow the weight to be set on the weighing machine
+## 1. Allow the weighing machine to have a precision.
 
-A property with a private [backing field][docs.microsoft.com-properties-with-backing-fields] is appropriate here.
+A get only property that can be intialised only from the constructor.
```
ErikSchierboom, line 10: Hints are usually listed as a list (this comment also applies to the other hints):

```suggestion
- A get only property that can be intialised only from the constructor.
```

While we're at it, maybe rephrase a little bit to be a bit more "directive":

```suggestion
- You need to add a `get`-only property to store the precision in.
- You can initialize the `get`-only property from the constructor (and only from the constructor).
```

ErikSchierboom: Is there perhaps a page to link to that describes a get-only property?

exercises/concept/weighing-machine/.docs/hints.md
```diff
@@ -5,25 +5,29 @@
...
 
-## 2. Ensure that a negative input weight is rejected.
+## 2. Allow the weight to be set on the weighing machine
 
-Add [validation][stackoverflow.com-validating-properties] to the `InputWeight`'s `set` accessor to throw an exception.
+A property with a private [backing field][docs.microsoft.com-properties-with-backing-fields] is appropriate here
```
ErikSchierboom, line 14:
```suggestion
- You should add a property with a private [backing field][docs.microsoft.com-properties-with-backing-fields].
```

exercises/concept/weighing-machine/.docs/hints.md
```diff
@@ -5,25 +5,29 @@
...
 
-## 3. Allow the US weight to be retrieved
+## 3. Ensure that a negative input weight is rejected
 
-A property can return a reference to an object.
+Add [validation][stackoverflow.com-validating-properties] to the `Weight`'s `set` accessor to throw an exception.
```
ErikSchierboom, line 18:
```suggestion
- The `value` keyword is used to refer to the value the caller wants the property to have.
- Add [validation][stackoverflow.com-validating-properties] to the `Weight`'s `set` accessor to throw an exception when the value is invalid.
```

exercises/concept/weighing-machine/.docs/hints.md
```diff
@@ -5,25 +5,29 @@
...
 
-## 4. Allow the machine's units to be set to pounds
+## 4. Allow a tare adjustment to be applied to the weighing machine
 
-`Units` is a good candidate for an [auto-implemented property][docs.microsoft.com-auto-implemented-properties].
+`TareAdjustment ` is a good candidate for an [auto-implemented property][docs.microsoft.com-auto-implemented-properties].
```
ErikSchierboom, line 22:
```suggestion
- You should add an [auto-implemented property][docs.microsoft.com-auto-implemented-properties] for the tare adjustment.
```

exercises/concept/weighing-machine/.docs/hints.md
```diff
@@ -5,25 +5,29 @@
...
 
-## 5. Allow a tare adjustment to be applied to the weighing machine
+## 5. Ensure that the weighing machine has a default tare Adjustment.
 
-Accessors can have [different access levels][docs.microsoft.com-properties-and-restricted-access] to each other.
+You can initialize [auto-implemented property][docs.microsoft.com-auto-implemented-properties] similarly to fields.
```
ErikSchierboom, line 26:
```suggestion
- You can initialize an [auto-implemented property][docs.microsoft.com-auto-implemented-properties] like you can initialize a field.
```

exercises/concept/weighing-machine/.docs/hints.md
```diff
@@ -5,25 +5,29 @@
...
+
+## 6. Allow the weight to be retrieved
+
+Perform the right calculation in the `DisplayWeight`'s `get` acessor.
```
ErikSchierboom, line 30:
```suggestion
- You should add a `get`-only property in which the correct display weight is calculated.
```

exercises/concept/weighing-machine/.docs/instructions.md
```diff
@@ -2,99 +2,72 @@
 
 In this exercise you'll be modelling a weighing machine.
 
-The weight can be set and retrieved in pounds or kilograms and cannot be negative.
+You have 6 tasks each of which requires you to implement one or more properties:
 
-The weight can be displayed in SI units or US units
-, pounds and ounces.
+## 1. Allow the weighing machine to have a precision.
```
ErikSchierboom, line 7:
```suggestion
## 1. Allow the weighing machine to have a precision
```

exercises/concept/weighing-machine/.docs/instructions.md
```diff
@@ -2,99 +2,72 @@
...
-wm.InputWeight = 60m;
+wm.TareAdjustment = -10;
 
-var usw = wm.USDisplayWeight;
-// => usw.Pounds == 132 && usw.Ounces == 4
+// => wm.TareAdjustment == -10
 ```
 
-## 5. Allow the machine's unit to be set to pounds
+## 5. Ensure that the weighing machine has a default tare Adjustment.
```
ErikSchierboom, line 51:
```suggestion
## 5. Ensure that the weighing machine has a default tare adjustment
```

exercises/concept/weighing-machine/.docs/instructions.md
```diff
@@ -2,99 +2,72 @@
...
 
-Implement the `WeighingMachine.Unit` property:
+A tare adjustment can be applied to the weight they usually have a bias towards overestimating the weight.
```
ErikSchierboom, line 53:
```suggestion
After some thorough testing, it appears that due to a manifacturing issue all weighing machines have a bias towards overestimating the weight by `5`.
```

exercises/concept/weighing-machine/.docs/instructions.md
```diff
@@ -2,99 +2,72 @@
...
+Implement the `WeighingMachine.TareAdjustment` property to have the default value as 5.
```
ErikSchierboom, line 54:
```suggestion
Change the `WeighingMachine.TareAdjustment` property to `5` as its default value.
```

exercises/concept/weighing-machine/.docs/instructions.md
```diff
@@ -2,99 +2,72 @@
...
-- 60 kilograms == 132.2772 ponds
-- 132.2772 pounds == 132 pounds 4 ounces
-
-You have 5 tasks each of which requires you to implement one or
-more properties:
-
-## 1. Allow the weight to be set on the weighing machine
-
-Implement the `WeigingMachine.InputWeight` property to allow the weight to be get and set:
+It important to ensure that our weighing machine is precise.
```
ErikSchierboom, line 9: To cater to different demands, we allow each weighing machine to be customized with a precision (the number of digits  after the decimal separator).

exercises/concept/weighing-machine/.meta/Exemplar.cs
```diff
@@ -1,80 +1,40 @@
...
+    private int _weight;
 
-    private decimal inputWeight;
+    public WeighingMachine(int precision)
+    {
+        Precision = precision;
+    }
 
-    public Unit Unit { get; set; } = Unit.Kilograms;
+    // Goal: getter-only property initialized in constructor
```
ErikSchierboom, line 12: All the `// Goal: ...` comments in this file can be removed. I just added them to show what they were used for.

```suggestion
```

exercises/concept/weighing-machine/WeighingMachine.cs
```diff
@@ -1,32 +1,13 @@
...
-enum Unit
-{
-    Pounds,
-    Kilograms
-}
-
 class WeighingMachine
 {
-    // TODO: define the 'InputWeight' property
+   // TODO: define the 'Precision' property
```
ErikSchierboom, line 5:
```suggestion
    // TODO: define the 'Precision' property
```

exercises/concept/weighing-machine/WeighingMachineTests.cs
```diff
@@ -5,76 +5,49 @@
...
-    public void Set_weight_and_get_weight()
-    {
-        var wm = new WeighingMachine();
-        wm.InputWeight = 60m;
-        Assert.Equal(60m, wm.InputWeight, precision: 3);
-    }
-
-    [Fact(Skip = "Remove this Skip property to run this test")]
-    public void Negative_weight_is_invalid()
+    public void Get_Precision()
```
ErikSchierboom, line 8: Maybe add a second test case to prevent hard-coding the result?

exercises/concept/weighing-machine/.meta/Exemplar.cs
```diff
@@ -1,80 +1,40 @@
...
-        {
-            return ApplyTareAdjustment(inputWeight);
+            if (value < 0) throw new ArgumentOutOfRangeException();
+            _weight = value;
         }
     }
 
-    public USWeight USDisplayWeight
+    // Goal: getter-only property with logic 
+    public int DisplayWeight
```
ErikSchierboom, line 33: I just noticed that we're not actually using the `Precision` property anywhere. Maybe we change this property to a `string` instead of an `int` and have it use the precision? So its output would look like: "60.5 kg" for precision `1`, and `60.58` for precision`2`, etc..

exercises/concept/weighing-machine/.docs/instructions.md
```diff
@@ -2,99 +2,72 @@
 
 In this exercise you'll be modelling a weighing machine.
```
ErikSchierboom, line 3: Maybe add something about all the weights being in kilograms?

exercises/concept/weighing-machine/.docs/instructions.md
```diff
@@ -2,99 +2,72 @@
...
-## 2. Allow specifying the unit used by the weighing machine
-
-Implement the `WeigingMachine.Unit` property to allow the unit to be set. `Unit.Kilograms` should be the default value:
+## 2. Allow the weight to be set on the weighing machine
 
+Implement the `WeigingMachine.Weight` property to allow the weight to be get and set:
 
 ```csharp
-var wm = new WeighingMachine();
+var wm = new WeighingMachine(precision:3);
```
ErikSchierboom, line 23:
```suggestion
var wm = new WeighingMachine(precision: 3);
```

exercises/concept/weighing-machine/.docs/instructions.md
```diff
@@ -2,99 +2,72 @@
...
+wm.Weight = 60m;
```
ErikSchierboom, line 24: The `m` suffix was needed for it to be a `decimal` type, but we're only using `int` values now. Could you remove the `m` suffix everywhere?

exercises/concept/weighing-machine/.docs/instructions.md
```diff
@@ -2,99 +2,72 @@
...
 
-//  => wm.Unit == Kilograms
+//  => wm.Weight == 60m
 ```
 
 ## 3. Ensure that a negative input weight is rejected
 
-Add validation to the `WeighingMachine.InputWeight` property to throw an `ArgumentOutOfRangeException` when trying to set it to a negative weight:
+The weight can be set and retrieved, but it can't be negative.
```
ErikSchierboom, line 31: Maybe do something like:

```suggestion
Clearly, someone cannot have a negative weight. 
```

exercises/concept/weighing-machine/.docs/instructions.md
```diff
@@ -2,99 +2,72 @@
...
+var wm = new WeighingMachine(precision:3);
+wm.Weight = -10m; // Throws an ArgumentOutOfRangeException
 ```
 
-## 4. Allow the US weight to be retrieved
+## 4. Allow a tare adjustment to be applied to the weighing machine
 
-Implement the `WeighingMachine.USDisplayWeight` property and the `USWeight` class:
+The tare adjustment can be any value (even negative or a value that makes the display weight negative)
+Implement the `WeighingMachine.TareAdjustment` property to allow the tare adjustment to also be set:
```
ErikSchierboom, line 42:
```suggestion
Implement the `WeighingMachine.TareAdjustment` property to allow the tare adjustment to be set:
```

exercises/concept/weighing-machine/WeighingMachineTests.cs
```diff
@@ -5,76 +5,49 @@
...
-    }
-
-    [Fact(Skip = "Remove this Skip property to run this test")]
-    public void Get_us_display_weight_ounces()
+    public void Set_weight_and_get_weight()
     {
         var wm = new WeighingMachine();
-        wm.InputWeight = 60m;
-        Assert.Equal(4, wm.USDisplayWeight.Ounces);
+        wm.Weight = 60m;
```
ErikSchierboom, line 18: As mentioned before, all the `m` suffixes should be removed as we're using integers now and not `decimals.`

exercises/concept/weighing-machine/WeighingMachineTests.cs
```diff
@@ -5,76 +5,49 @@
...
-        wm.InputWeight = 175.5m;
-        Assert.Equal(8, wm.USDisplayWeight.Ounces);
+        wm.Weight = 100;
+        wm.TareAdjustment = 10;
+        Assert.Equal(90, wm.DisplayWeight);
     }
 
-    [Fact(Skip = "Remove this Skip property to run this test")]
-    public void Apply_tare_adjustment_and_get_display_weight()
+     [Fact(Skip = "Remove this Skip property to run this test")]
```
ErikSchierboom, line 38:
```suggestion
    [Fact(Skip = "Remove this Skip property to run this test")]
```

ErikSchierboom (commented):
This is shaping up nicely! I've left some comments.

exercises/concept/weighing-machine/.docs/hints.md
18-F-cali: @ErikSchierboom 
What do you think about this: https://stackoverflow.com/questions/2719699/when-should-use-readonly-and-get-only-properties

exercises/concept/weighing-machine/WeighingMachineTests.cs
18-F-cali: The precision would not make sense with ints.

exercises/concept/weighing-machine/.docs/instructions.md
18-F-cali: If we use ints then we would also have to remove the precision property which had the goal of students learning about a getter-only property.

ErikSchierboom:
@18-F-cali Sorry about the int/decimal confusion!

18-F-cali:
@ErikSchierboom  I also updated the tareAdjustment property to double, because realistically tare is also floating point.

ErikSchierboom:
> due to the way xunit asserts work , I made the test cases precise.
> can u suggest some way to allow tolerance if this isn't sufficient?

There is a `precision: ` named argument that you can provide. Maybe set that to the precision that is used in that test? See https://github.com/exercism/csharp/blob/489b94ad5b251eec66b349dda7a2129c00a64f7d/exercises/concept/cars-assemble/CarsAssembleTests.cs#L9 for an example.

PR #1676, opened 2021-10-04
Change Database.Commit() to Database.EndTransaction()

### LINKED ISSUE #1675: Instructions for Object Relational Mapping refer to wrong instance method
Instructions refer to a Database.Commit method which doesn't exist. They should be updated to refer to Database.EndTransaction method.

`Database.Commit() commits the transaction to the database. It may throw an exception if it can't close the transaction of if Database.BeginTransaction() had not been called.`

jodihamann:
Update instructions.md to refer to Database.EndTransaction() instead of Database.Commit() as the latter method does not exist.

resolves #1675


ErikSchierboom (approved):
Thanks for finding this issue!

PR #1679, opened 2021-10-05
Updated HyperiaForex.cs to remove build errors when starting development

CodeVelo:
With the new unit tests, the project won't build because of missing methods/operators in HyperiaForex.cs. I suggest adding a stub for each operator used in the tests.

### DIFF
--- a/exercises/concept/hyperia-forex/HyperiaForex.cs
+++ b/exercises/concept/hyperia-forex/HyperiaForex.cs
@@ -1,3 +1,5 @@
+using System;
+
 public struct CurrencyAmount
 {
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
@CodeVelo I understand why one would want to do this, but this exercise introduces new syntax (the `operator` keyword). Whenever an exercise introduces new syntax, we want the students to actually _write_ that syntax themselves, as that is the main way they'll learn how to work with certain syntax. Providing a stub defeats that purpose. For this reason, the exercise that teaches classes also only has a TODO for the student to implement the class.

PR #1697, opened 2021-10-22
🤖 Change difficulty `0` to `1`

ErikSchierboom:
Set difficulty to `1` for exercises which difficulty is currently `0`.
See https://github.com/exercism/docs/pull/290 and https://github.com/exercism/configlet/pull/447.

Please feel free to update the difficulty to another value (the allowed range is 1-10) if you think that makes more sense.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1754,7 +1754,7 @@
         "uuid": "cef7deef-54ce-4201-b263-7cd2098533f8",
         "practices": [],
         "prerequisites": [],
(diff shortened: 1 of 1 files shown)


FaisalAfroz:
Need to merge this quickly. I think bcz of this my configlet lint is failing and showing error message "The value of `exercises.practice.difficulty` is 0, but it must be between 1 and 10 (inclusive):"

PR #1756, opened 2021-10-27
Fixed the number ranges for short and ushort

bencarlton:
Following the previous version of the chart will result in failed unit tests.


ErikSchierboom:
Superseded by #1796 

Thanks for the effort @bencarlton!

Issue #1795, opened 2021-10-29
Hyper-optimized Telemetry: Two unit tests are not following specifications in instructions for encoding


kaelon141:
On the Hyper-optimized Telemetry exercise, the instructions specify a table of ranges and target types for determining which type should be used when converting a long to the smallest integral type. Of particular interest in this issue is the range 0 - 65535 which should be represented with a `ushort`: 
 
![image](https://user-images.githubusercontent.com/28660369/139478375-967596dd-a479-4031-b0a0-9b90f0d06eac.png)

However,  two unit tests are not following this specification. First one is `ToBuffer_upper_short`:

![image](https://user-images.githubusercontent.com/28660369/139478743-db560855-f2bb-4d9b-b74d-15d41bcd9cb7.png)
Since the supplied value 32767 is between 0 and 65353, this should be an unsigned short. Specifically `0x2, 0xff, 0x7f`.

The second unit tests not following the instructions is `ToBuffer_Zero`. It has the exact same problem, expecting 0 to be a signed short, while the instructions specify that it should be unsigned.
![image](https://user-images.githubusercontent.com/28660369/139479316-b28b4e5a-f3f6-4e2f-8bc4-abd8072bbb4e.png)

Issue #1621, opened 2021-09-18
Wrong expected value in 2 test cases in Hyper-optimized Telemetry


18-F-cali:
![image](https://user-images.githubusercontent.com/63421973/133880917-c2055861-ea53-4e2a-b1d7-9a56c2ced4a5.png)
But according to the range in problem description
![image](https://user-images.githubusercontent.com/63421973/133880955-a00c6cd5-65c8-4bce-9f8a-8912fc576ea9.png)

 the expected values should be 

32767?							=> 0xfe, 0xff, 0x7f, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0
//unsigned						=> 0x2, 0xff, 0x7f, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0 

0?								=> 0xfe, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0 
//unsigned						=> 0x1, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0

ErikSchierboom:
@18-F-cali I'm updating the instructions in #1626. It does look like there might be another issue with the tests.

PR #1796, opened 2021-10-29
fix: correct unit tests in hyper-optimised-telemetry exercise

J141:
This PR fixes two unit tests on the hyper-optimised-telemetry exercise that wrongly assume `0` and `Int16.MaxValue` (32767) should be encoded into signed shorts, rather than unsigned shorts as specified in the exercise instructions.

image

 Directly targets issue #1795 and #1621 (The latter having been closed, though it reports the same problem)

This PR is incompatible with PR #1756, as the author of that PR aims to change the instructions to be in line with how the unit tests have been set up, while this PR aims to do the exact inverse by changing the unit tests to match the instructions instead.

### DIFF
--- a/exercises/concept/hyper-optimized-telemetry/.meta/Exemplar.cs
+++ b/exercises/concept/hyper-optimized-telemetry/.meta/Exemplar.cs
@@ -25,15 +25,9 @@ public static class TelemetryBuffer
             bytes = BitConverter.GetBytes((int)reading);
             bytes.CopyTo(allBytes, 1);
         }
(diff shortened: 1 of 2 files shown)


ErikSchierboom:
@J141 Could you also update the exemplar implementation (.meta/Exemplar.cs)? That should fix CI.

J141:
@ErikSchierboom Whoops, i somehow missed the existence of that Exemplar.cs. I've updated it to match the changes to the unit tests now.

ErikSchierboom (approved):
This makes much more sense now!

PR #1798, opened 2021-10-29
Add sets to the robot-name as prerequisites

Grenkin1988:
https://github.com/exercism/csharp/blob/main/exercises/practice/robot-name/.meta/Example.cs#L7
requires people to know at least some kind of collections.

### DIFF
--- a/config.json
+++ b/config.json
@@ -511,8 +511,8 @@
         "slug": "robot-name",
         "name": "Robot Name",
         "uuid": "46b8aea8-1528-43cb-a754-495ec2790ce2",
-        "practices": ["randomness", "classes"],
-        "prerequisites": ["randomness", "strings", "classes"],
+        "practices": ["randomness", "classes", "sets"],
+        "prerequisites": ["randomness", "strings", "classes", "sets"],
         "difficulty": 3
       },
       {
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -511,8 +511,8 @@
         "slug": "robot-name",
         "name": "Robot Name",
         "uuid": "46b8aea8-1528-43cb-a754-495ec2790ce2",
-        "practices": ["randomness", "classes"],
-        "prerequisites": ["randomness", "strings", "classes"],
+        "practices": ["randomness", "classes", "sets"],
+        "prerequisites": ["randomness", "strings", "classes", "sets"],
```
ErikSchierboom, line 515: You don't necessarily _need_ to use sets to be able to solve this exercise (it can definitely help though). Maybe we add a prerequisite for `generic-types` instead (we don't yet have a `collections` concept`)?

```suggestion
        "prerequisites": ["randomness", "strings", "classes", "generic-types"],
```

Grenkin1988: If this is how indicate Callections, then yes sure. As I wrote we need required collections here

ErikSchierboom: Yes, for now we only have `generic-types`. Having `collections` might be nice too later.

ErikSchierboom (approved):

PR #1799, opened 2021-10-30
need-for-speed: Add more tests in order to prevents negative battery drain

Grenkin1988:
Car can be implemented like this:
``` csharp
    public bool BatteryDrained()
    {
        return batteryCharge == 0;
    }
    public int DistanceDriven() => distanceDriven;
    public void Drive()
    {
        if (batteryCharge != 0)
        {
            distanceDriven += speed;
            batteryCharge -= batteryDrain;
        }
    }
```
This is not correct as it is easy to step over this `==0` check and get into negative `batteryCharge` and continue driving


ErikSchierboom (approved):
Good addition!

PR #1806, opened 2021-11-06
Fixed minor doc issue at concepts/optional-parameters

dalprahcd:
I believe there were minor typos when calling method at the documentation for concepts/optional-parameters


ErikSchierboom (approved):
Good catch!

PR #1811, opened 2021-11-10
Upgrade  projects to .NET 6

FaisalAfroz:
This PR resolves #1810 .

PR #1813, opened 2021-11-10
CI: Fix build errors

FaisalAfroz:
Update to .NEt 6 for fixing the build errors


ErikSchierboom:
@FaisalAfroz It looks like this change doesn't fix the build failure. These changes should actually be part of https://github.com/exercism/csharp/pull/1811, as we want them to be merged together in this case.

FaisalAfroz:
Yeah I think they should be committed together

ErikSchierboom:
@FaisalAfroz It looks like this change doesn't fix the build failure. These changes should actually be part of https://github.com/exercism/csharp/pull/1811, as we want them to be merged together in this case.

Issue #1805, opened 2021-11-06
rest-api exercise fails tests with Newtonsoft.Json or System.Text.Json


kchenery:
Hi,

I recently completed this exercise. I initially used `Newtonsoft.Json` and had all tests working locally.  Upon uploading the file the automated tests fail with the following error:

```
RestApi.cs(5,7): error CS0246: The type or namespace name 'Newtonsoft' could not be found (are you missing a using directive or an assembly reference?)
RestApi.cs(101,6): error CS0246: The type or namespace name 'JsonPropertyAttribute' could not be found (are you missing a using directive or an assembly reference?)
RestApi.cs(101,6): error CS0246: The type or namespace name 'JsonProperty' could not be found (are you missing a using directive or an assembly reference?)
RestApi.cs(95,6): error CS0246: The type or namespace name 'JsonPropertyAttribute' could not be found (are you missing a using directive or an assembly reference?)
RestApi.cs(95,6): error CS0246: The type or namespace name 'JsonProperty' could not be found (are you missing a using directive or an assembly reference?)
RestApi.cs(104,6): error CS0246: The type or namespace name 'JsonPropertyAttribute' could not be found (are you missing a using directive or an assembly reference?)
RestApi.cs(104,6): error CS0246: The type or namespace name 'JsonProperty' could not be found (are you missing a using directive or an assembly reference?)
RestApi.cs(107,6): error CS0246: The type or namespace name 'JsonPropertyAttribute' could not be found (are you missing a using directive or an assembly reference?)
RestApi.cs(107,6): error CS0246: The type or namespace name 'JsonProperty' could not be found (are you missing a using directive or an assembly reference?)
RestApi.cs(118,6): error CS0246: The type or namespace name 'JsonPropertyAttribute' could not be found (are you missing a using directive or an assembly reference?)
RestApi.cs(118,6): error CS0246: The type or namespace name 'JsonProperty' could not be found (are you missing a using directive or an assembly reference?)
RestApi.cs(121,6): error CS0246: The type or namespace name 'JsonPropertyAttribute' could not be found (are you missing a using directive or an assembly reference?)
RestApi.cs(121,6): error CS0246: The type or namespace name 'JsonProperty' could not be found (are you missing a using directive or an assembly reference?)
RestApi.cs(124,6): error CS0246: The type or namespace name 'JsonPropertyAttribute' could not be found (are you missing a using directive or an assembly reference?)
[...]

PR #1820, opened 2021-11-12
Allow rest-api to work with System.Text.Json

kchenery:
This relates to #1805 

@ErikSchierboom mentioned that it would be good if the `rest-api` exercise was modified to work with `System.Text.Json` rather than `Newtonsoft.Json`.

This PR migrates the tests and the `Example.cs` to use `System.Text.Json` instead.

Its worth noting that the main differences are how the two serializers will serialize a double/float value.  `System.Text.Json` will strip off the decimal places if they are 0 (e.g. 1.0 becomes 1) whereas `Newtonsoft.Json` will always include it.  As it happens all tests use double values that are actually integers.  It might be worth considering changing at least one of these tests to use a non integer double value.

### DIFF
--- a/exercises/practice/rest-api/.meta/Example.cs
+++ b/exercises/practice/rest-api/.meta/Example.cs
@@ -1,13 +1,13 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
-using Newtonsoft.Json;
+using System.Text.Json;
 
 public class User
 {
     public string name { get; }
-    public IDictionary<string, double> owes { get; private set; }
-    public IDictionary<string, double> owed_by { get; private set; }
+    public IDictionary<string, double> owes { get;  set; }
+    public IDictionary<string, double> owed_by { get;  set; }
     public double balance => owed_by.Sum(x => x.Value) - owes.Sum(x => x.Value);
 
     public User(string name)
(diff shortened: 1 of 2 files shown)


ErikSchierboom (approved):
Great work!

ErikSchierboom:
> Its worth noting that the main differences are how the two serializers will serialize a double/float value. System.Text.Json will strip off the decimal places if they are 0 (e.g. 1.0 becomes 1) whereas Newtonsoft.Json will always include it. As it happens all tests use double values that are actually integers. It might be worth considering changing at least one of these tests to use a non integer double value.

I think the implementation is actually wrong here, as the amount should all be integers. Would you be interested in submitting a follow-up PR to fix that?

kchenery:
Yep can do.

TBH: I'm of the opinion balance should be double and the balances should be amended to represent that. Then it might not matter if learner uses `System.Text.Json` or `Newtonsoft.Json`. However; fully cognizant that this might make this version of the exercise different to other language tracks.

Though making them `Int` will have the same result re serializers.

ErikSchierboom:
Yes I can understand why using doubles (`decimal` in C#) would be preferrable. The tests are generated from https://github.com/exercism/problem-specifications/blob/main/exercises/rest-api/canonical-data.json, so you might want to open an issue there. Regarding updating this exercise's tests, you'll have to update the test generator for this exercise: https://github.com/exercism/csharp/blob/main/generators/Exercises/Generators/RestApi.cs

kchenery:
@ErikSchierboom would you like all tests using `Newtonsoft.Json` migrated to `System.Text.Json`? I see there a number (20 I think) of them They look to be referencing `Newtonsoft.Json.Linq` and using `JArray`.

PR #1822, opened 2021-11-13
Clarify case in Isogram exercise

okaram:
It's not clear that upper and lower case letters should be counted as repetition. Added it in the description, and as an example.

### DIFF
--- a/exercises/practice/isogram/.docs/instructions.md
+++ b/exercises/practice/isogram/.docs/instructions.md
@@ -2,7 +2,7 @@
 
 Determine if a word or phrase is an isogram.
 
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
Thanks for wanting to improve this exercise! The instructions.md file is auto-generated from the description.md document of the exercise in the exercism/problem-specifications repo: https://github.com/exercism/problem-specifications/blob/main/exercises/isogram/description.md. We'll thus need to update the instructions there, after which we can regenerate these instructions. Could you open a PR there? Once an update is made there, we can create a new PR to update this repo.

PR #1823, opened 2021-11-13
Fixing interest_is_interesting example

okaram:
Example value is incorrect, right? Should be consistent with next example?


ErikSchierboom (approved):
Good catch!

Issue #1832, opened 2021-11-20
building-telemtry exercise instructions are vague


gmf002:
In step 2, the instructions say that the GetTelemetry method should return false the provided serial number is less than the previous serial number.

In step 3, the instructions say that if GetTelemetry returns false to return "no data".  However, the unit test expects that "no data" should be returned if the car has not been driven.  Based on the instruction in step 2, GetTelemetry returns true in this case.

Either step 2 should be updated to say reflect less than or equal to the previous serial number, or step 3 should be updated to reflect that "no data" should also be returned if car has not been driven.

Based on the units tests, it appears that step 3 should be where the clarification goes.

ErikSchierboom:
Looking at the exemplar solution, it has the following code:

```csharp
if (serialNum > latestSerialNum)
{
    latestSerialNum = serialNum;
    batteryPercentage = this.batteryPercentage;
    distanceDrivenInMeters = this.distanceDrivenInMeters;
    return true;
}
else
{
    serialNum = latestSerialNum;
    batteryPercentage = -1;
    distanceDrivenInMeters = -1;
    return false;
}
```

So I think the description of step 2 needs to be updated.

gmf002:
@ErikSchierboom  I hadn't uploaded my solution yet when i wrote the issue, so seeing that now makes more sense.   So, the instructions for step 2 should be updated, and a new unit test for that condition probably should be added.

gmf002:
Actually, looking at what I needed to do to get my code to pass the tests, it may still be step 3 (based purely on the unit tests).

My step two method has

```csharp
        if (serialNum < latestSerialNum)
        {
            serialNum = latestSerialNum;
            batteryPercentage = -1;
            distanceDrivenInMeters = -1;
            return false;
        }

        latestSerialNum = serialNum;
        batteryPercentage = this.batteryPercentage;
        distanceDrivenInMeters = this.distanceDrivenInMeters;
        return true;
```

which is equivalent to the exemplar solution.

What I needed to do to get the step 3 method to pass the tests was

```csharp
        var dataAvailable =
            car.GetTelemetryData(ref serialNum, out var batteryPercentage, out var distanceDrivenInMeters);

        return dataAvailable == false || batteryPercentage == 100
            ? "no data"
            : $"usage-per-meter={(100 - batteryPercentage) / distanceDrivenInMeters}";
```

So, either the instructions or test for step 3 are missing requirements as it makes no mention of the need for the car to be started, or the unit test is wrong.

[...]

PR #1834, opened 2021-11-20
Fixes exercism/csharp#1832 Update instructions.md

gmf002:
Clarified that "no data" should be returned if the car has not been started.


gmf002:
Did not proofread well enough.  Update is incorrect.

PR #1835, opened 2021-11-22
Closes #1832 Update instructions.md

### LINKED ISSUE #1832: building-telemtry exercise instructions are vague

gmf002:
Correct instructions for step three


ErikSchierboom (approved):

ErikSchierboom:
Thanks!

PR #1837, opened 2021-11-23
DZ use HashSet in GoCounting

dzegel:
Currently the tests are very brittle assuming elements are added to the lists in a particular order.
Using `HashSet` instead of `IEnumerable` and `Array` makes the tests more robust.

### DIFF
--- a/exercises/practice/go-counting/GoCounting.cs
+++ b/exercises/practice/go-counting/GoCounting.cs
@@ -15,12 +15,12 @@ public class GoCounting
         throw new NotImplementedException("You need to implement this function.");
     }
 
-    public Tuple<Owner, IEnumerable<(int, int)>> Territory((int, int) coord)
+    public Tuple<Owner, ISet<(int, int)>> Territory((int, int) coord)
     {
         throw new NotImplementedException("You need to implement this function.");
     }
 
-    public Dictionary<Owner, IEnumerable<(int, int)>> Territories()
+    public Dictionary<Owner, ISet<(int, int)>> Territories()
     {
         throw new NotImplementedException("You need to implement this function.");
     }
--- a/exercises/practice/go-counting/GoCounting.cs
+++ b/exercises/practice/go-counting/GoCounting.cs
@@ -15,12 +15,12 @@ public class GoCounting
         throw new NotImplementedException("You need to implement this function.");
     }
 
-    public Tuple<Owner, ISet<(int, int)>> Territory((int, int) coord)
+    public Tuple<Owner, HashSet<(int, int)>> Territory((int, int) coord)
     {
         throw new NotImplementedException("You need to implement this function.");
     }
 
-    public Dictionary<Owner, ISet<(int, int)>> Territories()
+    public Dictionary<Owner, HashSet<(int, int)>> Territories()
     {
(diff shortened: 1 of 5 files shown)


generators/Exercises/Generators/GoCounting.cs
```diff
@@ -72,7 +72,7 @@ private string RenderTerritoryAssert()
         private string RenderTerritoriesAssert()
         {
             var territoriesAssert = new StringBuilder();
-            territoriesAssert.AppendLine(Render.AssertEqual("expected.Keys", "actual.Keys"));
+            territoriesAssert.AppendLine(Render.AssertEqual("expected.Keys.Count", "actual.Keys.Count"));
```
dzegel, line 75: being that al the keys are used in the assertions below checking the count is sufficient.
alternative approaches would be to convert the key collections to sets or to just remove this assertion

ErikSchierboom: I think: just remove the assertion.

ErikSchierboom:
> still need to update the SUT's method signatures to return ISets instead of IEnumerables but I'm still trying to figure out how those files are generated

You mean the stub `GoCounting.cs` files? If so, they are not generated.

ErikSchierboom:
@dzegel https://github.com/exercism/csharp/blob/main/generators/Exercises/Generators/GoCounting.cs is the generator class. You should be able to run it using `cd generators && dotnet run -e go-counting`

dzegel:
> @dzegel https://github.com/exercism/csharp/blob/main/generators/Exercises/Generators/GoCounting.cs is the generator class. You should be able to run it using `cd generators && dotnet run -e go-counting`

I did run it, the test file changes in the diff are from running that

exercises/practice/go-counting/GoCounting.cs
```diff
@@ -15,12 +15,12 @@ public GoCounting(string input)
...
     }
 
-    public Tuple<Owner, IEnumerable<(int, int)>> Territory((int, int) coord)
+    public Tuple<Owner, HashSet<(int, int)>> Territory((int, int) coord)
     {
         throw new NotImplementedException("You need to implement this function.");
     }
 
-    public Dictionary<Owner, IEnumerable<(int, int)>> Territories()
+    public Dictionary<Owner, HashSet<(int, int)>> Territories()
```
dzegel, line 23: manually updated

exercises/practice/go-counting/GoCountingTests.cs
```diff
@@ -16,7 +16,7 @@ public void Black_corner_territory_on_5x5_board()
             "  W  ";
         var sut = new GoCounting(board);
         var actual = sut.Territory(coordinate);
-        var expected = (Owner.Black, new[] { (0, 0), (0, 1), (1, 0) });
+        var expected = (Owner.Black, new HashSet<(int, int)> { (0, 0), (0, 1), (1, 0) });
```
dzegel, line 19: all changes in this file were auto generated

generators/Exercises/Generators/GoCounting.cs
dzegel: done

ErikSchierboom (approved):
Perfect!

PR #1841, opened 2021-11-24
Matrix: remove unused properties

dzegel:
These properties are not referenced in the [canonical-data](https://github.com/exercism/problem-specifications/blob/main/exercises/matrix/canonical-data.json) or in the [test file](https://github.com/exercism/csharp/blob/main/exercises/practice/matrix/MatrixTests.cs).

Having these properties seems to simply be confusing.

### DIFF
--- a/exercises/practice/matrix/Matrix.cs
+++ b/exercises/practice/matrix/Matrix.cs
@@ -6,22 +6,6 @@ public class Matrix
     {
     }
 
-    public int Rows
-    {
-        get
-        {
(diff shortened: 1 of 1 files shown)


dzegel:
@ErikSchierboom any comments on this?

BTW for the future is there a standard team or channel I should tag on changes like this?

ErikSchierboom:
@dzegel You can tag @exercism/csharp, although we get notifications for new PRs automatically, so there is no need for tagging anybody really.

ErikSchierboom (approved):

Issue #1842, opened 2021-11-25
SgfParsing: tests broken


dzegel:
`Assert.Equal()` is causing a number of tests here to fail https://github.com/exercism/csharp/blob/main/exercises/practice/sgf-parsing/SgfParsingTests.cs#L33 because its doing a reference comparison and not a deep comparison.

You can see in this example both the `Data` and `Children` the properties in `SgfTree` are empty arrays in both the expected and actual objects and the test is failing.
![image](https://user-images.githubusercontent.com/2836278/143458386-42de42f7-28d7-42fe-8eb0-f21f652321f6.png)

I recently raised a [similar issue](https://github.com/exercism/csharp/issues/1828) in `GoCounting` which we resolved by using `HashSet` instead of `IEnumerable` but I don't see a comparable fix for this one.

For reference this is my solution (which I'm not publishing yet because the tests are broken).
_(Tests pass when I updated the tests to use FluentAssertions but in https://github.com/exercism/csharp/issues/1828 it was concluded that we don't want to introduce that dependency.)_
```cs
using System;
using System.Collections.Generic;
using System.Linq;
using System.Text.RegularExpressions;

public class SgfTree
{
    public SgfTree(IDictionary<string, string[]> data, params SgfTree[] children)
    {
        Data = data;
        Children = children;
    }

    public IDictionary<string, string[]> Data { get; }
    public SgfTree[] Children { get; }
}

public class SgfParser
{
    private const char unescapedOpenParen = 'ᑕ';
    private const char unescapedCloseParen = 'ᑐ';
    private const char unescapedOpenBrac = 'ᒥ';
    private const char unescapedCloseBrac = 'ᒣ';
    private const char unescapedSemiCol = 'ᒏ';

    private static readonly Regex treeRegex = new(@"^ᑕᒏ(?<properties>[^ᒏᑕᑐ]*)?(?<children>[ᑕᒏ].*)?ᑐ$", RegexOptions.Compiled);
    private static readonly Regex propertyRegex = new(@"^(?<property>[A-Z]+(?:ᒥ[^ᒥᒣ]+ᒣ)+)", RegexOptions.Compiled);

    public static SgfTree ParseTree(string treeString) => ParseEscapedTree(string.Concat(ToEscapedCharsArray(treeString)));

    private static SgfTree ParseEscapedTree(string treeString) => treeRegex.Match(treeString) switch
    {
        { Success: true, Groups: var groups } => new SgfTree(
[...]

PR #1844, opened 2021-11-27
SgfParsing: fix test assertions

dzegel:
Issue
==
https://github.com/exercism/csharp/issues/1842

Summary
==
Currently test won't pass with default equality unless students implement equality on `SgfTree` to do a deep comparison, which is a bit out of scope of the exercise and not indicated in the exercise anywhere.

This change implements an assertion check that will do the deep comparison necessary.


dzegel:
@ErikSchierboom any comments on this?

ErikSchierboom (approved):
LGTM! Thanks

PR #1858, opened 2022-01-11
make argument names consistent format in Dictionaries exercise

ericnerby:
In the International Calling Connoisseur C# exercise, an argument on one of the methods didn't follow the same naming convention.

Updated so that camel case naming is consistent throughout file.


ErikSchierboom (approved):
Good catch!

PR #1866, opened 2022-02-05
Remove nbsp from HelloWorld.cs

angelikatyborska:
I took a peak at HelloWorld to help debug an unrelated problem, and I noticed a red dot in the editor. Upon inspection it turns out to be a non-breaking space. I assume it's not needed?
Screenshot 2022-02-05 at 18 02 59

### DIFF
--- a/exercises/practice/hello-world/HelloWorld.cs
+++ b/exercises/practice/hello-world/HelloWorld.cs
@@ -1,4 +1,4 @@
-﻿public static class HelloWorld
+public static class HelloWorld
 {
     public static string Hello() => "Goodbye, Mars!";
 }


ErikSchierboom (approved):
I think that is an UTF-8 BOM marker: https://en.wikipedia.org/wiki/Byte_order_mark But it is totally fine to remove it. I _think_ all the other exercises have the same issue, so if you feel like it, you could also update those (feel free to not do that).

wolf99 (approved):

angelikatyborska:
@ErikSchierboom I would love to remove them all but I cannot figure out how to effectively do that. My editors do not show this character and don't allow grepping for it 😕 even GitHub doesn't show it unless I start editing the file via GitHub. So it only ever shows up as a red dot in the editor on exercism.org and in the GitHub editor.

ErikSchierboom:
@angelikatyborska No problem! I'll see if there's another way.

ErikSchierboom:
@angelikatyborska I've managed to create [a PR](#1869) that removes the BOM marker from all C# files.

The bash script to do so was: `find exercises/ -name *.cs -exec sed -i $'1s/^\uFEFF//' {} \;`

PR #1871, opened 2022-02-20
update for Task 4. test 17 and test 23.

R0han-Yadav:
according to the given logic the prisoner cannot be freed as he will be asleep in both cases.
changed expected output of both test17 and 23 to false

### DIFF
--- a/exercises/concept/annalyns-infiltration/AnnalynsInfiltrationTests.cs
+++ b/exercises/concept/annalyns-infiltration/AnnalynsInfiltrationTests.cs
@@ -167,7 +167,7 @@ public class AnnalynsInfiltrationTests
         var archerIsAwake = false;
         var prisonerIsAwake = false;
         var petDogIsPresent = true;
-        Assert.True(QuestLogic.CanFreePrisoner(knightIsAwake, archerIsAwake, prisonerIsAwake, petDogIsPresent));
+        Assert.False(QuestLogic.CanFreePrisoner(knightIsAwake, archerIsAwake, prisonerIsAwake, petDogIsPresent));
     }
 
     [Fact]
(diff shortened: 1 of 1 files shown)


ynfle:
Thanks for the PR.

As per [the instructions](https://github.com/exercism/csharp/blob/4a1423cfcd29521026a4989a2f250810d0c34259/exercises/concept/annalyns-infiltration/.docs/instructions.md?plain=1#L18) as long as the dog is present and the archer is awake, the prisoner can be freed.

Updates to the instructions may be needed though.

R0han-Yadav:
If Annalyn has her pet dog with her she can rescue the prisoner if the archer is asleep. The knight is scared of the dog and the archer will not have time to get ready before Annalyn and the prisoner can escape.
If Annalyn does not have her dog then she and the prisoner must be very sneaky! Annalyn can free the prisoner if the prisoner is awake and the knight and archer are both sleeping, but if the prisoner is sleeping they can't be rescued: the prisoner would be startled by Annalyn's sudden appearance and wake up the knight and archer.

R0han-Yadav:
it will not matter if the dog is present and the archer is awake it will result in a fail.
furthermore if the prisoner is asleep it will be an immediate fail.

R0han-Yadav:
the dog will only help if the knight is awake.
if the archer is awake the rescue will not succeed.

ynfle:
It will only matter if the prisoner is awake, in the 2nd condition where the dog isn't present.

In both cases, the dog is present and the archer is asleep

R0han-Yadav:
yes

R0han-Yadav:
it seems that the instructions written in the rep. are different that that of the one given on the website

ynfle:
The exercise was updated on October 5, 2021. If you haven't update your exercise after starting it before that, it would show the old instructions. If you go to the overview page of this exercise, do you see a banner at the top telling you that the exercise is outdated?

R0han-Yadav:
It shows the exercise was updated in Feb of 22

ErikSchierboom:
@ynfle Is correct: the instructions should be read as mentioned. That said, the instructions have always been slightly unwieldy and hard to parse. Instead of changing the expected values, maybe you could try clarifying the instructions instead?

R0han-Yadav:
@ErikSchierboom I must agree the instructions are very confusing.

PR #1872, opened 2022-02-20
`need-for-speed`: Update method name `CanFinishTrack` to `TryFinishTrack`

ynfle:
This name change clarifies that the method is supposed to try and drive the car around the track to see if it can finish the track before it runs out of battery as opposed to exposing the state of the car and directly calculating the remaining distance that that car can travel

@ErikSchierboom, this differs slightly from your suggestion of `TryToFinishTrack`

Co-authored-by: Erik Schierboom <erik_schierboom@hotmail.com>


ErikSchierboom (approved):

ErikSchierboom:
Thanks!

ynfle:
My pleasure!

PR #1873, opened 2022-02-20
`need-for-speed`(`Exemplar.cs`): add condition for early exit

ynfle:
This checks to see if the car has already completed the track before trying to drive the car around the track another time in the while loop.


ynfle:
I'm not sure why the CI failed

ynfle:
My mistake.
Forgot I wasn't in nim a forgot the `()` prefix for methods

ErikSchierboom:
This is definitely more optimal. I think it needs rebasing though as I've just merged the rename PR.

ErikSchierboom (approved):

PR #1881, opened 2022-02-28
CI: various improvements

ErikSchierboom:
- gha: remove version comments
- gha: remove unused environment variable
- gha: only run on main branches

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,5 +11,3 @@ jobs:
         with:
           dotnet-version: "6.0.100"
       - run: pwsh ./test.ps1
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,6 +1,13 @@
 name: Test
 
-on: [push, pull_request]
+on:
+  pull_request:
+    branches:
+      - main
+  push:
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -1,15 +1,20 @@
...
     steps:
-      - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # 2.3.5
-      - uses: actions/setup-dotnet@9211491ffb35dd6a6657ca4f45d43dfe6e97c829 # 1.7.2
+      - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
+      - uses: actions/setup-dotnet@9211491ffb35dd6a6657ca4f45d43dfe6e97c829
         with:
           dotnet-version: "6.0.100"
       - run: pwsh ./test.ps1
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
wolf99, line 15: For understanding, is this here because checkout used to need it and now no longer does. Or was it never needed for this task?

wolf99 (approved):

iHiD (approved):

.github/workflows/test.yml
ErikSchierboom: It was never need. I think it was just a copy/paste error.

PR #1882, opened 2022-02-28
Updating calculation to avoid confusion

mdrakiburrahman:
`CentralBank.DisplayGDP(555f, 10000f)` should be **5550000**, not **5550000_00_** (2 extra zeros in readme). This can cause confusions to someone getting their head around type overflow on why those 2 extra zeros are there 😊


ErikSchierboom:
Thanks!

PR #1891, opened 2022-03-05
Updated clarity in match criteria for ListLinesWithPasswords()

mdrakiburrahman:
It wasn't clear to me that in Task 5, the exact match `password` shouldn't be returned, but `password123` should. The test cases don't make this clear either, so added a bit of extra description to avoid others being confused.

### DIFF
--- a/exercises/concept/parsing-log-files/.docs/instructions.md
+++ b/exercises/concept/parsing-log-files/.docs/instructions.md
@@ -82,7 +82,7 @@ lp.RemoveEndOfLineText("[INF] end-of-line23033 Network Failure end-of-line27");
 
 ## 5. List lines with extremely weak passwords so the guilty can be punished
 
-Before expunging the passwords from the file we need to list any instances where passwords begin with the text "password".
+Before expunging the passwords from the file we need to list any instances where the contents begin with the text "password" followed by any alphanumerics in the same word; i.e. an exact match of "password" is not flagged, but "passwordsecret" is - see example.
 
 Implement the `LogParser.ListLinesWithPasswords()` method to print out the offending password followed by the complete line.
 
--- a/exercises/concept/parsing-log-files/.docs/instructions.md
+++ b/exercises/concept/parsing-log-files/.docs/instructions.md
@@ -80,23 +80,25 @@ lp.RemoveEndOfLineText("[INF] end-of-line23033 Network Failure end-of-line27");
 // => "[INF]  Network Failure "
 ```
 
-## 5. List lines with extremely weak passwords so the guilty can be punished
+## 5. Return formatted lines with secret-code
 
-Before expunging the passwords from the file we need to list any instances where the contents begin with the text "password" followed by any alphanumerics in the same word; i.e. an exact match of "password" is not flagged, but "passwordsecret" is - see example.
+We're trying to spot for a secret-code (passed as an alphanumeric `string` parameter). Specifically, we're looking for that secret-code to be present in the lines meeting the following rules:
+1. is preceded by a white-space character; e.g. ` 5up3rSecr3tC0de` - see [here](https://docs.microsoft.com/en-us/dotnet/standard/base-types/character-classes-in-regular-expressions#whitespace-character-s)
+2. is followed by one or more _non_white-space characters ; e.g. ` 5up3rSecr3tC0de123` - see [here](https://docs.microsoft.com/en-us/dotnet/standard/base-types/character-classes-in-regular-expressions#non-whitespace-character-s)
+3. is case-insensitive; e.g. ` 5Up3RSEcr3tC0dE`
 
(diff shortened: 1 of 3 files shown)


exercises/concept/parsing-log-files/.docs/instructions.md
```diff
@@ -82,7 +82,7 @@ lp.RemoveEndOfLineText("[INF] end-of-line23033 Network Failure end-of-line27");
 
 ## 5. List lines with extremely weak passwords so the guilty can be punished
 
-Before expunging the passwords from the file we need to list any instances where passwords begin with the text "password".
+Before expunging the passwords from the file we need to list any instances where the contents begin with the text "password" followed by any alphanumerics in the same word; i.e. an exact match of "password" is not flagged, but "passwordsecret" is - see example.
```
ErikSchierboom, line 85: I completely agree that this needs some improvement. I'm not sure the above change fully captures what we want to do with the method, namely to use the `\s` and `\S` character classes: https://docs.microsoft.com/en-us/dotnet/standard/base-types/character-classes-in-regular-expressions#whitespace-character-s

If you check the [exemplar implementation](https://github.com/exercism/csharp/blob/f89b5cc5fd09e24ca0d90f435b5f387c326484ba/exercises/concept/parsing-log-files/.meta/Exemplar.cs#L35), you'll see the code that we want a student to write (or something very similar).
[The tests](https://github.com/exercism/csharp/blob/f89b5cc5fd09e24ca0d90f435b5f387c326484ba/exercises/concept/parsing-log-files/ParsingLogFilesTests.cs#L67-L84) are fairly basic though, so I can see where the confusion is.
What we want is to check if a line contains a password that is weak, which means that the password:

- is preceded by a white-space character
- starts with the text "password"
- followed by one or more non-whitespace characters

Could you perhaps rephrase the entire section, as the following line is just wrong:

> Lines containing an offending password should be returned prefixed with ": ".

As can be seen in the example:

```csharp
var lp = new LogParser();
lp.ListLinesWithPasswords(new string[] {"my passwordsecret is great"});
// => "passwordsecret: my passwordsecret is great"
lp.ListLinesWithPasswords(new string[] {"my password secret"});
// => {"--------: my password secret"}
```

We want to return the actual password followed by a `:` if it is weak.

mdrakiburrahman: @ErikSchierboom I've updated the instructions, removed the references to "password" in Task 5 to avoid the confusion above^, and updated the Test Cases to align with the instructions a bit more. 

I think the "gotcha" someone would have going through this is figuring out the `\s` and `\S` which was missing from the previous setup.

E.g. In [my first pass](https://github.com/mdrakiburrahman/exercism_dotnet/blob/509c9e4a43843cead6c9f2fb02a1e9f4ddb52ca9/parsing-log-files/ParsingLogFiles.cs#L64) of the solution I had no idea I was supposed to use `\s` and `\S` from the Test Cases/instructions.

Lmk if you think of further changes.

exercises/concept/parsing-log-files/.docs/instructions.md
```diff
@@ -80,23 +80,25 @@ lp.RemoveEndOfLineText("[INF] end-of-line23033 Network Failure end-of-line27");
 // => "[INF]  Network Failure "
 ```
 
-## 5. List lines with extremely weak passwords so the guilty can be punished
+## 5. Return formatted lines with secret-code
```
ErikSchierboom, line 83: I think the new version is an improvement, but it does make things about secrets codes now instead of being about passwords.
Having looked into the exercise a bit more, I think we want users to use two things:

1. Character classes (`\S` and `\s`)
2. Use capture groups to extract text

The current implementation is just very unintuitive. I'm wondering if we should maybe completely rething the task. The current modification still looks quite a lot like the existing one, and I'm wondering whether that is more of a burden than a benefit.

Do you maybe have any other ideas? One idea I had was to tie this task into the first task, which is about identifying garbled log lines. What if the current task's goal was to un-garble a log line?

So `"Any old [ERR] text"` would become `"[ERR] Any old text"`. We could then allow for whitespace before or after the brackets to nudge students to the `\s` character class: `"Any old \t  [ERR] \t\t text"`  would become `"[ERR] Any old text"`

michalporeba:
@ErikSchierboom it has been over a year since the discussion on this PR stopped and it is on `instructions.md` file, which as I understand is getting an overhaul to standardise exercises. Would it make sense to close this PR?

ErikSchierboom:
> @ErikSchierboom it has been over a year since the discussion on this PR stopped and it is on instructions.md file, which as I understand is getting an overhaul to standardise exercises. Would it make sense to close this PR?

I think it makes sense, but the overhaul is "just" for practice exercises, and this is a concept exercise.

PR #1917, opened 2022-05-02
Remove C language switch statement link

pstarliu:


wolf99:
Seems like this link is actually referenced in the text of  [`about.md`](https://github.com/exercism/csharp/blob/main/concepts/switch-statements/about.md) where the text specifically references switch statements in C.

ErikSchierboom (approved):
> Seems like this link is actually referenced in the text of [about.md](https://github.com/exercism/csharp/blob/main/concepts/switch-statements/about.md) where the text specifically references switch statements in C.

That must be why it is listed here, as I initially just extracted all the external links from the `about.md` files to have some content in the `links.json` files.

PR #1914, opened 2022-04-27
Fix expected values for bert

vjba:
As per the docs for the exercise [here](https://exercism.org/tracks/csharp/exercises/developer-privileges), Bertrand's philtrum width is stated as 0.8, and Address 1 as Paris.

To conform to these constraints, I have updated the test's expected values to match the docs.

PR #1918, opened 2022-05-03
Update test cases

vjba:
Follow up PR for #1914 where @ErikSchierboom recommended some test case updates.


ErikSchierboom (approved):
Lovely!

PR #1937, opened 2022-05-13
modify expected exception in case of divide by zero attempt

Gautzilla:
The test case that tried to divide by zero and expected a `InvalidOperationException` should now expect a `DivideByZeroException`.


ErikSchierboom (approved):

ErikSchierboom:
Thanks!

PR #1948, opened 2022-05-21
Added extra task for Squeaky Clean exercise

trippshelnutt:
I was working on the Squeaky Clean exercise and it wasn't explicitly clear to me that letters should be removed. I added an extra task and instruction to clarify the requirements.

### DIFF
--- a/exercises/concept/squeaky-clean/.docs/instructions.md
+++ b/exercises/concept/squeaky-clean/.docs/instructions.md
@@ -3,7 +3,7 @@
 In this exercise you will implement a partial set of utility routines to help a developer
 clean up identifier names.
 
-In the 4 tasks you will gradually build up the routine `Clean` A valid identifier comprises
+In the 5 tasks you will gradually build up the routine `Clean` A valid identifier comprises
 zero or more letters and underscores.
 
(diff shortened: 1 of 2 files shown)


ErikSchierboom (commented):
Thanks for the PR! The instruction do state:

> A valid identifier comprises zero or more letters and underscores.

but I agree that this might be too subtle and a new test is a good idea. On small tweak: could you make this the fourth task and the existing fourth task the fifth? 

I just looked at the [test](https://github.com/exercism/csharp/blob/main/exercises/concept/squeaky-clean/SqueakyCleanTests.cs), and I feel like they could use a fair bit of attention too. For one, the order of the tests and the task to which they're linked doesn't fully match the instructions.
Would you be willing to also update the tests at the same time?

trippshelnutt:
No problem! I reordered the tasks in the instructions and moved some of the tests around to match the tasks. I also added a test explicitly to test for numbers.

The last test `Combine_conversions` seems to combine elements from all tasks. I left it associated with the last task. Do you have any feedback as to how the `Task` attribute should account for this?

ErikSchierboom (approved):
> The last test Combine_conversions seems to combine elements from all tasks. I left it associated with the last task. Do you have any feedback as to how the Task attribute should account for this?

That's perfectly fine!

PR #1950, opened 2022-06-01
Fix links in hints.md

BTowersCoding:
A couple of links to the strings tutorial were not being rendered as links because the names were different.


ErikSchierboom (commented):
Thanks!

ErikSchierboom (approved):

PR #1952, opened 2022-06-09
improved instruction clarity

davidsenk:
prior instruction was not clear what it actually meant at first glace, the "+ 16 = 25 = 5^2" is extraneous information. It makes it feel like there is a missing piece beyond the actual meaning of 3^2 + 4^2 = 5^2

### DIFF
--- a/exercises/practice/pythagorean-triplet/.docs/instructions.md
+++ b/exercises/practice/pythagorean-triplet/.docs/instructions.md
@@ -16,7 +16,7 @@ a < b < c
 For example,
 
 ```text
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
Thanks for the PR! The text for the exercise comes from a central repository: https://github.com/exercism/problem-specifications/blob/main/exercises/pythagorean-triplet/description.md Could you open an issue there? If the concensus is that this is a change we want to make, you can do a PR there and we can update things across all tracks.

davidsenk:
https://github.com/exercism/problem-specifications/pull/2055

PR #1956, opened 2022-06-18
Adds section on splitting strings

msdeibel:
In a recent mentoring discussion on the _Log Levels_ exercise the mentee complained that the splitting topic as a whole was missing on the concepts page.

It is kind of present with the link to the String methods, but also burried in there if you don't know what you're looking for.

The new part adds the classic  `Substring` and `IndexOf` way of doing the splitting, but withholding the possibility of a simple `string.Split` felt wrong for
me, since it makes things a lot easier from my point of view.

### DIFF
--- a/concepts/strings/about.md
+++ b/concepts/strings/about.md
@@ -17,10 +17,7 @@ escaped == verbatim;
 // => true
 ```
 
-At some point you might need to separate a string into multiple parts. This might be because you want to concatenate different parts from multiple strings or you 
-need to analyse a specific part more closely but don't need the rest. That happens mostly with application logs.
-
-The easiest way to do this is with the [Substring method][substring] 
+If you only need a part of a string, you can use the [`Substring()` method][substring] to extract just that part:
 
 ```csharp
 string sentence = "Frank chases the bus.";
--- a/concepts/strings/about.md
+++ b/concepts/strings/about.md
@@ -17,6 +17,19 @@ escaped == verbatim;
 // => true
 ```
 
+At some point you might need to separate a string into multiple parts. This might be because you want to concatenate different parts from multiple strings or you 
+need to analyse a specific part more closely but don't need the rest. That happens mostly with application logs.
+
+The easiest way to do this is with the [Substring method][substring] 
+
+```csharp
+string sentence = "Frank chases the bus.";
+string name = sentence.Substring(0,5)
+// => "Frank"
+```
+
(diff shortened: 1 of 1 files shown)


concepts/strings/about.md
```diff
@@ -17,6 +17,19 @@ escaped == verbatim;
 // => true
 ```
 
+At some point you might need to separate a string into multiple parts. This might be because you want to concatenate different parts from multiple strings or you 
+need to analyse a specific part more closely but don't need the rest. That happens mostly with application logs.
+
+The easiest way to do this is with the [Substring method][substring]
```
ErikSchierboom, line 23: This text is slightly verbose. Maybe we could shorten this a bit? I also don't know if we should be listing reasons for using `Substring`.

What about:

```suggestion
If you only need a part of a string, you can use the [`Substring()` method][substring] to extract just that part:
```

concepts/strings/about.md
```diff
@@ -17,6 +17,19 @@ escaped == verbatim;
...
 ```
 
+At some point you might need to separate a string into multiple parts. This might be because you want to concatenate different parts from multiple strings or you 
+need to analyse a specific part more closely but don't need the rest. That happens mostly with application logs.
+
+The easiest way to do this is with the [Substring method][substring] 
+
+```csharp
+string sentence = "Frank chases the bus.";
+string name = sentence.Substring(0,5)
```
ErikSchierboom, line 27:
```suggestion
string name = sentence.Substring(0, 5);
```

concepts/strings/about.md
```diff
@@ -17,6 +17,19 @@ escaped == verbatim;
...
+// => "Frank"
+```
+
+If you need to split the string at a certain character you can use [Substring][substring] in combination with [IndexOf][indexof]. In recent C# versions also the [Split method][splitting] has become very powerful. Check this out as well.
```
ErikSchierboom, line 31: Maybe we can expand on this (`IndexOf()` a bit more?

````suggestion
The [`IndexOf`() method][indexof] can be used to find the index of the first occurence of a `string` within a `string`, returning `-1` if the specified value could not be found:

```csharp
"continuous-integration".IndexOf("integration")
// => 11

"continuous-integration".IndexOf("deployment")
// => -1
```
````

I'd prefer not to mention the `Split()` method, as that returns an array which the student won't be familiar with a this point (for the same reason, I've used strings with `IndexOf`).

ErikSchierboom (commented):
Thanks for the PR! I've left some comments.

ErikSchierboom (approved):
Thanks!

ErikSchierboom:
Thanks! 🎉

Issue #1951, opened 2022-06-06
Suggestions for Remote Control Cleanup Exercise


haemmer-dave:
Hi,

a few suggestions for improvment. First, the theory part is very short and lacks details/examples about nested classes accessing containing classes and vice versa. It makes it look like the solution is to create a nested "Telemetry" class. However, looking at the tests, "Telemetry" is used as a property. Simply creating a nested Telemetry class will lead to compiler error CS0572. For C# beginners this is extremely confusing.

Second, the wording of task 2 and its hint are confusing as well. How an Interface, instead of a simple access modifier, is the solution is neither handled in the theory part nor understandable to me using C# docs.

ErikSchierboom:
Hi! Would you be willing to submit/work on a PR to improve this exercise?

haemmer-dave:
Yea, I can submit something to improve the theory and task 1.

However, I don't feel confident in improving task 2: I don't see the connection between the, imho, actual solution -- which is simply using an access modifier -- and the Interface hint. Maybe I'm missing something but I'm confused about task 2, hence do not really know what the original author intended and how exactly it can be improved.

ErikSchierboom:
I also feel that task 2 is quite weird. Not sure how we originally envisioned that working.

JeanAr7:
Hi it's 2024, the confusion still exists.

PR #1960, opened 2022-07-02
Instruction and task improvements

haemmer-dave:
This relates to https://github.com/exercism/csharp/issues/1951.
Clarified how to access inner type and role of "Telemetry" in instructions for task 1.

### DIFF
--- a/exercises/concept/remote-control-cleanup/.docs/instructions.md
+++ b/exercises/concept/remote-control-cleanup/.docs/instructions.md
@@ -4,7 +4,7 @@ Some "other" developers have been working on the remote control car project [exe
 
 ## 1. Separate concerns between the car itself and the telemetry system
 
-Refactor the `RemoteControlCar` class to move to separate out telemetry methods and properties.
+Refactor the `RemoteControlCar` class to move to separate out telemetry methods and properties accessed via a `Telemetry` property.
 
 ```csharp
 var car = new RemoteControlCar();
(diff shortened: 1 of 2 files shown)


exercises/concept/remote-control-cleanup/.docs/instructions.md
```diff
@@ -4,7 +4,7 @@ Some "other" developers have been working on the remote control car project [exe
 
 ## 1. Separate concerns between the car itself and the telemetry system
 
-Refactor the `RemoteControlCar` class to move to separate out telemetry methods and properties.
+Refactor the `RemoteControlCar` class to move to separate out telemetry methods and properties accessed via a `Telemetry` property.
```
ErikSchierboom, line 7: I like this!

exercises/concept/remote-control-cleanup/.docs/introduction.md
```diff
@@ -17,4 +17,18 @@ var inner = new Outer.CInner();
...
+    public class CInner
+    {
+        private Outer parent;
+
+        public CInner(Outer parent) {
+            this.parent = parent
+        }
+    }
+}
+```
```
ErikSchierboom, line 34: I'm not sure about this change. It feels a bit like overexplaining things to me, whereas we prefer to have students figure things out for themselves where possible.

haemmer-dave: Hi Erik. Thanks for your feedback. My understanding is that the Introduction section should provide some theory necessary for the exercise. Is that correct? I found the theory a bit lacking which is why I added the explanation. Would you be okay with adding a link to the C# documentation instead?

haemmer-dave: @ErikSchierboom I reverted this addition

ErikSchierboom (approved):

PR #1965, opened 2022-07-14
Fix typo in phone number analysis instructions

kytrinyx:
A coma is a state of unconsciousness where a person is unresponsive and cannot be woken.

Here we want commas, not comas.


LukeDias42 (approved):
Indeed, let's hope that expressions are not separated by comas o.o
That would make tuples very hard to do.

LukeDias42 (approved):

ErikSchierboom (approved):

PR #1979, opened 2022-09-30
Fix missing periods in building-telemetry headers

tomrijnbeek:
The second and third tasks in the building-telemetry exercise are missing periods, which means the numbers are not correctly stripped in the Exercism UI:

image


ErikSchierboom (approved):
Nice!

PR #1990, opened 2022-10-19
reverse-string: add approach introduction

ErikSchierboom:


exercises/practice/reverse-string/.approaches/performance/content.md
```diff
@@ -0,0 +1,109 @@
+# Performance
+
+On the [approaches page][approaches], we described several approaches:
+
+1. Using LINQ
+2. Using `Array.Reverse()`
+3. Using a `StringBuilder`
+
+In this document, we'll find out which of these approaches is the most performant one.
```
iHiD, line 9: I don't love "document" here but I don't have a better suggestion.

ErikSchierboom: I've restructured things a bit

ErikSchierboom: I think `Span<T>` is really its own approach, and the `performance` document should link to it.

exercises/practice/reverse-string/.approaches/array-reverse/content.md
```diff
@@ -0,0 +1,34 @@
...
+        return new string(chars);
+    }
+}
+```
+
+The `string` class' [`ToCharArray()`][to-char-array] method returns the string's characters as a `char[]`.
+
+```exercism/caution
+The `char[]` returned by `ToCharArray()` is a **copy** of the `string`'s characters.
+Modifying the values in the `char[]` does not update the `string` it was created from.
```
ErikSchierboom, line 21:
```suggestion
Modifying the values in the `char[]` does **not** update the `string` it was created from.
```

iHiD (approved):

PR #2000, opened 2022-10-26
reverse-string: add general guidance section

ErikSchierboom:


bobahop (approved):

bobahop:
Does it make sense for me to have more privileges for this repo? Or maybe e in exercism-reviewers? Whatever makes life easiest for others is okay with me.

kytrinyx:
@bobahop I've added you to reviewers. I think that makes more sense than just C# since you're going to be adding approaches to a lot of different repositories.

ErikSchierboom:
>  I've added you to reviewers

@kytrinyx Note that right now, that doesn't really do much I think as the reviewers team only has permissions on a limited number of repos (which doesn't include C#)

kytrinyx:
Ah, that's less than ideal. I'll ping you in DMs to see if we can figure out a better approach that will work across the board, not just for C#.

kytrinyx:
@bobahop I'm going to remove you from that team (as it's probably going to add more noise than it's worth).

PR #2004, opened 2022-10-27
bob answer array shortening fix

bobahop:
Fixes the `Any` Shortening section that was placed in the middle of the `single-line-if` Shortening section.


bobahop:
@ErikSchierboom Sorry, I pasted the `Any` Shortening section in the middle of the `single-line-if` Shortening section for the Answer Array approach

PR #2008, opened 2022-10-28
Leap fix repeated word in guidance

bobahop:
The The was a music group, but it doesn't belong here!


bobahop:
@ErikSchierboom or @kytrinyx : small fix for a repeated word in Leap guidance section

kytrinyx (approved):

kytrinyx:
Nice catch!

PR #2010, opened 2022-10-29
two-fer: add dig deeper content

ErikSchierboom:


iHiD (approved):
Approving blind but this looks great at a tiny glance! LGTM 🎉

exercises/practice/two-fer/.approaches/optional-parameter/content.md
```diff
@@ -0,0 +1,53 @@
...
+    public static string Speak(string name = "you")
+    {
+        return $"One for {name}, one for me.";
+    }
+}
+```
+
+We define a single `Speak()` method with one optional parameter: `name`.
+This optional parameter has its default value set to the string `"you"`.
+The caller of a method with an optional parameter can either pass in a value (which is then used) or not pass in a value (which caused the default value to be used).
```
kytrinyx, line 15: ... which then causes...

(Rather than "caused")

exercises/practice/two-fer/.approaches/method-overloading/content.md
```diff
@@ -0,0 +1,73 @@
...
+## Optional parameters vs. method overloads
+
+The main alternative to using method overloads is to use an [optional parameter approach][approach-optional-parameter]. If you're interested in how these two solutions compare to each other, go check out our [optional parameters vs method overloads article][article-optional-parameters-vs-method-overloading].
+
+[approach-optional-parameter]: https://exercism.org/tracks/csharp/exercises/two-fer/approaches/optional-parameter
+[article-optional-parameters-vs-method-overloading]: https://exercism.org/tracks/csharp/exercises/two-fer/articles/optional-parameters-vs-method-overloading
+[article-string-formatting]: https://exercism.org/tracks/csharp/exercises/two-fer/articles/string-formatting
+[optional-parameters-introduction]: https://learn.microsoft.com/en-us/archive/msdn-magazine/2010/july/csharp-4-0-new-csharp-features-in-the-net-framework-4#named-arguments-and-optional-parameters
+[method-overloading]: https://www.pluralsight.com/guides/overload-methods-invoking-overload-methods-csharp
+[string-interpolation]: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated
```
kytrinyx, line 73: Did we settle on needing to reference all the approaches from all the individual pages?

exercises/practice/two-fer/.articles/optional-parameters-vs-method-overloading/content.md
```diff
@@ -0,0 +1,181 @@
...
+
+```csharp
+public static string Speak(string name = "you")
+{
+    return $"One for {name}, one for me.";
+}
+```
+
+Using an optional parameter, we can assign a default value to that parameter which is then used if no value is passed for that parameter: `Speak()` and `Speak("you")` are equivalent.
+This new syntax allowed for a more succinct way of defining default values for parameters.
```
kytrinyx, line 34: Allows (rather than allowed)

exercises/practice/two-fer/.articles/optional-parameters-vs-method-overloading/content.md
```diff
@@ -0,0 +1,181 @@
...
+    IL_0012: br.s IL_0014
+
+    IL_0014: ldloc.0
+    IL_0015: ret
+} // end of method TwoFer::SpeakOptional
+```
+
+### Comparison
+
+The big differences are that the generated IL code:
```
kytrinyx, line 156: I would usually say big difference if there's only one, and main differences for more than one. No idea if this is just me, though.

exercises/practice/two-fer/.articles/string-formatting/content.md
```diff
@@ -0,0 +1,53 @@
+# String formatting
+
+There are various ways in which you can format the return string.
+
+## Option 1: string interpolation
+
+[String interpolation][string-interpolation] was introduced in C# 6.0 and is the most idiomatic way to build up a string with one more variable parts.
```
kytrinyx, line 7: One **or** more?

exercises/practice/two-fer/.articles/string-formatting/content.md
```diff
@@ -0,0 +1,53 @@
...
+
+```csharp
+$"One for {name}, one for me.";
+```
+
+## Option 2: string concatenation
+
+As there are few variable parts in the returned string (just one), regular [string concatentation][string-concatenation] works well too:
```
kytrinyx, line 15: I would probably rephrase: _As there is just one variable part_

kytrinyx (commented):
This is great. I have one question and a couple of minor nitpicks.

exercises/practice/two-fer/.approaches/method-overloading/content.md
ErikSchierboom: We didn't, and I don't do it for reverse-string, but felt it was more natural here due to me linking to an article that mentions the syntax the other approach uses.

PR #2031, opened 2022-11-11
reverse-string: approaches: remove `performance` directory

ee7:
~Commit 78fe449aec0c added approaches for `reverse-string`, [including a `performance` directory with the required files](https://github.com/exercism/csharp/tree/5b3e2fbf777219451ea5f76f1ebe1215bcd96771/exercises/practice/reverse-string/.approaches/performance), but it was not referenced in the `config.json` file.~

---

Commit 78fe449aec0c added approaches for `reverse-string`, including [one for `performance`](https://github.com/exercism/csharp/tree/5b3e2fbf777219451ea5f76f1ebe1215bcd96771/exercises/practice/reverse-string/.approaches/performance).

Commit e55d60b9af2f then added an article with the same content.

Commit bd162bb85542 removed the `approaches` object with the slug `performance` from the `.approaches/config.json`. However, it didn't remove the corresponding files. Do so.

In the future, `configlet lint` will produce an error when there is an approach or article directory with a name that does not correspond to a `slug` in the `config.json` file.


ErikSchierboom:
> So this PR should instead delete the exercises/practice/reverse-string/.approaches/performance directory?

Yep! The snippet difference is intentional

ErikSchierboom:
Thanks a lot!

ErikSchierboom (approved):

PR #2032, opened 2022-11-11
performace approach -> performance article

cmcaine:
The linked was changed from being an "approach" to an "article".


ErikSchierboom (approved):
Thanks!

PR #2041, opened 2022-11-14
difference-of-squares: add approaches

ErikSchierboom:
- difference-of-squares: add LINQ approach
- difference-of-squares: add for statement approach
- difference-of-squares: add math approach


kytrinyx (approved):
🎉

PR #2043, opened 2022-11-14
Update instructions.md

sergeifmf:
delete an extra colon and two zeroes in the end of the datetime format


ErikSchierboom (approved):
Thanks!

PR #2046, opened 2022-11-17
pig-latin: add approach

ErikSchierboom:
- pig-latin: add regular expression approach
- More work
- Restructure
- More work


kytrinyx (approved):
This is really, delightfully well-written!

PR #2079, opened 2023-02-10
darts: fix approach links

ErikSchierboom:


exercises/practice/darts/.approaches/introduction.md
```diff
@@ -22,8 +22,8 @@ public static int Score(double x, double y)
 }
 ```
 
-This approach uses an [`if`-statement][if-statement] for the "distance from center" checks.
-For more information, check the [`if`-statement approach][approach-if-statement].
+This approach uses an [`if` statements][if-statements] for the "distance from center" checks.
```
MatthijsBlom, line 25: @ErikSchierboom This link does not render.

ErikSchierboom: Thanks for the heads up! Fixed in https://github.com/exercism/csharp/pull/2092

PR #2086, opened 2023-03-04
Update introduction.md for tuples

wkuch:
This is a very early concept in the c# track and a conditional operator might be confusing for a beginner as it wasn't explained before.

### DIFF
--- a/concepts/tuples/introduction.md
+++ b/concepts/tuples/introduction.md
@@ -25,9 +25,13 @@ the second `Item2`, etc. Non-default names are discussed below.
 vertices = (60, 60, 60);
 
 //  return value
-(bool, int) GetSameOrBigger(int num1, int num2)
+(int, int) SumAndProduct(int x, int y)
 {
-    return (num1 == num2, num1 > num2 ? num1 : num2);
+    int sum = x + y;
+    int product = x * y;
+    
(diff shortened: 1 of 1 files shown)


concepts/tuples/introduction.md
```diff
@@ -25,9 +25,13 @@ the second `Item2`, etc. Non-default names are discussed below.
...
 //  return value
-(bool, int) GetSameOrBigger(int num1, int num2)
+(int, int) SumAndProduct(int x, int y)
 {
-    return (num1 == num2, num1 > num2 ? num1 : num2);
+    int sum = x + y;
+    int product = x * y;
+    
+    // return a tuple with both values
+    return (sum, product);
```
ErikSchierboom, line 34: I'd prefer the tuple to return two different types.

BTW This content will like also be in the concept's `about.md` file

michalporeba:
How about getting away from maths? Would something like this work? 

```csharp
(int, string) GetTodaysInfo()
{
    var today = DateTime.Now;
    return (today.DayOfYear, today.DayOfWeek.ToString());
}
```

It focuses on the return value with two different types, without obscuring the example by parameters.

ErikSchierboom:
Yep, that would be fine!

PR #2102, opened 2023-03-29
Simple list with LIFO behaviour

michalporeba:
**Discussion on the forum:** [(link)](http://forum.exercism.org/t/incorrect-solutions-passing-tests-in-c-track/4840/14).
**The problem:** The exercise [Simple Linked List (specification)](https://github.com/exercism/problem-specifications/blob/main/exercises/simple-linked-list/description.md) explains simple linked list as something with Last In First Out (LIFO) behaviour. It is at odds with what is currently [the reference implementation](https://github.com/exercism/csharp/blob/main/exercises/practice/simple-linked-list/.meta/Example.cs) in C# track, and what is required to pass [the tests](https://github.com/exercism/csharp/blob/main/exercises/practice/simple-linked-list/SimpleLinkedListTests.cs). 

The track's expected behaviour is First In First Out (FIFO). At the same time, the implementation, while technically correct, it is less classic, especially in the context of object oriented language like C#. There is no encapsulation of the list internals which can lead to problems with references should LIFO behaviour be implemented. Imagine the below situation:

```csharp
var a = new SimpleLinkedList<int>(1,2,3);
var b = a;
/* at this stage both a and b point to node with value 3 */
a = a.Remove();  /* LIFO pop operation */
/* now a points to 2 but b still points to 3 */
```

That is why the common implementation (examples from other tracks: [go](https://github.com/exercism/go/blob/main/exercises/practice/simple-linked-list/.meta/example.go), [c++](https://github.com/exercism/cpp/blob/main/exercises/practice/simple-linked-list/.meta/example.h), [java](https://github.com/exercism/java/blob/main/exercises/practice/simple-linked-list/.meta/src/reference/java/SimpleLinkedList.java)) uses a container external object encapsulating the linked list logic. As you can see in the provided ode the implementation is actually simpler than doing it without encapsulation.

**The solution:** What I'm proposing means big changes, but it bring the c# track in line with tracks from languages with similar philosophies while at the same time maintaining the .net / C# feel by maintaining `IEnumerable`, using `Add(T value)` and `Remove()` rather than `Push()` or `Pop()`. It introduces or exercises the concept without forcing solutions that have significant problems (hanging references). 

This is not finished product. I have done minimal work on the tests, only enough to make them work with the new structure again. If this is the direction we want to go in, I will work on more comprehensive test suite. 

**Open questions:**
* Is this a step we can take considering all the existing solutions would be invalidated (but future user will have better learning experience)
* Should we introduce `Count()` function in line with the problem spec?
* Is Remove the right method name? In C# typically it is used to remove a specific element from a collection. Perhaps RemoveFirst() would be an option, ~~or perhaps building on top of `IEnumerable` we could use Next()?~~ *(This would not work as iterating over the list would empty it!)*

### DIFF
--- a/exercises/practice/simple-linked-list/.docs/instructions.append.md
+++ b/exercises/practice/simple-linked-list/.docs/instructions.append.md
@@ -4,7 +4,7 @@ This exercise requires you to create a linked list data structure which can be i
 
 1. Implement the `Count` property - it should not be possible to change its value from the outside
 2. Implement the `Push(T value)` method that adds a value to the list at its head. 
-3. Implement `Pop()` method which removes and returns a value from the head. 
+3. Implement the `Pop()` method which removes and returns a value from the head. 
 4. Add a constructor to allow initialisation with a single value, or with an interable
 5. Implement the `IEnumerable<T>` interface. For more information, see [this page](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.ienumerable-1).
 6. Ensure `Reverse()` method is available. 
--- a/exercises/practice/simple-linked-list/.docs/instructions.append.md
+++ b/exercises/practice/simple-linked-list/.docs/instructions.append.md
@@ -1,4 +1,10 @@
 # Hints
 
-This exercise requires you to create a linked list data structure which can be iterated. This requires you to implement the IEnumerable\<T> interface.
-For more information, see [this page](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.ienumerable-1).
+This exercise requires you to create a linked list data structure which can be iterated. 
+
+1. Implement `Count` property - it should not be possible to change its value from the outside
+2. Implement `Push(T value)` method that adds a value to the list at its head. 
+3. Implement `Pop()` method which removes and returns a value from the head. 
+4. Add a constructor to allow initialisation with a single value, or with an interable
+5. Implement the `IEnumerable<T>` interface. For more information, see [this page](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.ienumerable-1).
+6. Ensure `Reverse()` method is available. 
--- a/exercises/practice/simple-linked-list/.docs/instructions.append.md
+++ b/exercises/practice/simple-linked-list/.docs/instructions.append.md
@@ -3,7 +3,7 @@
 This exercise requires you to create a linked list data structure which can be iterated. 
 
 1. Implement the `Count` property - it should not be possible to change its value from the outside
-2. Implement `Push(T value)` method that adds a value to the list at its head. 
+2. Implement the `Push(T value)` method that adds a value to the list at its head. 
 3. Implement `Pop()` method which removes and returns a value from the head. 
 4. Add a constructor to allow initialisation with a single value, or with an interable
 5. Implement the `IEnumerable<T>` interface. For more information, see [this page](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.ienumerable-1).
--- a/exercises/practice/simple-linked-list/.docs/instructions.append.md
+++ b/exercises/practice/simple-linked-list/.docs/instructions.append.md
@@ -1,4 +1,4 @@
-# Hints
+# Instruction append
 
 This exercise requires you to create a linked list data structure which can be iterated. 
 
--- a/exercises/practice/simple-linked-list/.docs/instructions.append.md
+++ b/exercises/practice/simple-linked-list/.docs/instructions.append.md
@@ -2,7 +2,7 @@
 
 This exercise requires you to create a linked list data structure which can be iterated. 
 
-1. Implement `Count` property - it should not be possible to change its value from the outside
+1. Implement the `Count` property - it should not be possible to change its value from the outside
 2. Implement `Push(T value)` method that adds a value to the list at its head. 
 3. Implement `Pop()` method which removes and returns a value from the head. 
 4. Add a constructor to allow initialisation with a single value, or with an interable
--- a/exercises/practice/simple-linked-list/.meta/Example.cs
+++ b/exercises/practice/simple-linked-list/.meta/Example.cs
@@ -1,63 +1,49 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
-using System.Linq;
 
 public class SimpleLinkedList<T> : IEnumerable<T>
 {
-    public SimpleLinkedList(T value) :
-        this(new[] { value })
-    {
+    private class Node { 
+        public T Value { get; set; }
+        public Node Next { get; set; }
     }
+    
+    private Node head;
 
-    public SimpleLinkedList(IEnumerable<T> values)
+    public SimpleLinkedList() { }
+
+    public SimpleLinkedList(params T[] values)
     {
-        var array = values.ToArray();
-
-        if (array.Length == 0)
-        {
-            throw new ArgumentException("Cannot create tree from empty list");
-        }
-
-        Value = array[0];
-        Next = null;
-
-        foreach (var value in array.Skip(1))
-        {
+        foreach(var value in values) { 
             Add(value);
         }
     }
 
-    public T Value { get; }
-
-    public SimpleLinkedList<T> Next { get; private set; }
-
-    public SimpleLinkedList<T> Add(T value)
+    public void Add(T value)
     {
-        var last = this;
+        var node = new Node { Value = value, Next = this.head };
+        this.head = node;
+    }
 
-        while (last.Next != null)
-        {
-            last = last.Next;
+    public T Remove()
+    {
+        if (this.head == null) { 
+            throw new InvalidOperationException("List is empty!");
         }
-
-        last.Next = new SimpleLinkedList<T>(value);
-
-        return this;
+        var value = head.Value;
+        head = head.Next;
+        return value;
     }
 
     public IEnumerator<T> GetEnumerator()
     {
-        yield return Value;
-
-        foreach (var next in Next?.AsEnumerable() ?? Enumerable.Empty<T>())
-        {
-            yield return next;
+        var current = this.head;
+        while(current != null) { 
+            yield return current.Value;
+            current = current.Next;
         }
     }
 
-    IEnumerator IEnumerable.GetEnumerator()
-    {
-        return GetEnumerator();
-    }
+    IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
 }
--- a/exercises/practice/simple-linked-list/.meta/Example.cs
+++ b/exercises/practice/simple-linked-list/.meta/Example.cs
@@ -21,7 +21,7 @@ public class SimpleLinkedList<T> : IEnumerable<T>
     }
 
     public int Count { get; private set; } = 0;
-
+    
     public void Push(T value)
     {
         var node = new Node { Value = value, Next = this.head };
--- a/exercises/practice/simple-linked-list/.meta/Example.cs
+++ b/exercises/practice/simple-linked-list/.meta/Example.cs
@@ -20,10 +20,13 @@ public class SimpleLinkedList<T> : IEnumerable<T>
         }
     }
 
+    public int Count { get; private set; } = 0;
+
     public void Push(T value)
     {
         var node = new Node { Value = value, Next = this.head };
         this.head = node;
+        this.Count++;
     }
 
     public T Pop()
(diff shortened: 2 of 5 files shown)


michalporeba:
When you look at the example, please keep in mind that we should discuss the approach rather than treat it as a code review at this stage.

ErikSchierboom:
@michalporeba I haven't yet found the time to look at this, sorry!

michalporeba:
No problem, @ErikSchierboom, take your time.

ErikSchierboom:
> Should we introduce Count() function in line with the problem spec?

I like that!

> Is Remove the right method name?

I think it is usually called `Unshift`, but that's not a great name either. `RemoveFirst` or `RemoveHead` could work.

exercises/practice/simple-linked-list/.meta/Example.cs
```diff
@@ -1,63 +1,49 @@
...
-using System.Linq;
 
 public class SimpleLinkedList<T> : IEnumerable<T>
 {
-    public SimpleLinkedList(T value) :
-        this(new[] { value })
-    {
+    private class Node { 
+        public T Value { get; set; }
+        public Node Next { get; set; }
```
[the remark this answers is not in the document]
michalporeba: Of course, but that is exactly why I started there to have a reference solution to test my tests with. It also servers as a conversation aid for what to do with this exercise.

exercises/practice/simple-linked-list/SimpleLinkedListTests.cs
```diff
@@ -1,83 +1,65 @@
...
-        Assert.Null(list.Next);
-    }
-
-    [Fact(Skip = "Remove this Skip property to run this test")]
+    [Fact]
     public void Two_item_list_first_value()
     {
-        var list = new SimpleLinkedList<int>(2).Add(1);
-        Assert.Equal(2, list.Value);
+        var list = new SimpleLinkedList<int>(1, 2);
```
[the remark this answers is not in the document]
michalporeba: I would like to build this exercise more with more granular steps, perhaps using `Add`/`Count` or `Add`/`Remove` first. Definitely Add makes sense to start with. 

The constructors are, in my opinion, very idiomatic for .net so it would be good to keep them, but they could be a separate step with some narrative to help understand why. 

I'll propose something soon (I hope, if not then next week as I will be mostly offline Thursday to Sunday).

michalporeba:
> I think it is usually called `Unshift`, but that's not a great name either. `RemoveFirst` or `RemoveHead` could work.
Well... it is in a doubly linked list. In a simple list like this I believe it is `push` and `pop`. But that is not so relevant in C# I think. Microsoft was rather consistently ignoring this standards favouring `Add` and `Remove` or `Enqueue` and `Dequeue` or similar words for similar operations throughout their libraries and tools. The `Stack<T>` implementation with `Push`, `Peek` and `Pop` is the exception.

The added challenge is that `Remove...` normally do not return. In that regard dotnet follows command query separation. Both `Add` and `Remove` cannot be used to get a value.

I see we have three options to implement the spirit of the exercise and keep dotnet ways of doing things

### Option 1 - be stack-like
The list behaves like a stack (LIFO) and so we could adopt the stack methods `Push` and `Pop`

### Option 2 - follow list naming and add `GetFirst`
Have three methods - `Add`, `RemoveFirst` and `GetFirst`. 

### Option 3 - GetAndRemoveFirst
Some .net libraries opt for methods like `GetAndRemove` breaking the command query separation in a very explicit name.  We could have `GetAndRemoveFirst` method

To be clear, for me option 3 is the least favourite by some margin. But I cannot make up my mind between options 1 and 2.

exercises/practice/simple-linked-list/SimpleLinkedListTests.cs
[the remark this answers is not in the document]
michalporeba: I think the tests should follow the TDD style, and with steps they can. The C# track already does it so much better than many other tracks I have tried, and with reflection at our disposal we can easily make it work great leading the student through the exercise step by step. 

If you don't mind @ErikSchierboom, I'd like to give it a go, implement something like this so we can talk specific pros and cons on an example.

ErikSchierboom:
Personally, my preference would be to go with option 1. I think we'll get the cleanest interface that way.

michalporeba:
Biggest change, but the cleanest end result. I'll got with that.

ErikSchierboom:
> The biggest change compared to what is there now, but the cleanest end result. I'll got with that.

Great! Let me know when it is ready to be reviewed.

michalporeba:
@ErikSchierboom I got to the point where I need some help. I think the current code could be a good starting point. Constructors are not needed, the `IEnumerable<int>` implementation is not yet needed, and all that is needed for the first task to pass is to add
```
public int Count { get; private set; }
```

But because there is no stub for the property `Count` the automated tests on the track will fail. What should I do here? 

Also, while I have added `[Task(x)]` attributes I haven't added instructions. I looked at a few other exercises with steps, and looked for documentation but I don't know what is the best / current / recommended way to do it, especially considering the efforts to standardise descriptions. Can you help?

ErikSchierboom:
> But because there is no stub for the property Count the automated tests on the track will fail. What should I do here?

Add a stub :) I think the default stub could be `public int Count => throw new NotImplementedException("Please implement this");`

> Also, while I have added [Task(x)] attributes I haven't added instructions. I looked at a few other exercises with steps, and looked for documentation but I don't know what is the best / current / recommended way to do it, especially considering the efforts to standardise descriptions. Can you help?

You can remove those. Practice exercises don't yet have tasks associated with them, although we are indeed try to start with that. For now, I'd not have them.

michalporeba:
The stub is in place, tasks removed and some information added to help guide the implementation. 

The previous version of the instructions stipulated that input will be correct, now this has been removed. As such, and for completness of the solution should we expect that `Pop()` throws `InvalidOperation` when there is nothing in the list?

ErikSchierboom:
> As such, and for completness of the solution should we expect that Pop() throws InvalidOperation when there is nothing in the list?

Yes please!

exercises/practice/simple-linked-list/.docs/instructions.append.md
```diff
@@ -1,4 +1,10 @@
 # Hints
 
-This exercise requires you to create a linked list data structure which can be iterated. This requires you to implement the IEnumerable\<T> interface.
-For more information, see [this page](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.ienumerable-1).
+This exercise requires you to create a linked list data structure which can be iterated. 
+
+1. Implement `Count` property - it should not be possible to change its value from the outside
```
ErikSchierboom, line 5:
```suggestion
1. Implement the `Count` property - it should not be possible to change its value from the outside
```

exercises/practice/simple-linked-list/.docs/instructions.append.md
```diff
@@ -1,4 +1,10 @@
...
+2. Implement `Push(T value)` method that adds a value to the list at its head. 
+3. Implement `Pop()` method which removes and returns a value from the head. 
+4. Add a constructor to allow initialisation with a single value, or with an interable
+5. Implement the `IEnumerable<T>` interface. For more information, see [this page](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.ienumerable-1).
+6. Ensure `Reverse()` method is available.
```
ErikSchierboom, line 10: This is not something that students need to implement as they get it for free by implement IEnumerable, right?

exercises/practice/simple-linked-list/.docs/instructions.append.md
```diff
@@ -1,4 +1,10 @@
 # Hints
 
-This exercise requires you to create a linked list data structure which can be iterated. This requires you to implement the IEnumerable\<T> interface.
-For more information, see [this page](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.ienumerable-1).
+This exercise requires you to create a linked list data structure which can be iterated. 
+
+1. Implement `Count` property - it should not be possible to change its value from the outside
+2. Implement `Push(T value)` method that adds a value to the list at its head. 
+3. Implement `Pop()` method which removes and returns a value from the head.
```
ErikSchierboom, line 7:
```suggestion
3. Implement the `Pop()` method which removes and returns a value from the head. 
```

exercises/practice/simple-linked-list/.docs/instructions.append.md
```diff
@@ -1,4 +1,10 @@
 # Hints
 
-This exercise requires you to create a linked list data structure which can be iterated. This requires you to implement the IEnumerable\<T> interface.
-For more information, see [this page](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.ienumerable-1).
+This exercise requires you to create a linked list data structure which can be iterated. 
+
+1. Implement `Count` property - it should not be possible to change its value from the outside
+2. Implement `Push(T value)` method that adds a value to the list at its head.
```
ErikSchierboom, line 6:
```suggestion
2. Implement the `Push(T value)` method that adds a value to the list at its head. 
```

exercises/practice/simple-linked-list/SimpleLinkedListTests.cs
```diff
@@ -1,83 +1,117 @@
+using System;
 using System.Linq;
+using System.Collections.Generic;
 using Xunit;
 
 public class SimpleLinkedListTests
 {
-    [Fact]
-    public void Single_item_list_value()
+    [Fact(Skip = "Remove this Skip property to run this test")]
```
ErikSchierboom, line 8: The first test is not skipped

```suggestion
    [Fact] 
```

exercises/practice/simple-linked-list/SimpleLinkedListTests.cs
```diff
@@ -1,83 +1,117 @@
...
+
+    private static SimpleLinkedList<int> CreateSimpleLinkedList(int value)
+    {
+        var type = typeof(SimpleLinkedList<>).MakeGenericType(typeof(int));
+		var constructor = type.GetConstructor(new Type[] { typeof(int) });
+		return (SimpleLinkedList<int>)constructor?.Invoke(new object[]{ value })
+			?? CreateSimpleLinkedList(new int[] { value });
+    }
+	
+	private static SimpleLinkedList<int> CreateSimpleLinkedList(params int[] values)
```
ErikSchierboom, line 64:
```suggestion
    private static SimpleLinkedList<int> CreateSimpleLinkedList(params int[] values)
```

exercises/practice/simple-linked-list/SimpleLinkedListTests.cs
```diff
@@ -1,83 +1,117 @@
...
+    {
+        var type = typeof(SimpleLinkedList<>).MakeGenericType(typeof(int));
+		var constructor = type.GetConstructor(new Type[]{typeof(int[])});	
+		return (SimpleLinkedList<int>)constructor.Invoke(new object[]{ values });
     }
 
     [Fact(Skip = "Remove this Skip property to run this test")]
-    public void Implements_enumerable()
+    public void Singlevalue_initialisation()
```
ErikSchierboom, line 72:
```suggestion
    public void Single_value_initialisation()
```

exercises/practice/simple-linked-list/SimpleLinkedListTests.cs
```diff
@@ -1,83 +1,117 @@
...
     {
-        var values = new SimpleLinkedList<int>(2).Add(1);
-        Assert.Equal(new[] { 2, 1 }, values);
+        var list = CreateSimpleLinkedList(7);
+        Assert.Equal(1, list.Count);
+        Assert.Equal(7, list.Pop());
+    }
+
+    [Fact(Skip = "Remove this Skip property to run this test")]
+    public void Multivalue_initialisation()
```
ErikSchierboom, line 80:
```suggestion
    public void Multi_value_initialisation()
```

exercises/practice/simple-linked-list/.docs/instructions.append.md
michalporeba: If they do it right. If they don't then the test will fail.

ErikSchierboom (approved):
I think this is great!

michalporeba:
Yes, please!

I'd like to submit more PRs for other things, but it looks like there are many open and it would be good to close them and have perhaps smaller changes to avoid work with rebasing and so on. Sort of limiting the work in progress (WIP).

ErikSchierboom:
That sounds great.

PR #2104, opened 2023-04-02
Palindrome product article

michalporeba:
**Forum discussion:** [[Post your C# Approaches](https://forum.exercism.org/t/post-your-c-approaches/4758)](http://forum.exercism.org/t/post-your-c-approaches/4758/25)

This is first draft of an article for the [Palindrome Products](https://github.com/exercism/csharp/tree/main/exercises/practice/palindrome-products) exercise taking a look at the side challenge of deciding if a number is a palindrome number or not.

### DIFF
--- a/.github/workflows/pssanalyzer.yml
+++ b/.github/workflows/pssanalyzer.yml
@@ -13,5 +13,5 @@ jobs:
   analyze:
     runs-on: ubuntu-20.04
     steps:
-      - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c
+      - uses: actions/checkout@8f4b7f84864484a7bf31766abe9204da3cbe65b3
       - run: pwsh ./powershell-script-analyzer.ps1
--- a/.github/workflows/pssanalyzer.yml
+++ b/.github/workflows/pssanalyzer.yml
@@ -13,5 +13,5 @@ jobs:
   analyze:
     runs-on: ubuntu-20.04
     steps:
-      - uses: actions/checkout@8f4b7f84864484a7bf31766abe9204da3cbe65b3
+      - uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab
       - run: pwsh ./powershell-script-analyzer.ps1
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
   test:
     runs-on: ubuntu-20.04
     steps:
-      - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c
+      - uses: actions/checkout@8f4b7f84864484a7bf31766abe9204da3cbe65b3
       - uses: actions/setup-dotnet@607fce577a46308457984d59e4954e075820f10a
         with:
           dotnet-version: "7.0.100"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
   test:
     runs-on: ubuntu-20.04
     steps:
-      - uses: actions/checkout@8f4b7f84864484a7bf31766abe9204da3cbe65b3
+      - uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab
       - uses: actions/setup-dotnet@607fce577a46308457984d59e4954e075820f10a
         with:
           dotnet-version: "7.0.100"
--- a/exercises/practice/palindrome-products/.articles/config.json
+++ b/exercises/practice/palindrome-products/.articles/config.json
@@ -4,7 +4,7 @@
             "uuid": "987cf2d1-7225-436f-9228-b88120ae2e70",
             "slug": "is-palindrome-check",
             "title": "Is a palindrome number check",
-            "blurb": "A few ways to check if a number is a palindrome.",
+            "blurb": "Different ways to check if a number is a palindrome.",
             "authors": ["michalporeba"]
         }
     ]
--- a/exercises/practice/palindrome-products/.articles/config.json
+++ b/exercises/practice/palindrome-products/.articles/config.json
@@ -1,11 +1,13 @@
 {
-    "articles": [
-        {
-            "uuid": "987cf2d1-7225-436f-9228-b88120ae2e70",
-            "slug": "is-palindrome-check",
-            "title": "Is a palindrome number check",
-            "blurb": "Different ways to check if a number is a palindrome.",
-            "authors": ["michalporeba"]
-        }
-    ]
+  "articles": [
+    {
+      "uuid": "987cf2d1-7225-436f-9228-b88120ae2e70",
+      "slug": "is-palindrome-check",
+      "title": "Is a palindrome number check",
+      "blurb": "Different ways to check if a number is a palindrome.",
+      "authors": [
+        "michalporeba"
+      ]
+    }
+  ]
 }
--- a/exercises/practice/palindrome-products/.articles/config.json
+++ b/exercises/practice/palindrome-products/.articles/config.json
@@ -1,0 +1,11 @@
+{
+    "articles": [
+        {
+            "uuid": "987cf2d1-7225-436f-9228-b88120ae2e70",
+            "slug": "is-palindrome-check",
+            "title": "Is a palindrome number check",
+            "blurb": "A few ways to check if a number is a palindrome.",
+            "authors": ["michalporeba"]
+        }
+    ]
+}
--- a/exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
+++ b/exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
@@ -46,8 +46,8 @@ Before we look at the code, let's look at an example. The number is 754.
 | Step | number | n % 10 | n / 10 |    list |
 |-----:|-------:|-------:|-------:|:--------|
 |    1 |    754 |      4 |     75 | [4]     |
-|    2 |     75 |      5 |      7 | [4,5]   |
-|    3 |      7 |      7 |      0 | [4,5,7] |
+|    2 |     75 |      5 |      7 | [5,4]   |
+|    3 |      7 |      7 |      0 | [7,5,4] |
 
 So now we have digits, already in reversed order `[7,5,4]`. We have to combine them into a number.
 
--- a/exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
+++ b/exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
@@ -8,10 +8,79 @@ There are multiple ways to do it in C# and we will look at them in this article.
 ## Approaches
 ### String reversal
 
-### Math and stack
+```csharp
+private static bool IsPalindrome(int number)
+{
+    var original = number.ToString();
+    var reversed = new String(original.Reverse().ToArray());
+    return original.Equals(reversed);
+}
+```
 
-### Just math
+```csharp
+private static bool IsPalindromeString(int number)
+=>  $"{number}".Equals(new String($"{number}".Reverse().ToArray()));
+```
+
+### List of digits
+
+```csharp
+private static bool IsPalindrome(int number) { 
+    var original = number;
+    var digits = new List<int>();
+    
+    while (number > 0) { 
+        digits.Add(number % 10);
+        number /= 10;
+    }
+
+    var reversed = 0;
+    foreach(var digit in digits) {
+        reversed = reversed * 10 + digit;
+    }
+    
+    return original == reversed;
+}
+```
+
+### Math 
+
+```csharp
+private static bool IsPalindrome(int number) { 
+    var original = number;
+    var reversed = 0;
+    
+    while (number > 0) { 
+        reversed = reversed * 10 + number % 10;
+        number /= 10;
+    }
+    
+    return original == reversed;
+}
+```
 
 ## Performance considerations
 
 ## Which one to choose?
+
+
+
+## The extension method options
+
+```csharp
+private static bool IsPalindrome(int number) { /*...*/ }
+
+public static void Main() 
+{
+    if (IsPalindrome(123)) { /* do something */ };
+}
+```
+
+```csharp
+private static bool IsPalindrome(this int number) { /*...*/ }
+
+public static void Main() 
+{
+    if (123.IsPalindrome()) { /* do something */ };
+}
+```
--- a/exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
+++ b/exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
@@ -41,21 +41,21 @@ private static bool IsPalindrome(int number)
 
 Another approach is to use [modulo division][modulo-division-link] to extract individual digits from the least significant to the most significant. 
 We can capture them in a list, and then read them again using maths to reconstruct the number in the reverse order. 
-Before we look at the code, let's look at an example. The number is 123. 
+Before we look at the code, let's look at an example. The number is 754. 
 
 | Step | number | n % 10 | n / 10 |    list |
 |-----:|-------:|-------:|-------:|:--------|
-|    1 |    123 |      3 |     12 | [3]     |
-|    2 |     12 |      2 |      1 | [3,2]   |
-|    3 |      1 |      1 |      0 | [3,2,1] |
+|    1 |    754 |      4 |     75 | [4]     |
+|    2 |     75 |      5 |      7 | [4,5]   |
+|    3 |      7 |      7 |      0 | [4,5,7] |
 
-So now we have digits, already in reversed order `[3,2,1]`. We have to combine them into a number.
+So now we have digits, already in reversed order `[7,5,4]`. We have to combine them into a number.
 
 | Step | number | digit | number * 10 + digit |
 |-----:|-------:|------:|:--------------------|
-|    1 |      0 |     3 | 3                   |
-|    2 |      3 |     2 | 32                  |
-|    3 |     32 |     1 | 321                 |
+|    1 |      0 |     7 | 7                   |
+|    2 |      7 |     5 | 75                  |
+|    3 |     75 |     4 | 754                 |
 
 And now the same in code:
 
--- a/exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
+++ b/exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
@@ -7,7 +7,10 @@ There are multiple ways to do it in C# and we will look at them in this article.
 
 ## The string reversal method
 
-Possibly the most obvious solution is to treat a number as a string of characters and reverse it. Let's have a look:
+Possibly the most obvious solution is to treat a number as a string of characters.
+We can then use string operations to see if the string is equal to its reversed version. 
+We are checking if it "reads" the same left-to-right and right-to-left.  
+Let's have a look:
 
 ```csharp
 using System.Linq;
@@ -26,7 +29,9 @@ We have to use `.Reverse()` extension method provided by [Language-Integrated Qu
 However, it returns an iterator and so, it has to be collected with `.ToArray()` and converted into a `String` with `new String(char[])` constructor. 
 
 
-Does this looks like a lot of code? 
+## The string reversal with string interpolation method
+
+Does the above, the [string reversal method](#the-string-reversal-method) look like a lot of code? 
 It might be slightly shortened using [string interpolation][interpolation-link] to convert a number into a text like in the example below. 
 However, before you use it read [the performance considerations](#performance-considerations) below.
 
@@ -103,12 +108,8 @@ private static bool IsPalindrome(int number)
 ```
 
 ## Converting to an extension method
-All of the above solutions correctly check if a number is a palindrome. 
-They differ in performance, something we will look at in a moment.
-But they also differ in readability. You will have to decide which one makes the most sense to you. 
-
-Talking about readability of code, we have another decision to make. 
-All of the above solutions can be implemented like they are above, or as an [extension method][extension-methods-link].
+Talking about readability of code, we have another point to make. 
+All of the above solutions can be implemented like they are above, or as an [extension methods][extension-methods-link].
 
 What does it change? The code reads differently. Instead of: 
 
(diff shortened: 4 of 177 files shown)


exercises/practice/palindrome-products/.articles/is-palindrome-check/snippet.md
```diff
@@ -0,0 +1 @@
+*I don't really know what this file is for, but the github action is failing without it*
```
ErikSchierboom, line 1: It's what's displayed on the right hand side of the dig deeper page as a small summary of the article.
See https://exercism.org/docs/building/tracks/practice-exercises#h-file-articles-article-slug-snippet-txt for more information.

Example:

image

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
```diff
@@ -0,0 +1,165 @@
+# Is number a palindrome?
```
ErikSchierboom, line 1: I think I would make this more matter-of-factly, not a question:

```suggestion
# Check if a number is a palindrome
```

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
```diff
@@ -0,0 +1,165 @@
+# Is number a palindrome? 
+
+There are multiple approaches to solve the palindrome product exercise. 
+Most of them require us to check if a given number is a palindrome.
```
ErikSchierboom, line 4: Don't all of them require this?

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
```diff
@@ -0,0 +1,165 @@
+# Is number a palindrome? 
+
+There are multiple approaches to solve the palindrome product exercise. 
+Most of them require us to check if a given number is a palindrome. 
+
+## Approaches
```
ErikSchierboom, line 6: I'd remove the header (this is an article, and approaches might be confusing here).

```suggestion
```

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
```diff
@@ -0,0 +1,165 @@
...
+    var original = number.ToString();
+    var reversed = new String(original.Reverse().ToArray());
+    return original.Equals(reversed);
+}
+```
+
+If you come from some other languages you might be surprised by the above code.
+Unfortunately, C# doesn't come with a built in method to reverse a string. 
+We have to use `.Reverse()` extension method provided by 
+[Language-Integrated Query (LINQ)](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/linq/). However, it returns an iterator and so it has to be collected with `.ToArray()`
```
ErikSchierboom, line 28: Our [markdown guidelines](https://exercism.org/docs/building/markdown/markdown#h-links) suggest to use reference links.

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
```diff
@@ -0,0 +1,165 @@
...
+If this looks like a lot of code to you, it might be slightly shortened using 
+[string interpolation](https://learn.microsoft.com/en-us/dotnet/csharp/tutorials/string-interpolation)
+to convert a number into a text like in the example below. However, before you use it
+read the performance considerations below.
+
+```csharp
+using System.Linq;
+
+private static bool IsPalindrome(int number)
+=>  $"{number}".Equals(new String($"{number}".Reverse().ToArray()));
```
ErikSchierboom, line 41: This is more consistent with existing docs:

```suggestion
private static bool IsPalindrome(int number) =>
    $"{number}".Equals(new String($"{number}".Reverse().ToArray()));
```

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
```diff
@@ -0,0 +1,165 @@
...
+There are multiple approaches to solve the palindrome product exercise. 
+Most of them require us to check if a given number is a palindrome. 
+
+## Approaches
+There are multiple ways to do it in C# and we will look at them in this article. 
+
+### String reversal
+
+Possibly the most obvious solution is to treat a number as a string of characters
+and reverse it. Let's have a look:
```
ErikSchierboom, line 12: You're currently breaking up lines at a cutoff point, but our [style guide](https://exercism.org/docs/building/markdown/markdown#h-one-sentence-per-line) suggests using one sentence per line (no cutoff). That would make this:

```suggestion
Possibly the most obvious solution is to treat a number as a string of characters and reverse it.
Let's have a look:
```

Of course, this would also apply elsewhere.

ErikSchierboom: And maybe be explicit about the last part of the logic too (comparing the strings):

```suggestion
The most straightforward solution is to treat a number as a string of characters and see if it is equal to the reversed string.
Let's have a look:
```

Feel free to change the wording.

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
```diff
@@ -0,0 +1,165 @@
...
+And now the same in code:
+
+```csharp
+using System.Collections.Generic;
+
+private static bool IsPalindrome(int number) { 
+    var original = number;
+    var digits = new List<int>();
+    
+    while (number > 0) {
```
ErikSchierboom, line 75: For consistency with other code:

```suggestion
    while (number > 0)
    { 
```

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
```diff
@@ -0,0 +1,165 @@
...
+|    1 |      0 |     3 | 3                   |
+|    2 |      3 |     2 | 32                  |
+|    3 |     32 |     1 | 321                 |
+
+And now the same in code:
+
+```csharp
+using System.Collections.Generic;
+
+private static bool IsPalindrome(int number) {
```
ErikSchierboom, line 71:
```suggestion
private static bool IsPalindrome(int number)
{ 
```

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
```diff
@@ -0,0 +1,165 @@
...
+    var original = number;
+    var digits = new List<int>();
+    
+    while (number > 0) { 
+        digits.Add(number % 10);
+        number /= 10;
+    }
+
+    var reversed = 0;
+    foreach(var digit in digits) {
```
ErikSchierboom, line 81:
```suggestion
    foreach(var digit in digits)
    {
```

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
```diff
@@ -0,0 +1,165 @@
...
+The list in the above code helps us to see what is going on. 
+The numbers are extracted, stored in the list and then collected. 
+But, this can be done on the fly, using just math and loops without the storage. 
+
+```csharp
+private static bool IsPalindrome(int number) { 
+    var original = number;
+    var reversed = 0;
+    
+    while (number > 0) {
```
ErikSchierboom, line 100:
```suggestion
    while (number > 0)
    { 
```

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
ErikSchierboom: Once this is removed, the other headings can become level two headings.

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
```diff
@@ -0,0 +1,165 @@
...
+    var reversed = 0;
+    foreach(var digit in digits) {
+        reversed = reversed * 10 + digit;
+    }
+    
+    return original == reversed;
+}
+```
+
+### The math without the list
```
ErikSchierboom, line 89: Maybe just:

```suggestion
### Math (without list)
```

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
```diff
@@ -0,0 +1,165 @@
...
+    while (number > 0) { 
+        reversed = reversed * 10 + number % 10;
+        number /= 10;
+    }
+    
+    return original == reversed;
+}
+```
+
+## The extension method options
```
ErikSchierboom, line 109:
```suggestion
## Converting to an extension method
```

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
```diff
@@ -0,0 +1,165 @@
...
+All of the above solutions correctly check if a number is a palindrome. 
+They differ in performance, something we will look at in a moment.
+But they also differ in readability. You will have to decide which one makes
+the most sense to you.
```
ErikSchierboom, line 113: I'd probably remove this section.

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
```diff
@@ -0,0 +1,165 @@
...
+```
+
+If you come from some other languages you might be surprised by the above code.
+Unfortunately, C# doesn't come with a built in method to reverse a string. 
+We have to use `.Reverse()` extension method provided by 
+[Language-Integrated Query (LINQ)](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/linq/). However, it returns an iterator and so it has to be collected with `.ToArray()` 
+and converted into a `String` with `new String(char[])` constructor. 
+
+
+If this looks like a lot of code to you, it might be slightly shortened using
```
ErikSchierboom, line 32: I would probably put this in a sub-heading.

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
```diff
@@ -0,0 +1,165 @@
...
+```
+
+You decide which one fits better with your code and expectations. 
+
+
+## Performance considerations
+
+OK, let's talk performance. I have tested the four methods above using
+[BenchmarkDotNet](https://benchmarkdotnet.org/). I won't share the details
+of where I run it, on what hardware as we are interested in relative performance only.
```
ErikSchierboom, line 138: Could you maybe include the code that was used to run the benchmarks? Like https://github.com/exercism/csharp/tree/main/exercises/practice/leap/.articles/performance/code does. The article itself can then link to the code (see https://exercism.org/tracks/csharp/exercises/leap/articles/performance), which allows students to reproduce it locally.

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
```diff
@@ -0,0 +1,165 @@
...
+```csharp
+if (number.IsPalindrome()) { /* do something */ }
+```
+
+You decide which one fits better with your code and expectations. 
+
+
+## Performance considerations
+
+OK, let's talk performance. I have tested the four methods above using
```
ErikSchierboom, line 136: We usually use "we" instead of "I" in our documents.

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
```diff
@@ -0,0 +1,165 @@
...
+[BenchmarkDotNet](https://benchmarkdotnet.org/). I won't share the details
+of where I run it, on what hardware as we are interested in relative performance only. 
+What is important, that in an individual test I am checking all number between 1 and 100,000
+to see if they are a palindrome or not. Here are the results:
```
ErikSchierboom, line 140: I think this (what numbers are being checked) are a key detail. Ideally, the benchmark would test with different numbers, from small to large. See https://github.com/exercism/csharp/blob/main/exercises/practice/reverse-string/.articles/performance/code/Program.cs#L10-L11 for an example.

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
```diff
@@ -0,0 +1,165 @@
...
+## Which one to choose?
+
+So which one to chose? The fastest always will be the fastest. But the difference is too small
+to perceive in human scales, even when comparing 100,000 numbers it is at most 20ms. 
+
+For 1,000,000 checks  the difference between the first string option and the math loop is 163ms to 12ms. 
+Still almost impossible to spot with our human senses. 
+
+And so, with performance so close, you can choose the one which is easiest to understand,
+one that reads the best.
```
ErikSchierboom, line 165: Great conclusion!

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
```diff
@@ -0,0 +1,165 @@
...
+What is important, that in an individual test I am checking all number between 1 and 100,000
+to see if they are a palindrome or not. Here are the results:
+
+
+|        Method |        Mean |     Error |    StdDev |
+|-------------- |------------:|----------:|----------:|
+|  String1Check | 13,131.6 μs | 100.98 μs |  94.45 μs |
+|  String2Check | 19,578.5 μs | 324.35 μs | 303.40 μs |
+|     ListCheck |  5,690.9 μs |  59.71 μs |  52.93 μs |
+| MathLoopCheck |    935.2 μs |  12.60 μs |  11.79 μs |
```
ErikSchierboom, line 148: Maybe use different method names that use similar names to the headings used in this file?
Something like:

```suggestion
|  ReversedString | 13,131.6 μs | 100.98 μs |  94.45 μs |
|  ReversedStringInterpolated | 19,578.5 μs | 324.35 μs | 303.40 μs |
|     ListOfDigits |  5,690.9 μs |  59.71 μs |  52.93 μs |
| MathLoop |    935.2 μs |  12.60 μs |  11.79 μs |
```

ErikSchierboom (commented):
Lovely work! I've left some comments.

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
michalporeba: Not all of them. It is possible to generate palindrome numbers in which case you don't need to test for the palindrome. Once you generate the numbers (by composing digits following set patterns) you only have to discover the factors.

ErikSchierboom: Ooh neat!

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
michalporeba: I had a quick look at recent style guides at [Microsoft Learn](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions), [dotnet project](https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/coding-style.md) and [Google](https://google.github.io/styleguide/csharp-style.html). Non of them provide clear guidance what to do in situations where `=>` operator is used to define a function where function has a single statement, but is too long for a single line. The common practice is to run with too long lines which make readability suffer. 

The style which is more common, as you suggest, in the documentation is to leave the `=>` hanging at the end of the line, but then it might be difficult to spot and quickly understand the meaning of the statement. The alternative is to start the next line with the operator. This is something I have recently picked up while mentoring the C# track. Some people out there use it to great effect. I also like how it falls in line with how we see the text. If you haven't seen it I do recommend [Seven Ineffective Coding Habits of Many Programmers by Kevlin Henney](https://www.youtube.com/watch?v=ZsHMHukIlJY), especially [the part about style](https://www.youtube.com/watch?v=ZsHMHukIlJY&t=1027s).

So, with some vagueness around the official style guides and emerging practice, perhaps we could improve here and over time help this become the majority option?

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
michalporeba: Done

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
michalporeba: Done

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
michalporeba: Thanks

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
michalporeba: I think this could be an interesting aspect but the article is getting long and complex as it is. My intention was to show that there are differences in performance for those curious about things like that, but there isn't much in it. Perhaps we could write a separate article to deal with performance specifically and there it could be explored in much more detail with various options shown. 

talking about performance, is there a way to write a tack level article? I think it would be interesting to explore either tools, or specifically BenchmarkDotNet?

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
michalporeba: This could be an interesting follow up article. The solution becomes more complex than the problem itself so I don't think including a generative approach in the approaches will be helpful, but it might be an interesting thing for those who are into digging deep. 

It would create a nice series of challenges - something I proposed in forum but is not really the way exercises work right now. You could have three connected exercises: Check if a number is a palindrome, palindrome product and refactoring to a generative approach. Each next one could have previous code as a starting point. It's a longer discussion for another day

ErikSchierboom: > This could be an interesting follow up article. The solution becomes more complex than the problem itself so I don't think including a generative approach in the approaches will be helpful, but it might be an interesting thing for those who are into digging deep.

Well, it might be actuall! For the reverse string exercise, I also included a Span-based approach which you normally wouldn't use, but could still be interesting to students. I think this would fit well too. But an article would be fine too.

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
michalporeba: Done

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
michalporeba: Done

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
michalporeba: I renamed the headers. Tell me what you think about the latest version.

exercises/practice/palindrome-products/.articles/is-palindrome-check/content.md
michalporeba: agreed. reading it again it doesn't add much substance to the content

michalporeba:
@ErikSchierboom I think the article is for another thorough review of it all. I have worked on / resolved most of the suggestions you had. Two, I'd rather not, and I have put my reasons above.

.github/workflows/pssanalyzer.yml
```diff
@@ -13,5 +13,5 @@ jobs:
   analyze:
     runs-on: ubuntu-20.04
     steps:
-      - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c
+      - uses: actions/checkout@8f4b7f84864484a7bf31766abe9204da3cbe65b3
```
ErikSchierboom, line 16: Why was this change made? As in: does this PR require it? If it isn't, could you revert it?

.github/workflows/test.yml
```diff
@@ -13,7 +13,7 @@ jobs:
   test:
     runs-on: ubuntu-20.04
     steps:
-      - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c
+      - uses: actions/checkout@8f4b7f84864484a7bf31766abe9204da3cbe65b3
```
ErikSchierboom, line 16: Why was this change made? As in: does this PR require it? If it isn't, could you revert it?

exercises/practice/palindrome-products/.articles/config.json
```diff
@@ -0,0 +1,11 @@
+{
+    "articles": [
+        {
+            "uuid": "987cf2d1-7225-436f-9228-b88120ae2e70",
+            "slug": "is-palindrome-check",
+            "title": "Is a palindrome number check",
+            "blurb": "A few ways to check if a number is a palindrome.",
```
ErikSchierboom, line 7: Minor tweak:

```suggestion
            "blurb": "Different ways to check if a number is a palindrome.",
```

This makes it slightly more appealing I feel.

ErikSchierboom (approved):

exercises/practice/palindrome-products/.articles/is-palindrome-check/snippet.md
```diff
@@ -0,0 +1,4 @@
+private static bool IsPalindrome(int number)
+{
+    /* implementation details */
+}
```
ErikSchierboom, line 4: I think we probably want something else here. I can come up with two things:

- A summary of the performance of the various methods
- A snippet of the most interesting implementation

ErikSchierboom (commented):
I've approved with a couple of comments you could look into.

michalporeba:
@ErikSchierboom I managed to fix the issue with that extra change in files which I haven't done that was picked up by my rebase, but now I have a configlet problem. I noticed you recently published about releasing new version, can you help see what's wrong here? I don't even know what configlet is. Thanks.

michalporeba:
@ErikSchierboom I don't see an option to rerun the actions. Could you do it?

exercises/practice/palindrome-products/.articles/is-palindrome-check/snippet.md
michalporeba: I have no started [a post on the forum](https://forum.exercism.org/t/article-snippets/5272) to get wider feedback on the best approach.

ErikSchierboom: I've posted there

ErikSchierboom: This snippet is the only thing left to do I think. What do you think about a tiny bit of text for the snippet?

michalporeba: I agree, that's the last thing left. I've spent a lot of time thinking about it and cannot find anything that really works in my head. The performance benchmark output doesn't give any value or information. I cannot think of a text that would work in scenario where we have

1. short text - the snippet
2. title
3. blurb - a short text

It breaks the flow we are used from newspapers, books, blogposts all around us. It's normally

1. something graphical to capture our attention
2. a short catchy title
3. a brief description to help you decide if you want to click it

The best I can think of is something like this
```
bool IsPalindrome(int number)
{
    /* Let's look how to implement it! */
}
```

It shows the article is about the `IsPalindrome` implementation, it explains what you will get from clicking the link, it will look like code sa be a bit more 'graphical', and not favour one implementation over the other or pre-empt what is in the article. 

That's the best I can come up with, but I'm open to suggestions, or examples from other articles or tracks. Please help, @ErikSchierboom

michalporeba:
```
bool IsPalindrome(int number)
{
    /* Let's implement it! */
}
```

ErikSchierboom: Let's just go with that

michalporeba: OK, let's try it and see how it looks. We can always change it. 
I have added the backticks to wrap around the code as you said in a forum that it is just a mardown text. I hope that's right.

ErikSchierboom (approved):

PR #2119, opened 2023-04-16
Fix grid formatting in instructions.md

psobolik:
The sample grid is not displayed correctly. This change surrounds it with preformatted text markup to correct the problem.


ErikSchierboom:
Will be fixed in https://github.com/exercism/csharp/pull/2122

PR #2121, opened 2023-04-17
Configlet beta11 has a peculiar test for formatting

michalporeba:
The release of 4.0.0 beta.11 configlet broke all the builds in C# track as all existing article and approach configs were considered to be not-formatted. 

I have doing some reading and posting about it I followed suggestion of [MatthijsBlom](https://forum.exercism.org/u/MatthijsBlom) I have run `configlet fmt -u`. Here is the result. 

[The issue has been discussed in the forum](https://forum.exercism.org/t/configlet-4-0-0-beta-11-breaks-c-track-builds/5277).

### DIFF
--- a/.github/workflows/pssanalyzer.yml
+++ b/.github/workflows/pssanalyzer.yml
@@ -13,5 +13,5 @@ jobs:
   analyze:
     runs-on: ubuntu-20.04
     steps:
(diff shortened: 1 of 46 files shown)


michalporeba:
I see IsaacG was faster. I'll close this pull request.

PR #2128, opened 2023-05-05
Update instructions.md

jxbrenner:
Speed is distance per unit time, updated all references to speed units of meters to meters per second

### DIFF
--- a/exercises/concept/need-for-speed/.docs/instructions.md
+++ b/exercises/concept/need-for-speed/.docs/instructions.md
@@ -2,7 +2,7 @@
 
 In this exercise you'll be organizing races between various types of remote controlled cars. Each car has its own speed and battery drain characteristics.
 
-Cars start with full (100%) batteries. Each time you drive the car using the remote control, it covers the car's speed in meters and decreases the remaining battery percentage by its battery drain.
+Cars start with full (100%) batteries. Each time you drive the car using the remote control, it increases the car's speed in meters per second and decreases the remaining battery percentage by its battery drain.
 
 If a car's battery is below its battery drain percentage, you can't drive the car anymore.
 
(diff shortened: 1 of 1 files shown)


exercises/concept/need-for-speed/.docs/instructions.md
```diff
@@ -2,7 +2,7 @@
 
 In this exercise you'll be organizing races between various types of remote controlled cars. Each car has its own speed and battery drain characteristics.
 
-Cars start with full (100%) batteries. Each time you drive the car using the remote control, it covers the car's speed in meters and decreases the remaining battery percentage by its battery drain.
+Cars start with full (100%) batteries. Each time you drive the car using the remote control, it increases the car's speed in meters per second and decreases the remaining battery percentage by its battery drain.
```
ErikSchierboom, line 5: To me this reads as: the speed is increased, which it isn't. I like the other two suggestions, but maybe change or revert this?

jxbrenner: You're right, increases the speed doesn't make sense if the car is already travelling. How about, "it causes the car to travel at its rated speed for 1 second and decreases the remaining battery percentage by its battery drain rate."? It's wordier but I think it helps lead the reader to `totalDistanceTravelled += speed * time`.

michalporeba: Hi, there is no time in this exercise, the speed is in distance per click of a button, not per unit of time. 
How about

"Every time you drive the car using the remote control, the car travels distance equal to its current speed and its battery level drop by the the value of battery drain." Or something like this?

jxbrenner: Distance per click of a button is just distance. But the exercise is full of references to speed, and anywhere you talk about speed, you infer time by the definition of speed. My suggestion is either:
- replace all references to speed to distance travelled per use of the remote control, or
- define the amount of time the car travels at rated speed per use of the remote control so you have the information you need to convert rated speed to distance travelled, ex. `_totalDistance += _speed * 1;` if the operating the remote control causes the car to travel at its rated speed for 1 second

ErikSchierboom: > Every time you drive the car using the remote control, the car travels distance equal to its current speed and its battery level drop by the the value of battery drain

I like this suggestion.

jxbrenner: > > Every time you drive the car using the remote control, the car travels distance equal to its current speed and its battery level drop by the the value of battery drain
> 
> I like this suggestion.

"travels distance equal to its current speed" makes no sense, speed does not equal distance.

PR #2132, opened 2023-05-13
extra test for lasagna

michalporeba:
**Problem:** It is possible to pass tests with incorrect implementation

I have seen a student submit the following for review of a solution that passes all the tests

```csharp
public int RemainingMinutesInOven(int minutes){
    return minutes - 10;
}
```

It is one of the first exercises students do and some of them are coming to Exercism with little programming experience. While, as previously discussed, Exercism is not intended to be a primary learning source, I think we can improve the test coverage to help them realise that they are doing something wrong. 

**Proposed Solution:**

I have added extra test to any solution which is not equivalent of the expected `return ExpectedMinutesInOven()-minutes` will fail.

### DIFF
--- a/exercises/concept/lucians-luscious-lasagna/LuciansLusciousLasagnaTests.cs
+++ b/exercises/concept/lucians-luscious-lasagna/LuciansLusciousLasagnaTests.cs
@@ -12,9 +12,16 @@ public class LuciansLusciousLasagnaTests
 
     [Fact]
     [Task(2)]
-    public void Remaining_minutes_in_oven()
+    public void Remaining_minutes_in_oven_for_fifteen()
     {
         Assert.Equal(15, new Lasagna().RemainingMinutesInOven(25));
+    }
+
+    [Fact]
+    [Task(2)]
+    public void Remaining_minutes_in_oven_for_seven()
+    {
+        Assert.Equal(7, new Lasagna().RemainingMinutesInOven(33));
     }
 
     [Fact]


exercises/concept/lucians-luscious-lasagna/LuciansLusciousLasagnaTests.cs
```diff
@@ -12,11 +12,18 @@ public void Expected_minutes_in_oven()
 
     [Fact]
     [Task(2)]
-    public void Remaining_minutes_in_oven()
+    public void Remaining_minutes_in_oven_for_fifteen()
```
ErikSchierboom, line 15: My only tiny nit is that I think I would prefer the test name to be about the input, so:

```suggestion
    public void Remaining_minutes_in_oven_after_twenty_five_minutes()
```

exercises/concept/lucians-luscious-lasagna/LuciansLusciousLasagnaTests.cs
```diff
@@ -12,11 +12,18 @@ public void Expected_minutes_in_oven()
...
     {
         Assert.Equal(15, new Lasagna().RemainingMinutesInOven(25));
     }
 
+    [Fact]
+    [Task(2)]
+    public void Remaining_minutes_in_oven_for_seven()
```
ErikSchierboom, line 22: See above comment

ErikSchierboom (commented):
Excellent idea! Two tiny nits.

ErikSchierboom (approved):

PR #2136, opened 2023-05-22
Fix links and typo in attack-of-the-trolls introduction

kwacky1:

PR #2139, opened 2023-05-23
Fix links and typo in attack-of-the-trolls introduction

kwacky1:
Couldn't reopen #2136 because I was trigger happy with my branch/fork deletion.

So creating anew as per https://github.com/exercism/csharp/pull/2136#issuecomment-1557039229

cc/ @ErikSchierboom


ErikSchierboom (approved):
Thanks!

PR #2145, opened 2023-06-01
Combine where and sum when using sum only in 'perfect numbers' exercise approach

aage:
https://forum.exercism.org/t/adding-a-dig-deeper-approach-to-the-c-perfect-numbers-exercise/5762

### DIFF
--- a/exercises/practice/perfect-numbers/.approaches/if-statements/content.md
+++ b/exercises/practice/perfect-numbers/.approaches/if-statements/content.md
@@ -47,7 +47,7 @@ For the lower bound, we can use `1`, as that is always the smallest factor.
 For the upper bound, we can leverage the fact that the second lowest factor is `2`, which means that any numbers greater than `number / 2` cannot be factors.
 
 We can use the [Enumerable.Sum() method][enumerable-sum] to sum all the factors, but since not all numbers are factors we need to deal with the numbers that aren't.
-Providing the [Enumerable.Sum() method][enumerable-sum] with a function tells it to keep every factor and use a `0` in place of every number that isn't, effectively disregarding it when summing.
+We provide the [Enumerable.Sum() method][enumerable-sum] with a lambda that tells it to keep every factor and use a `0` in place of every number that isn't, effectively disregarding it when summing.
 
 ```csharp
 var sum = Enumerable.Range(1, number / 2)
--- a/exercises/practice/perfect-numbers/.approaches/if-statements/content.md
+++ b/exercises/practice/perfect-numbers/.approaches/if-statements/content.md
@@ -18,13 +18,14 @@ public class PerfectNumbers
         if (number < 1)
             throw new ArgumentOutOfRangeException(nameof(number));
 
-        var sum = Enumerable.Range(1, number / 2)
-            .Sum(n => number % n == 0 ? n : 0);
+        var sumOfFactors = Enumerable.Range(1, number / 2)
+            .Where(factor => number % factor == 0)
+            .Sum();
 
-        if (sum < number)
+        if (sumOfFactors < number)
             return Classification.Deficient;
 
-        if (sum > number)
+        if (sumOfFactors > number)
             return Classification.Abundant;
 
         return Classification.Perfect;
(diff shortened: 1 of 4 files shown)


exercises/practice/perfect-numbers/.approaches/if-statements/content.md
```diff
@@ -18,14 +18,13 @@ public class PerfectNumbers
         if (number < 1)
             throw new ArgumentOutOfRangeException(nameof(number));
 
-        var sumOfFactors = Enumerable.Range(1, number / 2)
-            .Where(factor => number % factor == 0)
-            .Sum();
+        var sum = Enumerable.Range(1, number / 2)
```
ErikSchierboom, line 21: I quite like `sumOfFactors` better to be honest. It better reflects the domain IMHO.

exercises/practice/perfect-numbers/.approaches/if-statements/content.md
```diff
@@ -41,39 +40,42 @@ if (number < 1)
...
+A factor is a number that you can divide another number with and not end up with a remainder, which we can check using the modulo (`%`) operator: `number % factor == 0`.
 
 We then need to decide which numbers could possibly be factors.
 For the lower bound, we can use `1`, as that is always the smallest factor.
 For the upper bound, we can leverage the fact that the second lowest factor is `2`, which means that any numbers greater than `number / 2` cannot be factors.
 
-We can use the [`Enumerable.Range() method][enumerable-range] to iterate over the possible factors, then use [`Where()`][enumerable-where] to select only valid factors and then finally sum them via the [`Sum() method`][enumerable-sum]:
+We can use the [Enumerable.Sum() method][enumerable-sum] to sum all the factors, but since not all numbers are factors we need to deal with the numbers that aren't.
+Providing the [Enumerable.Sum() method][enumerable-sum] with a function tells it to keep every factor and use a `0` in place of every number that isn't, effectively disregarding it when summing.
```
ErikSchierboom, line 50:
```suggestion
We provide the [Enumerable.Sum() method][enumerable-sum] with a lambda that tells it to keep every factor and use a `0` in place of every number that isn't, effectively disregarding it when summing.
```

exercises/practice/perfect-numbers/.approaches/if-statements/content.md
```diff
@@ -41,39 +40,42 @@ if (number < 1)
...
 
-if (sumOfFactors > number)
+if (sum > number)
     return Classification.Abundant;
 
 return Classification.Perfect;
 ```
 
+## Sum vs. Where and Sum
+The line `.Sum(n => number % n == 0 ? n : 0);` can also be written as `.Where(n => number % n == 0).Sum();`, using [Enumerable.Where() method][enumerable-where] to _keep_ only the factors and then [Enumerable.Sum() method][enumerable-sum] the result, but this is a little more verbose.
```
ErikSchierboom, line 70: I feel like this is the wrong way around. In other approaches, I tend to start with the code that I then refactor. That way you show to the student how things can be improved.

ErikSchierboom (commented):
Some minor nits

exercises/practice/perfect-numbers/.approaches/if-statements/content.md
aage: Agreed. I wanted to align this with the other approach, but I'll submit a separate PR to fix this there as well.

exercises/practice/perfect-numbers/.approaches/if-statements/content.md
aage: Good point, I'll look into this. I'll also fix this in mentioned PR for other approach so it's more cohesive.

aage:
@ErikSchierboom In general, does Exercism prefer amended commits when dealing with feedback? Or doesn't it matter since PR's get merged?

aage:
@ErikSchierboom I've fixed some stuff in line with your comments, could you have a look?

ErikSchierboom:
> @ErikSchierboom In general, does Exercism prefer amended commits when dealing with feedback? Or doesn't it matter since PR's get merged?

Whatever you prefer basically :)

exercises/practice/perfect-numbers/.approaches/if-statements/content.md
```diff
@@ -77,6 +77,17 @@ return sumOfFactors < number ? Classification.Deficient :
        Classification.Perfect;
 ```
 
+## Combining Where and Sum vs. using Sum only
+Instead of using [Where][enumerable-where] and then [Sum][enumerable-sum] we could actually implement this by using [Sum][enumerable-sum] only.
```
ErikSchierboom, line 81:
```suggestion

Instead of using [Where][enumerable-where] and then [Sum][enumerable-sum] we could actually implement this by using [Sum][enumerable-sum] only.
```

ErikSchierboom (approved):
Lovely

PR #2170, opened 2023-08-27
tim-from-marketing exercise docs mistake.

Munib97:
Fixed a mistake in tim-from-marketing exercise introduction where the null-coalescing operator (??) was shown to be used without an assignment operator.


ErikSchierboom:
Thanks

PR #2172, opened 2023-09-02
Collatz Conjecture / Dig Deeper / Append optional parameter version to recursion approach

micha-b:
The recursion approach in the Dig Deeper section for the Collatz Conjecture exercise can be presented in a more concise way using a default parameter for `stepCount` instead of overloading the method. 
This version of the recursion approach is currently added as an alternative to the overloaded version, which is still kept at it's place. It might also be possible that this updated solution replaces the code snippet using overloading.

### DIFF
--- a/exercises/practice/collatz-conjecture/.approaches/introduction.md
+++ b/exercises/practice/collatz-conjecture/.approaches/introduction.md
@@ -59,7 +59,25 @@ private static int Steps(int number, int stepCount)
 }
 ```
 
-This approach uses a recursion to calculate the steps, with each call to `Steps(int number, int stepCount)` representing a single application of the algorithm.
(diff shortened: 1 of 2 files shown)


ErikSchierboom (commented):
It's a nice alternative, although I do still prefer the first version to be honest. The default value approach to me can be slightly confusing to the caller, as it might seem that the caller can pass in the value they desire.

What I would do is to leave the `introduction.md` untouched, but add a section to the end of the approach itself listing this as an alternative, with my above comment as a caveat.

micha-b:
@ErikSchierboom I updated the pull request with your feedback.

I removed the sections I added in the introduction.md document and added them in the recursion/content.md document instead in a newly created section.

I hope this is how you imagined it to be.

ErikSchierboom (approved):
Perfect!

PR #2175, opened 2023-09-10
fix: Greek character in SqueakyCleanTests.cs file

wallacewi:
Just a simple fix:
```csharp
Assert.Equal("MyΟFinder", Identifier.Clean("MyΟβιεγτFinder")); // before
```
The string `"MyΟFinder"` contains a Greek character remaining: `Ο` (_omicron_)


iHiD:
Discussion at https://github.com/exercism/csharp/pull/2175/files

I'm going to leave this closed as I think a different solution is required. But please do give your thoughts on the forum post :)

PR #2189, opened 2023-10-12
Fix prefix description in Log Files Description

Gautzilla:
The lines not containing an offending password should be returned prefixed with `"--------: "`, and so does the unit test expect.

I fixed the prefix in the exercise description, which was missing one "-".


ErikSchierboom (approved):
Good catch

PR #2193, opened 2023-10-18
feat: improve tampering test on exercise "Authentication System"

salvatore-piccione:
Tests for `Authentication System` exercises 4 and 5 can be improved by trying to actually tamper the data to ensure that the expected changes have been done. This PR adds two test cases:

- the former is about the tampering check on the admin
- the latter is about the tampering of developers by replacing an element in the dictionary

Exemplar solution has been updated to be more inline with the exercise request (i.e. `getDevelopers()` returns an instance of the interface `IDictionary` despite of an instance of `ReadOnlyDictionary`).

### DIFF
--- a/exercises/concept/authentication-system/AuthenticationSystemTests.cs
+++ b/exercises/concept/authentication-system/AuthenticationSystemTests.cs
@@ -15,6 +15,17 @@ public class AuthenticationSystemTests
     }
 
     [Fact]
+    [Task(4)]
+    public void CheckAdminCannotBeTampered()
+    {
+        var admin = new Identity { EyeColor = "green", Email = "admin@ex.ism" };
+        var authenticator = new Authenticator(admin);
+        var tamperedAdmin = authenticator.Admin;
+        tamperedAdmin.Email = "admin@hack.ed";
+        Assert.NotEqual(tamperedAdmin.Email, authenticator.Admin.Email);
+    }
+
+    [Fact]
     [Task(5)]
     public void GetDevelopers()
     {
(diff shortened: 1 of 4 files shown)


salvatore-piccione:
Hello @ErikSchierboom, I saw that checks on Configlet failed. AFAIK this is not due by the changes on this PR since the failure affects other exercises.

exercises/concept/authentication-system/AuthenticationSystemTests.cs
```diff
@@ -14,6 +14,17 @@ public void GetAdmin()
...
 
+    [Fact]
+    [Task(4)]
+    public void CheckAdminCannotBeTampered()
+    {
+        var admin = new Identity { EyeColor = "green", Email = "admin@ex.ism" };
+        var authenticator = new Authenticator(admin);
+        var tamperedAdmin = authenticator.Admin;
+        tamperedAdmin.Email = "admin@hack.ed";
+        Assert.NotEqual(tamperedAdmin.Email, authenticator.Admin.Email);
```
ErikSchierboom, line 25: Maybe do:

```suggestion
        Assert.Equal("admin@ex.ism", authenticator.Admin.Email);
```

This is a slightly stronger assertion, as the current test theoretically still passes when the admin email is changed to something else than the tampered admin.

exercises/concept/authentication-system/AuthenticationSystemTests.cs
```diff
@@ -24,4 +35,22 @@ public void GetDevelopers()
...
+
+        Identity tamperedDev = new Identity { EyeColor = "grey", Email = "anders@hack.ed" };
+        try
+        {
+            devs["Anders"] = tamperedDev;
+            Assert.True(false, "Unexpected change to developers list.");
+        } catch (NotSupportedException)
+        {
+            Assert.True(true);            
+        }
```
ErikSchierboom, line 54: This should also work I think:

```suggestion
        Assert.Throws<NotSupportedException>(() => devs["Anders"] = tamperedDev);
```

exercises/concept/authentication-system/AuthenticationSystemTests.cs
salvatore-piccione: Totally agree!

exercises/concept/authentication-system/AuthenticationSystemTests.cs
salvatore-piccione: Changes done.

exercises/concept/authentication-system/AuthenticationSystemTests.cs
salvatore-piccione: Changes done.

ErikSchierboom (approved):
Thanks!

PR #2207, opened 2024-01-03
Sync fetch-configlet.ps1 script

ErikSchierboom:
This PR updates the bin/fetch-configlet.ps1 script to the latest version. This fixes the issue mentioned in this forum post: https://forum.exercism.org/t/problem-running-the-powershell-version-of-fetch-configlet/8383.

References https://github.com/exercism/configlet/issues/839

### DIFF
--- a/exercises/practice/reverse-string/.approaches/span/content.md
+++ b/exercises/practice/reverse-string/.approaches/span/content.md
@@ -1,12 +1,20 @@
 # Span&lt;T&gt;
 
 ```csharp
-Span<char> chars = stackalloc[input.Length];
-for (var i = 0; i < input.Length; i++)
+public static class ReverseString
 {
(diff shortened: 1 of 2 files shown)


exercises/practice/reverse-string/.approaches/span/content.md
```diff
@@ -1,12 +1,20 @@
 # Span&lt;T&gt;
```
wolf99, line 1: Seems like some other change to an unrelated file is added in this PR. Expected or incorrect?

wolf99 (changes requested):

exercises/practice/reverse-string/.approaches/span/content.md
ErikSchierboom: Whoa. Missed this comment. It was unintended but correct

PR #2222, opened 2024-01-31
Show Test Report Summary in CI

### LINKED ISSUE #1949: Show test summary in CI
With Github Actions now supporting Job Summaries, we can leverage this to show the test results in a nice way.
For that, we can use the Test Summary Action: https://github.com/test-summary/action
See https://github.com/test-summary/examples/tree/main/dotnet for how to integrate this with .NET.
Note: we should only do this when run in CI, so the `test.ps1` script should detect this (probably via an environment variable, see https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables).

putquo:
### Description
This PR aims to resolve #1949.

### Changes
* Added [test-summary/action](https://github.com/test-summary/action) to the `Test` workflow
* Added `JunitXml.TestLogger` as a dependency to all existing exercises
* Added `JunitXml.TestLogger` as a dependency to new exercises via the `add-new-exercise.ps1` script
* Updated the `test.ps1` script to use the test logger when the `CI` environment variable evaluates to `true`

### Notes
Since the tests do not specify a namespace, the logger defaults to `UnknownNamespace`. [Here](https://github.com/putquo/csharp/actions/runs/7678263985) is an example of what this looks like.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,3 +18,7 @@ jobs:
         with:
           dotnet-version: "7.0.100"
       - run: pwsh ./test.ps1
+      - uses: test-summary/action@fee35d7df20790255fe6aa92cf0f6d28092ecf2f
+        with:
+          paths: build/**/results/**/*.xml
+        if: always()
--- a/test.ps1
+++ b/test.ps1
@@ -51,7 +51,7 @@ function Enable-All-UnitTests($BuildDir) {
 function Test-Refactoring-Projects($PracticeExercisesDir) {
     Write-Output "Testing refactoring projects"
     @("tree-building", "ledger", "markdown") | ForEach-Object {
-        Invoke-CallScriptExitOnError { dotnet test "$practiceExercisesDir/$_" }
+        Invoke-Tests -Path "$PracticeExercisesDir/$_"
     }
 }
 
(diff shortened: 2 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -18,3 +18,7 @@ jobs:
         with:
           dotnet-version: "7.0.100"
       - run: pwsh ./test.ps1
+      - uses: test-summary/action@v2
```
ErikSchierboom, line 21: Could you pin this to a SHA? See https://exercism.org/docs/building/github/gha-best-practices#h-pin-actions-to-shas

ErikSchierboom (changes requested):
Thanks for working on this! I'd like to suggest a slightly different approach. Right now, students will have to download the `JunitXml.TestLogger` package even if they're just working locally, which doesn't really offer them anything. What I would suggest to do instead is to modify the `test.ps1` script to dynamically add the package (via `dotnet add package JunitXml.TestLogger`) when it detects that it runs in CI.

putquo:
> Thanks for working on this! I'd like to suggest a slightly different approach. Right now, students will have to download the `JunitXml.TestLogger` package even if they're just working locally, which doesn't really offer them anything. What I would suggest to do instead is to modify the `test.ps1` script to dynamically add the package (via `dotnet add package JunitXml.TestLogger`) when it detects that it runs in CI.

Good point! I have made the change as you suggested. This does increase the run time of the workflow quite a bit (if that is of any concern), but I managed to shave off about a couple of minutes with the `--no-restore` option since it will run during `dotnet test` anyway. Maybe you have some ideas if you think it may be problematic?

ErikSchierboom:
> Maybe you have some ideas if you think it may be problematic?

I think this is fine!

ErikSchierboom (approved):
Excellent work!

ErikSchierboom:
Maybe a silly question, but where can I see the summary?

putquo:
> Maybe a silly question, but where can I see the summary?

I wondered that at first too :) but you can view it over [here](https://github.com/exercism/csharp/actions/runs/7744453074). In general, you’ll need to go to the “summary” section of a workflow run (top left corner on web) and scroll to the bottom of the page to see it. It seems like you also have to be signed in to see it.

ErikSchierboom:
Right! It's unimpressive right now because everything is passing, but reading up on the docs, it will be very useful when tests are failing:

> show: the test results to summarize in a table (optional) This controls whether a test summary table is created or not, as well as what tests are included. It could be all, none, pass, skip, or fail. The default is fail - that is, the summary table will only show the failed tests. For example, if you wanted to show failed and skipped tests:

Right now finding failing tests involves some annoying text searching.

ErikSchierboom:
Thanks a lot! If you care for it, you'd be welcome to do the same thing for `exercism/vbnet` or `exercism/fsharp`.

PR #2227, opened 2024-02-16
Update instructions.md

dawidsip:
At the end of  Step 2 and Step 4 there was a sentence "The program must also report any values that are out of range.' As this information, being served twice, sreamed redundant , I think it would be better to leave it only for Step 4, and remove it from Step 2.


ErikSchierboom:
The text for this exercise is automatically synced from https://github.com/exercism/problem-specifications/blob/main/exercises/say/description.md. If you think the text can be improved, please open an issue on the forum: https://forum.exercism.org/c/exercism/4 We can then discuss any changes there.

PR #2231, opened 2024-02-27
Update a test for secure-munchester-united with the correct expected value

CodeVelo:
In the `DisplaySecurityPass_security` there was a mismatch between the expected value and the actual title for `Security` class. This pull request updates the expected value to match the title from the `Security` class - `Security Team Member`.


ErikSchierboom:
Could you maybe read task 2 again? It explicitly states that " Priority Personnel" must be added to any `Security` class.

PR #2253, opened 2024-04-20
Correct wrong claim about StringBuilder performance.

Pedro-V:
The current text on the Dig Deeper article for the C# exercise "Reverse String" incorrectly says the `StringBuilder` approach has the worst performance of those listed, but it's actually the 2nd worst, behind the LINQ-based.

Probably just a typo, for which this PR provides a correction.


ErikSchierboom (approved):
Thanks

PR #2272, opened 2024-05-19
Fix typo for Telemetry in remote-control-cleanup

dmarcoux:
Nothing major as it's going to be removed anyway to finish the exercise, but it's still a typo.


dmarcoux:
I didn't start a discussion on the forum since this is such a tiny change, that I don't see why it would be rejected.

PR #2287, opened 2024-07-12
Update GameOfLifeTests.cs

errolpais:
The rule in the instructions "Any dead cell with exactly three live neighbors becomes a live cell." is not explicitly covered by the test "Dead_cells_with_three_live_neighbors_become_alive". 

This PR updates the relevant test to match the rule from the instructions more explicitly by covering both the scenarios where there are exactly 3 live neighbors and when there are 4 live neighbors for a cell.


ErikSchierboom:
The tests are synced from a shared source (https://github.com/exercism/problem-specifications). If you think we should expand the tests, please open a forum post on our forum to discuss the change: https://forum.exercism.org/c/exercism/4

PR #2293, opened 2024-07-31
Sync reverse-string tests

BNAndras:


ErikSchierboom (commented):
I'm not entirely sure I'm in favor of this change. It would break all existing solutions. I think I'd rather prefer a separate Concept Exercise that explained Unicode and grapheme clusters.

BNAndras:
Under stood.

BNAndras:
I'll revert the changes and just mark the Unicode tests as not included.

ErikSchierboom:
Thanks!

PR #2296, opened 2024-08-12
Fix CultureInfo being overridden by user settings

huynhloc-1110:
`new CultureInfo(string name)` by default sets `useUserOverride` to `true`, which uses Windows user customized settings. This results in local tests failed on some machines with customized date time format. Use `CultureInfo.GetCultureInfo(string name)` instead which gets the cached instance or creates a new instance with `useUserOverride` set to `false`.

### DIFF
--- a/exercises/concept/beauty-salon-goes-global/.meta/Exemplar.cs
+++ b/exercises/concept/beauty-salon-goes-global/.meta/Exemplar.cs
@@ -86,7 +86,7 @@ public static class Appointment
                 cultureId = "fr-FR";
                 break;
         }
-        return new CultureInfo(cultureId);
+        return CultureInfo.GetCultureInfo(cultureId);
     }
 
     private static string GetTimeZoneId(Location location)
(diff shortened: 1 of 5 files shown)


ErikSchierboom:
The F# track likely has the same mistake. Would you be willing to submit a PR to that too?

ErikSchierboom:
CI is failing:

```
Error: /home/runner/work/csharp/csharp/bin/build/practice/ledger/Ledger.cs(36,34): error CS0119: 'Ledger.CultureInfo(string)' is a method, which is not valid in the given context [/home/runner/work/csharp/csharp/bin/build/practice/ledger/Ledger.csproj]
Error: /home/runner/work/csharp/csharp/bin/build/practice/ledger/Ledger.cs(37,34): error CS0119: 'Ledger.CultureInfo(string)' is a method, which is not valid in the given context [/home/runner/work/csharp/csharp/bin/build/practice/ledger/Ledger.csproj]
```

huynhloc-1110:
Hi, I have fixed the issue with `CultureInfo` in `Ledger` exercise. You can check the latest commit. About the F# track, yes, I will check the repo and submit a PR soon.

PR #2300, opened 2024-08-27
Add more tests to LandGrabInSpaceTests.cs

cizra:
See discussion at https://forum.exercism.org/t/land-grab-in-space-has-insufficient-test-coverage-for-the-plot-with-longest-side/9615

### DIFF
--- a/exercises/concept/land-grab-in-space/LandGrabInSpaceTests.cs
+++ b/exercises/concept/land-grab-in-space/LandGrabInSpaceTests.cs
@@ -76,9 +76,7 @@ public class LandGrabInSpaceTests
     public void GetLongestSideDiagonal1()
     {
         var ch = new ClaimsHandler();
-        // Bottom left, bottom right, top left, top right. The longest side is 10, diagonal is >14
         var shorter = CreatePlot(new Coord(0, 0), new Coord(0, 10), new Coord(10, 0), new Coord(10, 10));
-        // Bottom left, top left, top right, bottom right, longest side 11
         var longer = CreatePlot(new Coord(0, 0), new Coord(11, 0), new Coord(11, 11), new Coord(0, 11));
         ch.StakeClaim(shorter);
         ch.StakeClaim(longer);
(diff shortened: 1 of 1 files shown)


exercises/concept/land-grab-in-space/LandGrabInSpaceTests.cs
```diff
@@ -59,6 +59,32 @@ public void GetLongestSide()
...
+    }
+
+    [Fact]
+    [Task(4)]
+    public void GetLongestSideDiagonal1()
+    {
+        var ch = new ClaimsHandler();
+        // Bottom left, bottom right, top left, top right. The longest side is 10, diagonal is >14
+        var shorter = CreatePlot(new Coord(0, 0), new Coord(0, 10), new Coord(10, 0), new Coord(10, 10));
+        // Bottom left, top left, top right, bottom right, longest side 11
```
ErikSchierboom, line 81: Not sure about the comments. Does the student need them?

```suggestion
        var shorter = CreatePlot(new Coord(0, 0), new Coord(0, 10), new Coord(10, 0), new Coord(10, 10));
```

ErikSchierboom:
Thanks!

Naeddyr:
Adding these tests made the "learning exercise" significantly harder. A better solution would be to add a guarantee that the coordinates of the plot are always in the correct order.

cizra:
> Adding these tests made the "learning exercise" significantly harder. A better solution would be to add a guarantee that the coordinates of the plot are always in the correct order.

Disagreed, @Naeddyr - before adding these tests, the most difficult task in the exercise could be completed by just returning the 1st element of the list, which is not in the spirit of the task. The task was about calculating sizes of plots, and this just wasn't necessary. To be honest, the bit with dialog could have been skipped, tho.

Naeddyr:
The diagonals test makes this learning exercise (originally about hash sets and hashing and marginally about the data itself) an intermediate exercise about generating a convex polygon from a set of points or coming up with some other algorithm. You shouldn't just add a completely new category of test because it feels right. The correct way to do this would have been to make the old tests better. Randomize the elements of the list, add more elements to the list, that sort of thing.

PR #2318, opened 2024-10-25
fix: Added Skip to the Fact attribute

DerSimeon:
The Title says it all. All Tests in the csharp track except this one have the Skip Attribute on each test except the first one. This Exercise was left out.


ErikSchierboom:
This is a refactoring exercise and the stub implementation should therefore pass all the tests from the start. There is no need for TDD.

PR #2344, opened 2024-12-05
Updated DiamondTests.cs

spirit11:
`Rows` method adjusted to use system dependent new line characters.

Using '\n' instead of `Environment.NewLine` caused an issue, when all strings except the last one contain '\r' extra character, which leads to property tests failure.


kotp:
Forum discussion is at https://forum.exercism.org/t/c-track-diamond-task-test-fix-suggestion/14056

spirit11 (commented):
Rows method adjusted

spirit11 (commented):
Row method adjusted

PR #2354, opened 2025-01-29
Improved test for roll_the_die

nash8114:
- Tests now check whether multiple rolls supply different results
 - RollDie is checked for returning all values in the expected range
 - Increased for-loop from 100 times to 1000 times to prevent false negative. Encountered that 100 rolls was not sufficient to have all numbers in range(1, 18) to be generated.

Rationale:

My first implementation (thanks XKCD) made the test suite pass.

```
public int RollDie()
{
  return 4; // chosen by fair dice roll.
            // guaranteed to be random.
}
```

### DIFF
--- a/exercises/concept/roll-the-die/RollTheDieTests.cs
+++ b/exercises/concept/roll-the-die/RollTheDieTests.cs
@@ -1,5 +1,6 @@
 using Xunit;
 using Exercism.Tests;
+using System.Collections.Generic;
 
 public class RollTheDieTests
 {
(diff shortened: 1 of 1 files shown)


exercises/concept/roll-the-die/RollTheDieTests.cs
```diff
@@ -1,25 +1,38 @@
...
         var player = new Player();
-        var strength = player.GenerateSpellStrength();
-        Assert.InRange(strength, 0.0, 100.0);
+        for (var i = 0; i < rollCount; i++)
+        {
+            var strength = player.GenerateSpellStrength();
+            rolls.Add(strength);
+            Assert.InRange(strength, 0.0, 100.0);
+        }
+        Assert.Equal(rollCount, rolls.Count);
```
ErikSchierboom, line 36: Does this mean that you expect all 100 rolls to be unique, correct? If so, I wonder if we should maybe be slightly more tolerant of accidental duplicates. Maybe something like `Assert.True(rolls.Count >= rollCount - 5)` or something like that

ErikSchierboom (commented):
Great idea! I've left a comment.

KaranChadha10:
@ErikSchierboom I have raised a PR https://github.com/exercism/csharp/pull/2442 with the comment resolved on this PR, kindly take a look thanks.

KaranChadha10:
Hi @ErikSchierboom ,

This PR can be closed as PR #2442 has already been merged with the same changes. The open PR now has conflicts due to this.

Thanks!

PR #2402, opened 2025-03-05
Add `baffling-birthdays` exercise

ErikSchierboom:


BethanyG:
Hope I'm not being to nosey here...
I just saw this fly past in my GH queue.  **Nice one!**  
Are you going to propose this one for problem specs?  If so - I am definitely on board for it!  😄

ErikSchierboom:
> Hope I'm not being to nosey here...
Not at all!

> I just saw this fly past in my GH queue. Nice one!
Thanks!

> Are you going to propose this one for problem specs? If so - I am definitely on board for it! 😄
100%! I usually go for a C# implementation to come up with concrete test cases, and then convert to prob-specs. Great to hear you like it!

PR #2422, opened 2025-04-24
chore: fix typo

lapiselazul:
Initially, this hint wasn't indicating that we need an extra, separate field to track the battery charge, as demonstrated by many popular solutions. This led me to think that I should keep the field I made from answer 2 as it is, and calculate the battery charge based on that.

I propose to change this to better reflect expectations: that the solver has to create a field that tracks the battery charge value separately.


ErikSchierboom (approved):
Well spotted!

PR #2423, opened 2025-04-27
Add missing information about accessing DB

tvedeane:
This PR tries to solve issues discussed briefly here: https://forum.exercism.org/t/adds-database-class-to-objectrelationalmapping-cs/16314

### DIFF
--- a/exercises/concept/object-relational-mapping/.docs/instructions.md
+++ b/exercises/concept/object-relational-mapping/.docs/instructions.md
@@ -10,10 +10,19 @@ Note that, internally, the database transitions through a number of state: Close
 
 The database has the following instance methods:
 
-- `Database.BeginTransaction()` starts a transaction on the database. If this is called when the database is not in a `Closed` state then an exception is thrown. If successful the internal state of the database will change to `TransactionStarted`.
-- `Database.Write(string data)` writes data to the database within the transaction. If it receives bad data an exception will be thrown. An attempt to call this method without `BeginTransaction()` having been called will cause an exception to be thrown. If successful the internal state of the database will change to `DataWritten`.
-- `Database.EndTransaction()` commits the transaction to the database. It may throw an exception if it can't close the transaction or if `Database.BeginTransaction()` had not been called.
+- `Database.Begin()` starts a transaction on the database. If this is called when the database is not in a `Closed` state then an exception is thrown. If successful the internal state of the database will change to `TransactionStarted`.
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
I appreciate that this is confusing. The proper fix is to move the `Database` class out of the tests file (https://github.com/exercism/csharp/blob/main/exercises/concept/object-relational-mapping/ObjectRelationalMappingTests.cs#L83) and into it's own, separate file. It then needs to be added to the `files.editor` key in the `.meta/config.json` file (https://github.com/exercism/csharp/blob/main/exercises/concept/object-relational-mapping/.meta/config.json).

tvedeane:
Thanks for explaining @ErikSchierboom . I followed your suggestions, but I wasn't sure about the details. So this should be definitely properly reviewed (and ideally tested).

exercises/concept/object-relational-mapping/.docs/instructions.md
```diff
@@ -10,11 +10,20 @@ Note that, internally, the database transitions through a number of state: Close
 
 The database has the following instance methods:
 
-- `Database.BeginTransaction()` starts a transaction on the database. If this is called when the database is not in a `Closed` state then an exception is thrown. If successful the internal state of the database will change to `TransactionStarted`.
-- `Database.Write(string data)` writes data to the database within the transaction. If it receives bad data an exception will be thrown. An attempt to call this method without `BeginTransaction()` having been called will cause an exception to be thrown. If successful the internal state of the database will change to `DataWritten`.
-- `Database.EndTransaction()` commits the transaction to the database. It may throw an exception if it can't close the transaction or if `Database.BeginTransaction()` had not been called.
+- `Database.Begin()` starts a transaction on the database. If this is called when the database is not in a `Closed` state then an exception is thrown. If successful the internal state of the database will change to `TransactionStarted`.
```
ErikSchierboom, line 13: Isn't the original correct though?

exercises/concept/object-relational-mapping/.meta/config.json
```diff
@@ -11,7 +11,8 @@
       "ObjectRelationalMapping.cs"
     ],
     "test": [
-      "ObjectRelationalMappingTests.cs"
+      "ObjectRelationalMappingTests.cs",
+      "Database.cs"
     ],
```
ErikSchierboom, line 16: I meant like this:

```suggestion
    ],
  "invalidator": [
      "Database.cs"
    ],
```

exercises/concept/object-relational-mapping/.docs/instructions.md
tvedeane: Ah, true. I got confused by the method in `Orm` that is called `Begin`.

ErikSchierboom:
Thanks

PR #2424, opened 2025-04-28
Add `relative-distance`

BNAndras:
I still need to figure updating the generator template, but I think this gets us most of the way otherwise. I'll poke at this more in a few days.

### DIFF
--- a/config.json
+++ b/config.json
@@ -2735,8 +2735,16 @@
         "slug": "relative-distance",
         "name": "Relative Distance",
         "uuid": "4f4229ce-5a2a-415c-a682-6a56cf3846ac",
-        "practices": [],
-        "prerequisites": [],
+        "practices": [
+          "constructors",
+          "dictionaries"
+        ],
+        "prerequisites": [
+          "arrays",
+          "classes",
+          "constructors",
+          "dictionaries"
+        ],
         "difficulty": 5
       },
       {
--- a/config.json
+++ b/config.json
@@ -2732,6 +2732,14 @@
         "difficulty": 6
       },
       {
+        "slug": "relative-distance",
+        "name": "Relative Distance",
+        "uuid": "4f4229ce-5a2a-415c-a682-6a56cf3846ac",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 5
+      },
+      {
         "slug": "split-second-stopwatch",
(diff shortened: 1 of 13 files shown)


config.json
```diff
@@ -2731,6 +2731,14 @@
         ],
         "difficulty": 6
       },
+      {
+        "slug": "relative-distance",
+        "name": "Relative Distance",
+        "uuid": "4f4229ce-5a2a-415c-a682-6a56cf3846ac",
+        "practices": [],
+        "prerequisites": [],
```
ErikSchierboom, line 2739: Could you add some prerequisites?

exercises/practice/relative-distance/RelativeDistanceTests.cs
```diff
@@ -0,0 +1,235 @@
+public class RelativeDistanceTests
+{
+    [Fact]
+    public void Direct_parent_child_relation()
+    {
+        var familyTree = new Dictionary<string, List<string>>
```
ErikSchierboom, line 6: Maybe use an array to indicate that the data is static?

```suggestion
        var familyTree = new Dictionary<string, string[]>
```

ErikSchierboom (commented):
> I still need to figure updating the generator template, but I think this gets us most of the way otherwise. 

Should be do-able just by looking at some of the existing generators.

BNAndras:
Locally, `configlet fm` gives a different result than the CI.

```plaintext
(base) ➜  exercism-csharp git:(add-relative-distance) ./bin/configlet fmt -uye object-relational-mapping
Found 42 Concept Exercises and 139 Practice Exercises in /home/anagy/Documents/GitHub/exercism-csharp/config.json
Looking for exercises that lack a formatted '.meta/config.json', '.approaches/config.json'
or '.articles/config.json' file...
The `object-relational-mapping` exercise config is already formatted!
```

ErikSchierboom (approved):

ErikSchierboom:
Needs a rebase for the ci error

BNAndras:
That generator was actually a lot simpler than I thought. :)

ErikSchierboom:
> That generator was actually a lot simpler than I thought. :)

That was my hope when writing it! Good to hear that was successful.

ErikSchierboom:
Thanks for the PR!

PR #2429, opened 2025-06-13
developer-privileges: Two tasks only

keiravillekode:
Bertrand and Anders are now both tested in Task 2.

This makes DeveloperPrivilegesTests.cs consistent with instructions.md

[no important files changed]


ErikSchierboom (approved):

ErikSchierboom:
Thanks

PR #2443, opened 2025-10-05
test: Add edge case tests for YearsBeforeDesiredBalance in Interest_Is_Interesting

### LINKED ISSUE #1976: interest-is-interesting: years before reaching desired balance when targetBalance <= balance
The "Interest is Interesting" currently does not have a test case for task 4 where the target balance has been met by the balance already. If you target savings is 1000, but your balance is already - say - 1200, I would expect the method to return "0": the target has already been met.

I have seen implementations using a do-while loop, which would return `1` at the very least, whereas a while loop would return `0`. Both currently pass as tests.

I would propose to add two test cases: one where the target balance is equal to the balance, and where the target balance is less than the balance. Both of which should return `0`. Furthermore, the instructions should probably clarify that this is expected behaviour (as a hint perhaps?)

Happy to create a PR for this, but wanted to check whether this is indeed the desired behaviour first. This particular task does not appear to be present in other tracks that have the interest-is-interesting exercise, so I did not have any comparison material.

KaranChadha10:
**Description**

This PR contains the following points:

- Add test case for when target balance equals current balance
- Add test case for when target balance is already exceeded
- Both cases should return 0 years as target is already met
- Prevents incorrect do-while implementations from passing tests

Fixes #1976

### DIFF
--- a/exercises/concept/interest-is-interesting/.meta/config.json
+++ b/exercises/concept/interest-is-interesting/.meta/config.json
@@ -1,7 +1,8 @@
 {
   "authors": [
     "ErikSchierboom",
-    "yzAlvin"
+    "yzAlvin",
+    "karanchadha10"
   ],
   "files": {
     "solution": [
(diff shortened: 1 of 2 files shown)


KaranChadha10:
@ErikSchierboom This PR was auto-closed by GitHub Actions, but I believe the issue has been properly resolved. Could you please review the changes when you have a chance?
The PR addresses the issue https://github.com/exercism/csharp/issues/1976.
Thank you!

